PR #5, opened 2023-02-17
Add general styling across the whole site.

lionell-pack-ttd:
Make some lint rules more friendly.
Switch crypto libraries to hopefully avoid intermittent dev problems.


lionell-pack-ttd:
Example screenshot.
image

jingyi-gao-ttd:
This PR looks great to me. just have a question, in which case the .scss should goes to styles foler? Should shared and theme.scss belongs to styles?

lionell-pack-ttd:
> This PR looks great to me. just have a question, in which case the .scss should goes to styles foler? Should shared and theme.scss belongs to styles?

Good catch - shared was just trying something out, I got rid of it and moved theme.scss and colors.scss into shared.

I think src/web/styles should be for any scss that doesn't belong with a specific component or page - I left App and index in the src/web folder because it feels like they belong with the root of the site.

jingyi-gao-ttd (approved):

PR #6, opened 2023-02-28
Custom login flow

jingyi-gao-ttd:


src/database/migrations/20230215233646_CreateParticipantTables.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  await knex.schema.alterTable('users', (table) => {
+    table.integer('participant_id').unsigned();
+    table.foreign('participant_id').references('participants.id');
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
+  knex.schema.alterTable('users', (table) => {
+    table.dropColumn('participant_id');
```
lionell-pack-ttd, line 34: The down operation wouldn't work for me - I added an await to the alterTable call and switched around the dropColumn and dropForeign calls (I think you need to remove the FK before you can drop the column).

src/database/seeds/ParticipantTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+    typeName: 'Data Provider',
+  },
+  {
+    typeName: 'Publisher',
+  },
+];
+
+export async function seed(knex: Knex): Promise<void> {
+  // Deletes ALL existing entries
+  await knex('participants_types').del();
```
lionell-pack-ttd, line 24: This will break the data every time the seeds run. e.g. if there's a Publisher 'Disney' linked to participants_types.id=1, when this drops all the participants and reinserts them Disney will be linked to a non-existent type 1, and there will be a new set of participants_types (ids 5 to 8).

I'd thought of these seeds as a way to insert testing data so a new dev just getting started doesn't have to set up all of their own test data. We could create two different set of seeds - one for everything to set up base data like this (which preserves IDs to not break existing relationships) and a different set for test data.

src/database/seeds/01_Users.ts
```diff
@@ -9,7 +9,7 @@ type UserType = ModelObject<User>;
 const sampleData: Optional<UserType, 'id'>[] = [
   { email: 'test@example.com', location: 'Sydney, AU', name: 'Test User', phone: '+61298765432' },
   {
-    email: 'lionell.pack@thetradedesk.com',
```
lionell-pack-ttd, line 12: Why the change here? My email address is all over commits anyway.
If you want a fake email address, you should use @example.com - see https://www.rfc-editor.org/rfc/rfc2606 (it's guaranteed to never be a real address).

lionell-pack-ttd: Oh - I guess the delete-all wasn't work so the re-seed would probably fail. But someone could register @thetradedeskexample.com and create that email address, and possibly use that to log in to a test system - a low risk, but you should always use an invalid address like something @example.com if you're not using a real address.

src/database/seeds/02_Participants.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from 'knex';
```
lionell-pack-ttd, line 1: I'm not sure ordering seed files is a good idea - if a seed file depends on data in another, imagine the order of operations where Users is first:

1. First run starts
2. Insert user A
3. Insert participant Z which adds user A to the team
4. Second run
5. Seed tries to delete and reinsert user A
6. Error!

I don't think we should rely on ordering at this stage - and Knex creates seeds without a leading number, so this breaks the pattern. If there's a good reason, sure.

lionell-pack-ttd: Maybe that's a bad example because the relationship is the other way - but I think we should solve this when it's a problem, not now.

src/database/seeds/01_Users.ts
jingyi-gao-ttd: Yes, I was getting a lot of duplicates from re-seed, I will change it to del the sample data

src/api/api.ts
```diff
@@ -1,20 +1,21 @@
 import bodyParser from 'body-parser';
 import cors from 'cors';
 import express from 'express';
-import KeycloakConnect from 'keycloak-connect';
+// import { setupAdminClient } from './kcAdminClient';
```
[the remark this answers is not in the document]
jingyi-gao-ttd: Removed

PR #8, opened 2023-03-08
UID2-711 account setup 

jingyi-gao-ttd:
- New User Sign up

https://user-images.githubusercontent.com/116539741/223873981-246b5bb0-195f-4dc1-8a5e-e54130b987b7.mov

- User login with an account created but not signed

https://user-images.githubusercontent.com/116539741/223874031-4b71cb13-f5be-4a96-859c-04b3e0afd27a.mov

- User login with an account created and signed

https://user-images.githubusercontent.com/116539741/223874042-da64eb92-9840-4c93-ba9d-45a91fd224da.mov

- User login without an account

https://user-images.githubusercontent.com/116539741/223874572-585c8534-87ca-46fd-a5f5-66009705d000.mov


lionell-pack-ttd:
Should this PR be merging to main instead of to custom-login-flow?

lionell-pack-ttd:
If I simulate an error on submit of the Create Account form, there's no feedback showing me that a request is happening, and then I end up stuck with no feedback that anything went wrong.
image

src/api/usersRouter.ts
```diff
@@ -0,0 +1,50 @@
...
+  const data = UserScheme.parse(req.body);
+  const user = await User.query().insert(data);
+  res.status(201).json(user);
+});
+
+const emailParser = z.object({
+  email: z.string(),
+});
+
+usersRouter.get('/byEmail', async (req, res) => {
```
lionell-pack-ttd, line 22: Most APIs I've used put the field in the query-string - so a request might look something like `/api/users/?email=test@example.com`. Unless you're following a specific pattern that we might look at adopting, I think we're better off sticking with a standard pattern.

src/api/usersRouter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+usersRouter.get('/:userId', async (req, res) => {
+  const { userId } = userIdParser.parse(req.params);
+  const user = await User.query().findById(userId);
+  return res.status(200).json(user);
+});
+
+usersRouter.get('/:userId/participant', async (req, res) => {
+  const { userId } = userIdParser.parse(req.params);
+  const paritcipant = await User.relatedQuery('participant').for(userId);
```
lionell-pack-ttd, line 48: There's a spelling error - should be `participant`.

src/web/App.tsx
```diff
@@ -1,82 +1,46 @@
...
     else rootRef.current!.classList.remove('darkmode');
   };
-  const currentLocationDescription = menu.filter((m) => m.path === location.pathname)[0]
-    .description;
-
   if (!initialized) return <div>Loading...</div>;
+  const fullname =
+    LoggedInUser?.profile.firstName && LoggedInUser?.profile.lastName
+      ? `${LoggedInUser?.profile.firstName} ${LoggedInUser?.profile.lastName}`
+      : undefined;
```
lionell-pack-ttd, line 30: If we somehow have a firstName or lastName but not both, we still have a name to display - I don't think it needs to be undefined just because we're missing one.

src/web/components/Core/Notification.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+import * as stories from './Notification.stories';
+
+const { WithIcon, WithTitle, WithoutIconAndTitle } = composeStories(stories);
+
+test('has icon rendered', async () => {
+  render(<WithIcon />);
+});
+
+test('has title rendered', async () => {
+  // No actual way to tell because we don't load the image in Jest - this is to provide more examples of Jest using Storybook data
```
lionell-pack-ttd, line 13: This comment doesn't seem to belong here!

src/web/components/Core/PortalHeader.tsx
```diff
@@ -68,7 +72,21 @@ export function PortalHeader({ email, fullname, setDarkMode = undefined }: Porta
             </div>
           </DropdownMenuItem>
           <DropdownMenuSeparator className='separator' />
-          <DropdownMenuItem>Log out</DropdownMenuItem>
+          <DropdownMenuItem onSelect={(event) => event.preventDefault()}>
```
lionell-pack-ttd, line 75: It looks like you have a merge error here - you've got a duplicated dark-mode switch.

jingyi-gao-ttd:
> If I simulate an error on submit of the Create Account form, there's no feedback showing me that a request is happening, and then I end up stuck with no feedback that anything went wrong. image

I was plan to add form and input error in the same branch, but I found it is necessary to add sever error first. Just added to the branch

keycloak/themes/uid2-theme/login/login-verify-email_backup.ftl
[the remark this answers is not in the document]
jingyi-gao-ttd: oh sorry, removed

src/api/usersRouter.ts
jingyi-gao-ttd: I think you are right, updated

lionell-pack-ttd:
When trying to select a job function on the create form screen using the keyboard, there's no highlight - I can't tell which item will be chosen if I hit enter.
image

lionell-pack-ttd:
If I simulate a network error on submitting the create account form, now it doesn't just give me no error message, but it resets the form so I have to start again.
chrome_D1enNhyUiC

PR #12, opened 2023-03-15
Containerize the spa and api

amay-ttd:
Created a dockerfile to run the spa and api in the same container
Added git workflows


Dockerfile_ssportal
```diff
@@ -0,0 +1,23 @@
+FROM node:16-alpine
```
jingyi-gao-ttd, line 1: Just a question, I thought we are upgrading to node 18, why not use the node 18 image in the image?

amay-ttd: Good call, updated

PR #14, opened 2023-03-17
Migrate keycloak config to envars

amay-ttd:


src/api/envars.ts
```diff
@@ -0,0 +1,10 @@
+const errorMessage = 'Unable to get envar value';
+
+export const SSP_KK_AUDIENCE = process.env.SSP_KK_AUDENCE ?? errorMessage;
+export const SSP_KK_ISSUER_BASE_URL = process.env.SSP_KK_ISSUER_BASE_URL ?? errorMessage;
+export const SSP_KK_REALM = process.env.SSP_KK_REALM ?? errorMessage;
+export const SSP_KK_AUTH_SERVER_URL = process.env.SSP_KK_AUTH_SERVER_URL ?? errorMessage;
+export const SSP_KK_SSL_REQUIRED = process.env.SSP_KK_SSL_REQUIRED ?? errorMessage;
+export const SSP_KK_SSL_RESOURCE = process.env.SSP_KK_SSL_RESOURCE ?? errorMessage;
```
jingyi-gao-ttd, line 8: I wonder if we could rename the keycloak env names to make the name more clear? Currently, I need to look at the value and then understand that SSP_KK_AUDIENCE is the API client id and SSP_KK_SSL_RESOURCE is the SPA client id

amay-ttd: I'm using SSP_KK_ the the upper-cased key the value maps against? I'm trying to keep the convention for now, but it if we run into places where it doesn't work, it might be worth coming back to?

PR #15, opened 2023-03-17
Create keycloak image and build workflow

amay-ttd:


Dockerfile_keycloak
lionell-pack-ttd, line 1: We've got the sql server dockerfile in the sqlServer/ folder, would it make sense to have this file in the Keycloak folder? Or possibly the reverse - if it's helpful to have dockerfiles in the root, maybe we could move the sql server one.

lionell-pack-ttd (approved):
Other than my question/suggestion about the dockerfile folder location, lgtm.

Dockerfile_keycloak
amay-ttd: Dockerfiles will only have access to sub-directories of their current folder, so typically, it's the most flexible to have them in the root. The alternative is to structure the codebase around the containers, but I don't think that makes sense in our case. IMO whoever ends up standing up the DB should probably address it at that point.

PR #21, opened 2023-03-20
Form validation

jingyi-gao-ttd:
Stories:http://localhost:6006/?path=/story/inputs-checkbox--with-label
specs: 
image


package.json
```diff
@@ -27,6 +27,7 @@
     "@types/react-dom": "^18.0.9",
     "axios": "^1.3.1",
     "body-parser": "^1.20.1",
+    "classnames": "^2.3.2",
```
lionell-pack-ttd, line 30: We already have clsx installed, will that work rather than adding a new dependency?

src/web/components/Input/RadioInput.spec.tsx
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable testing-library/no-unnecessary-act */
```
lionell-pack-ttd, line 1: Don't just disable lint warnings for the whole file - this stops the linter from finding actual problems which might be introduced later.
It looks like the lint rule is wrong this time (something to do with validation in react-hook-form requires it) but the docs all warn that most of the time, it's telling you about something that's actually wrong with your test.

jingyi-gao-ttd: You are right, i found the issue is actually using await in front of the userEvent. I've removed the act

src/web/components/Core/Form.spec.tsx
```diff
@@ -23,27 +22,34 @@ describe('Form Component', () => {
 
     const textInput = screen.getByTestId('text-input');
 
-    await act(async () => {
-      await userEvent.type(textInput, 'New value');
-      await userEvent.click(screen.getByRole('radio', { name: 'No' }));
-      await userEvent.click(screen.getByRole('checkbox', { name: 'Checkbox 1' }));
+    userEvent.type(textInput, 'New value');
```
lionell-pack-ttd, line 25: If you just await the user events, you shouldn't then need to use all the waitFor blocks.

src/web/components/Input/RadioInput.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+describe('RadioInput', () => {
+  it('renders correctly with label', () => {
+    render(<WithLabel />);
+    expect(screen.getByText('Select an option')).toBeInTheDocument();
+    expect(screen.getByText('Option 1')).toBeInTheDocument();
+    expect(screen.getByText('Option 2')).toBeInTheDocument();
+    expect(screen.getByText('Option 3')).toBeInTheDocument();
+  });
+
+  it('displays validation error message', async () => {
```
lionell-pack-ttd, line 18: I'm noticing a lot of identical tests - to me this is a sign something isn't quite right. I'll drop you a DM with some more ideas on this.

src/web/components/Core/Form.spec.tsx
jingyi-gao-ttd: the await will definitely cause the following error. I've spent hours debugging this last night and found the only to resolve this errors was remove the `await` and use waitFor to make sure all updates are done.
 
```
console.error
    Warning: An update to null inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```

lionell-pack-ttd (approved):

PR #22, opened 2023-03-20
update realm during start up

jingyi-gao-ttd:
image


package.json
```diff
@@ -64,6 +64,7 @@
     "test": "craco test",
     "api": "nodemon -r dotenv/config ./src/api/api.ts",
     "build-api": "npx tsc --build ./tsconfig-api.json",
+    "reset-realm": "run-s keycloak:*",
```
lionell-pack-ttd, line 67: I had a look through the run-s docs and couldn't see anything about guaranteed ordering - it seems like it would be better to just list the two tasks, seeing as they need to be run in a specific order to work. `run-s` seems like it always does them in order, but ... alphabetic? File order?

lionell-pack-ttd (approved):

PR #25, opened 2023-03-29
uid2-794 invite team member

jingyi-gao-ttd:
**Required step**
- Re-import realm-export.json using scripts (This will recreate the realm)
**or** 
manually assign manage-users, query-users and view-users to self_serve_portal_apis
image 
- Regenerate the secret for self-serve-portal-apis

Screenshot 2023-03-29 at 10 51 33 am

- Replace the secret in .env

### DIFF
--- a/src/api/participantsRouter.ts
+++ b/src/api/participantsRouter.ts
@@ -5,7 +5,7 @@ import { Participant, ParticipantSchema } from './entities/Participant';
 import { UserRole } from './entities/User';
 import { getKcAdminClient } from './keycloakAdminClient';
 import { createNewUser, sendInviteEmail } from './services/kcUsersService';
-import { createUserInPortal } from './services/usersService';
+import { createUserInPortal, isUserBelongsToParticipant } from './services/usersService';
 
 export const participantsRouter = express.Router();
 participantsRouter.get('/', async (_req, res) => {
@@ -44,6 +44,13 @@ participantsRouter.post('/:participantId/invite', async (req, res) => {
     if (!(await Participant.query().findById(participantId))) {
       return res.status(404).send([{ message: 'Participant not exist' }]);
     }
+
+    if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, participantId))) {
+      return res
+        .status(401)
+        .send([{ message: 'You are not allow to invite user to this participant' }]);
+    }
+
     const { firstName, lastName, email, jobFunction } = invitationParser.parse(req.body);
     const kcAdminClient = await getKcAdminClient();
     const user = await createNewUser(kcAdminClient, firstName, lastName, email);
@@ -54,6 +61,6 @@ participantsRouter.post('/:participantId/invite', async (req, res) => {
     if (err instanceof z.ZodError) {
       return res.status(400).send(err.issues);
     }
-    return res.status(500).json('Something went wrong please try again');
+    return res.status(500).send([{ message: 'Something went wrong please try again' }]);
   }
 });
--- a/src/api/participantsRouter.ts
+++ b/src/api/participantsRouter.ts
@@ -42,13 +42,13 @@ participantsRouter.post('/:participantId/invite', async (req, res) => {
   try {
     const { participantId } = idParser.parse(req.params);
     if (!(await Participant.query().findById(participantId))) {
-      return res.status(404).send([{ message: 'Participant not exist' }]);
+      return res.status(404).send([{ message: 'The participant cannot be found.' }]);
     }
 
     if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, participantId))) {
       return res
         .status(401)
-        .send([{ message: 'You are not allow to invite user to this participant' }]);
+        .send([{ message: 'You do not have permission to make this invitation.' }]);
     }
 
     const { firstName, lastName, email, jobFunction } = invitationParser.parse(req.body);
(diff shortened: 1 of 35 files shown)


src/api/participantsRouter.ts
```diff
@@ -28,10 +31,29 @@ const idParser = z.object({
...
-  res.status(200).json(adminUsers);
+const invitationParser = z.object({
+  firstName: z.string(),
+  lastName: z.string(),
+  email: z.string(),
+  jobFunction: z.nativeEnum(UserRole),
+});
+
+participantsRouter.post('/:participantId/invite', async (req, res) => {
+  try {
```
lionell-pack-ttd, line 42: Do we have enough set up to check that the signed-in user belongs to that participant?

lionell-pack-ttd:
Will regenerating the secret be required for everyone? It probably needs more than just a PR comment. Maybe something in readme?

lionell-pack-ttd:
The dialog closes if you click anywhere outside it - that would be really frustrating if someone was halfway through! e.g. they might be copy-pasting the name or email from another program, and if they click on the wrong place they lose all progress.

src/web/App.scss
```diff
@@ -36,7 +36,7 @@
   }
 
   button {
-    border: 1px solid var(--theme-button-border);
```
lionell-pack-ttd, line 39: Setting the border to the main button color variable doesn't really make sense. Just change the --theme-button-border color to be the same.

src/web/components/Core/Dialog.scss
```diff
@@ -0,0 +1,68 @@
+.dialog-overlay {
+  background-color: rgba(255, 255, 255, 0.8);
+  position: fixed;
+  inset: 0;
+  animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.dialog-content {
+  background-color: white;
```
lionell-pack-ttd, line 9: Make sure you use theme variables rather than colours. --theme-background-content is probably fine here I guess.

src/web/components/Core/Dialog.scss
```diff
@@ -0,0 +1,68 @@
...
+  border-radius: 5px;
+  box-shadow: 0px 4px 14px rgba(0, 0, 0, 0.25);
```
lionell-pack-ttd, line 11: Again, use a theme colour. I took a look at Figma and it looks like we've got some (0, 0, 0, 0.15) (which is --theme-box-shadow) and some (0, 0, 0, 0.25). Maybe just set this to --theme-box-shadow and let's check with Kristen if she has a preference, but I expect we can make them all the same.

src/web/components/Core/Dialog.scss
```diff
@@ -0,0 +1,68 @@
...
+  height: auto;
+
+  .icon-button {
+    position: absolute;
+    top: 47px;
+    right: 47px;
+    background: none;
+  }
+
+  h2 {
```
lionell-pack-ttd, line 30: Radix usually recommends using a className and targeting that for styling - but if you've seen something about the dialog maybe this is fine?

src/web/components/Core/Dialog.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+    fireEvent.click(openButton);
+
+    expect(screen.getByText('Dialog Title')).toBeInTheDocument();
+    expect(screen.getByText('Dialog content goes here')).toBeInTheDocument();
+    expect(screen.getByText('Close')).toBeInTheDocument();
+  });
+
+  it('renders without title', () => {
+    render(<WithoutTitle />);
+    expect(screen.queryByText('Dialog Title')).not.toBeInTheDocument();
```
lionell-pack-ttd, line 21: This isn't really testing anything - there could be a big title with different text! There's no string "Dialog Title" in this story so this wouldn't find anything no matter what.

I think it's fine to just drop this test anyway.

src/web/configureFontAwesomeLibrary.ts
```diff
@@ -1,20 +1,20 @@
...
   faChevronUp,
   faEllipsisH,
   faPencil,
   faTrashCan,
+  faXmark,
 } from '@fortawesome/free-solid-svg-icons';
 
 export function configureFontAwesomeLibrary(): void {
   library.add(faCheck);
-  library.add(faCircleCheck);
```
lionell-pack-ttd, line 14: Why remove this? It looks like it's being used in accountPending.tsx.

src/web/contexts/ParticipantProvider.tsx
```diff
@@ -48,7 +50,9 @@ function ParticipantProvider({ children }: { children: ReactNode }) {
     [participant]
   );
   return (
-    <ParticipantContext.Provider value={participantContext}>{children}</ParticipantContext.Provider>
+    <ParticipantContext.Provider value={participantContext}>
+      {loading ? <Loading /> : children}
```
lionell-pack-ttd, line 54: I'm not sure about this pattern, but let's use it for now.

src/web/configureFontAwesomeLibrary.ts
jingyi-gao-ttd: we are using the regular icon rather than the solid icon in accountPending page. I tried add the faCircleCheck from regular lib to the library, but it didn't work. so I am using  `findIconDefinition({ prefix: 'far', iconName: 'circle-check' })` instead

src/web/screens/addTeamMemberDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+          rules={{ required: 'Please specify first name.' }}
+        />
+        <TextInput
+          inputName='lastName'
+          label='Last Name'
+          rules={{ required: 'Please specify last name.' }}
+        />
+        <TextInput
+          inputName='email'
+          label='email'
```
lionell-pack-ttd, line 47: Capitalize 'Email' for the label.

PR #28, opened 2023-04-06
Update database naming conventions to use camelCase instead of snakec…

lionell-pack-ttd:
…ase - this avoids a number of problems with the snakeCaseMapper.

NOTE: This PR _changes older database migrations_. This is a breaking change for databases! You will need to drop and recreate your uid2_selfserve database, and re-run the migrations!

### DIFF
--- a/src/api/entities/User.ts
+++ b/src/api/entities/User.ts
@@ -1,7 +1,8 @@
-import { Model } from 'objection';
+import { Model, ModelObject } from 'objection';
 import { z } from 'zod';
 
 import { BaseModel } from './BaseModel';
+import { ModelObjectOpt } from './ModelObjectOpt';
 
 export interface IUser {}
 export enum UserRole {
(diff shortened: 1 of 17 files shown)


src/api/entities/User.ts
```diff
@@ -32,14 +33,21 @@ export class User extends BaseModel {
...
   id!: number;
   email!: string;
-  location: string | undefined;
-  phone: string | undefined;
+  firstName!: string;
+  lastName!: string;
+  location?: string;
+  phone?: string;
   role!: UserRole;
+  participantId!: number;
```
Kshitij-Banerjee, line 41: participant Id is optional in the schema 
`participantId: z.number().optional()`

lionell-pack-ttd: Seems to need to be optional for the create account workflow. I won't dig into it for now.

jingyi-gao-ttd: I think when a user is created, it always has a participantId, maybe we should make this column notNullable in schema

PR #30, opened 2023-04-12
kba-UID2-840 -  Error Boundaries

Kshitij-Banerjee:
### What does this PR do ? 

Adds an Error Boundary to capture unexpected errors in the SPA

#### How it renders
image

### TODO
Re-do the design of this error page based on Design recommendations.


src/web/utils/errorHandler.tsx
```diff
@@ -0,0 +1,129 @@
...
+            errorId={errorId}
+            errorHash={errorHash}
+            errorCount={errorCount}
+            message={message}
+            stack={stack}
+            componentStack={componentStack}
+          />
+        );
+      }
+      return <React.Fragment key={errorCount}>{children}</React.Fragment>;
```
jingyi-gao-ttd, line 108: what is errorCount for? is it always be 0?

Kshitij-Banerjee: Yeah, this is unused right now. It's purpose is to re-mount the children if the user wants to retry any intermittent loading errors (user can retry intermittent failure in creatives). Let me remove it for now, and we can always add it back if design wants it that way.

src/web/components/Core/Dialog.tsx
```diff
@@ -31,7 +31,7 @@ export function Dialog({
       <RadixDialog.Content
         className='dialog-container'
         // To prevent dialog closed from user clicking outside of the dialog
-        onInteractOutside={(fireEvent) => fireEvent.preventDefault()}
+        onInteractOutside={(fireEvent) => fireEvent.preventDefault()} // eslint-disable-line @typescript-eslint/no-unsafe-call
```
lionell-pack-ttd, line 34: I'm not getting a lint rule violation either in VS Code or `npm run lint` - is this needed?

src/web/utils/errorHandler.tsx
```diff
@@ -0,0 +1,129 @@
...
+  [key: string]: unknown;
+}
+
+/** Create an Error Boundary around a given component.
+ * @param ErrorComponent - Custom error component used to display the error.
+ */
+export function errorHandler<P extends ErrorBoundaryPropType>(
+  ErrorComponent: React.ComponentType<RenderedErrorProps>
+) {
+  class ErrorHandlerWrapperComponent extends React.Component<P, ErrorState> {
```
lionell-pack-ttd, line 63: Is this borrowed from somewhere else? If so, can you document where it came from? I'd prefer to avoid class-based components but if it's saving some time, maybe it's reasonable.

src/web/utils/errorHandler.tsx
```diff
@@ -0,0 +1,129 @@
...
+    ErrorComponent: React.ComponentType<RenderedErrorProps>; // eslint-disable-line react/no-unused-class-component-methods
+    constructor(props: P) {
+      super(props);
+      this.state = {
+        errorId: uuidv4() as string, // eslint-disable-line @typescript-eslint/no-unsafe-call
```
lionell-pack-ttd, line 68: If I remove the eslint-disable I don't get any error (same for the similar line further down) - do you need to disable the lint rule?

src/web/utils/errorHandler.tsx
Kshitij-Banerjee: Yeah, I don't like class components as well, but sadly react error boundaries currently mandate it [here](https://legacy.reactjs.org/docs/error-boundaries.html#introducing-error-boundaries)
```
Only class components can be error boundaries. In practice, most of the time you’ll want to declare an error boundary component once and use it throughout your application.
```



(I implemented this recently in creatives, which in-turn is borrowed from ad-platform. Do you think its okay to reference ttd-codebase from here?)

src/web/components/Core/Dialog.tsx
Kshitij-Banerjee: oh yeah! I swear I was getting a big black overlay before, but can't seem to reproduce anymore. Removed ✅

src/web/utils/errorHandler.tsx
Kshitij-Banerjee: Remove ✅ 



I get this if I remove [the one below](https://github.com/IABTechLab/uid2-self-serve-portal/pull/30/files#diff-dc7164d51566178ee537ff6c919a07c2c613e4a220a7ae81fa73a4563ea8558dR73) though.
```Unused method or property "ErrorComponent" of class "ErrorHandlerWrapperComponent"eslint[react/no-unused-class-component-methods](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/no-unused-class-component-methods.md)```

I think its because linter can't figure out the dynamic calls that are made to the HOC.

File: src/web/utils/errorHandler.tsx (hunk 1 above)
Comment (Kshitij-Banerjee) [reply src/web/utils/errorHandler.tsx]: Removed ✅

File: src/web/utils/PortalErrorBoundary.tsx [hunk 5]
```diff
@@ -0,0 +1,15 @@
+import { errorHandler, ErrorView, RenderedErrorProps } from './errorHandler';
+
+function CreativeErrorComponent(props: RenderedErrorProps) {
```
jingyi-gao-ttd, line 3: I guess we could rename this?

Kshitij-Banerjee: oh yes, thank you! 🤦

src/web/utils/errorHandler.tsx
lionell-pack-ttd: Maybe don't mention the TTD codebase directly - I was mostly curious :)

lionell-pack-ttd: Open source contributors can't see that code so it's not really useful to have a reference to it.

lionell-pack-ttd (approved):

PR #32, opened 2023-04-13
UID2-807 Block account creation by free email provider

jingyi-gao-ttd:
JIRA ticket: https://atlassian.thetradedesk.com/jira/browse/UID2-807
This PR adds the error message and steps to enable blocking account creation. 
Notice: Declarative User Profile is Technology Preview and the settings is not included in the realm-export. Manual steps are required
The error message will be rendered like this:
Capture


README.md
```diff
@@ -106,6 +106,16 @@ docker compose restart keycloak
...
 
+### Enable blocking account creation by free email providers
+
+> Declarative User Profile is Technology Preview and is not fully supported.
+
+1. Click on the `Realm Settings` on the left side menu and turn on the `User Profile Enabled` switch.
+2. Once you enable it and click on the `Save` button, you can access the `User Profile` tab.
+3. Click on `email`, scroll down to `validators` and click `create validator`.
+4. Select `pattern` from the list and add `^(?!.*@(gmail|hotmail|yahoo)\.com$).+@.+\..+$` as pattern and `errorPatternNoMatch` as error message key
+5. Click on the `Save` button
```
Kshitij-Banerjee, line 117: Could this be done via the realm-export?

jingyi-gao-ttd: I've tried export realm settings, and so far only `userProfileEnabled` is included in the  realm-export.json, the validators will not be included, so only step 1 could be removed, but we still need to go to the keycloak admin portal to set up the validator. 

There is a JSON editor in the `User profile` tab, we could provide the user profile attributes JSON, but it will include all user profile attributes not only the email field. Please let me know which way would you prefer.
image

Kshitij-Banerjee: Can the user profile json automatically setup? Or will it have to be manually placed in the tab ?

(If it can't be automated, not much benefit in using the json then I think)

jingyi-gao-ttd: Yes, it has to be manually placed and saved. This feature is still under review, I will keep an eye on it, I hope Keycloak will provide some way to setup this automatically.

Kshitij-Banerjee: Sounds good 👍 

There is a PUT api for user profile on these [docs](https://www.keycloak.org/docs-api/15.0/rest-api/index.html) 
I'm not sure if this can achieve the automation that we want. 

Either case, its probably not worth the effort to dig this rabbit hole right now.

lionell-pack-ttd (approved):

PR #33, opened 2023-04-17
UID2-792-Refactor Form component and related Input components

jingyi-gao-ttd:
This PR contains a refactoring of the Form component and related Input components to improve usability and maintainability.

1. Provides a solution for handling nested children in the Form component. By using the React Context API, we can pass down the control and handleSubmit properties more easily to all children components, regardless of their nesting level. This approach provides a more flexible and maintainable way to work with the Form component and its child components.
2. Improves performance by avoiding unnecessary re-renders and context value updates.


src/web/components/Core/Form.tsx
```diff
@@ -1,23 +1,33 @@
...
+  customizeSubmit?: boolean;
 };
 
+export type FormContextType<T extends FieldValues> = {
+  control: Control<T>;
+  handleSubmit: (e?: React.BaseSyntheticEvent) => Promise<void>;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const FormContext = createContext<FormContextType<any> | null>(null);
```
lionell-pack-ttd, line 22: What did you try to get rid of the any here?

src/web/components/Input/withFormContext.tsx
lionell-pack-ttd, line 1: Couldn't all of this be replaced just by using `const { control } = useContext(FormContext);` in the controls themselves?

jingyi-gao-ttd: Do you mean in the input? I did this in the first commit, but still find a lot of redundant code.   I think the error message is necessary for each component, so I think wrap could be easier for maintain and adding new input components later

jingyi-gao-ttd: https://github.com/IABTechLab/uid2-self-serve-portal/pull/33/commits/28692e7534dcfeff7bc8ca0de45b1e0c1079dd69#diff-8091550f70c9a967056702d53e84f93a68f037f6781a36fa602ebf136ae80c51 

here is the first commit

src/web/components/Core/Form.tsx
jingyi-gao-ttd: As the generic type, `T extends FieldValues` is not defined outside of the Form component, I tried to use just `FieldValues` but I get the error:

Type '{ control: Control<T, any>; handleSubmit: (e?: React.BaseSyntheticEvent<object, any, any> | undefined) => Promise<void>; }' is not assignable to type 'FormContextType<FieldValues>'.

src/web/components/Input/withFormContext.tsx
lionell-pack-ttd: You're right that you shouldn't be repeating the error message in each component, but there's a much simpler approach - just provide a `useFormContext` function which returns the context, and throws the error if it's unavailable.

It means it's still obvious in the component what's happening - instead of calling `useContext(FormContext)` you just call `useFormContext()`. It's a common pattern and much less code than the higher-order component you've used.

lionell-pack-ttd (approved):

PR #40, opened 2023-05-03
Setup email service

jingyi-gao-ttd:
This PR is for setup an email service that could send email via sendGrid in production env and mailhog in dev env. It is using handlebars for email templating

here is the output of a dev env email.
image

Screenshot 2023-05-08 at 10 21 18 am


Kshitij-Banerjee:
@jingyi-gao-ttd  - There seems to be some problem with this PR, and probably needs a rebase

jingyi-gao-ttd:
@Kshitij-Banerjee  Thanks KB, I've updated this PR

PR #42, opened 2023-05-11
TAM email template

jingyi-gao-ttd:
image

image


src/api/services/participantsService.ts
```diff
@@ -0,0 +1,30 @@
...
+  const emailService = createEmailService();
+  const requestor = newParticipant.users![0];
+  const templateData = {
+    participant: newParticipant.name,
+    participantType: participantTypes.map((pt) => pt.typeName).join(', '),
+    requestor: `${requestor.firstName} ${requestor.lastName}`,
+    requestorEmail: requestor.email,
+    participantLocation: requestor.location,
+    jobFunction: requestor.role,
+    link: 'http://localhost:3000/',
```
Kshitij-Banerjee, line 21: can pull this from env ?

jingyi-gao-ttd: Yes, we can, this link here is just for testing as we haven't confirmed what this link should be, should I just leave it blank for now?

.github/workflows/sync-sendgripd-template.yaml
```diff
@@ -29,4 +32,4 @@ jobs:
         uses: actions-js/push@master
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: jyg-UID2-833-setup-sendgrid
+          branch: jyg-UID2-833-TAM_email_template
```
Kshitij-Banerjee, line 35: Does the branch have to be coded here ? Can it not be selected from the dropdown while running the workflow?

src/api/services/participantsService.ts
Kshitij-Banerjee: Yeah, or we can keep it as SSP_WEB_HOST, which is a sane default and it would at least lead them to home page

src/api/services/participantsService.ts
```diff
@@ -0,0 +1,30 @@
...
+  };
+  const emailArgs: EmailArgs = {
+    subject: 'New Participant Request',
+    templateData,
+    template: 'newParticipantReadyForReview',
+    to: { name: 'TAMs', email: 'jingyi.gao@thetradedesk.com' },
```
Kshitij-Banerjee, line 27: Where would this list come from ? Would this be a list, or a group alias ?

.github/workflows/sync-sendgripd-template.yaml
jingyi-gao-ttd: The workflow_dispatch does not work until this is merged into main. I will remove this

PR #43, opened 2023-05-11
kba-uid-1073 add database migration docker container

Kshitij-Banerjee:
### What does this MR do ? 

Creating a docker container that will run as a migration job, before the deployment of pod.
This will be published via github actions, similar to other container images.

### DIFF
--- a/Dockerfile_ssportal_migration
+++ b/Dockerfile_ssportal_migration
@@ -1,0 +1,18 @@
+FROM node:18-alpine
+ENV NODE_ENV production
+WORKDIR /usr/src/app
+
+# Copy the relevant files for migration neeeds
+COPY *.js *.ts *json .npmrc ./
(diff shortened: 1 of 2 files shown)


Dockerfile_ssportal_migration
```diff
@@ -0,0 +1,21 @@
...
+# Copy the relevant files for migration neeeds
+COPY *.js *.ts *json .npmrc ./
+COPY src/api/envars.ts ./src/api/
+COPY src/database/migrations/* ./src/database/migrations/
+
+# The image set ENV production, but we need the dev deps to build, so install with prod false
+RUN npm ci --production=false
+
+# Delete the status completion file if present
+RUN rm -f /var/run/migration-status/completed
```
jingyi-gao-ttd, line 14: just curious, why does this need to be deleted, is this file blocking the migration?

PR #45, opened 2023-05-16
Enable substituting the localhost values in keycloak docker

Kshitij-Banerjee:
### What does this PR do ? 

We use realms-export.json for setting up keycloak configurations, but currently it only has development setting by default. 

For integration and prod environments, urls will have to change. To enable this without having to build images that change for each environment, this PR uses ENV variables to modify the realms-export before running the container. 
This ensures we can re-use the same image across environments.

### DIFF
--- a/Dockerfile_keycloak
+++ b/Dockerfile_keycloak
@@ -2,8 +2,13 @@ FROM quay.io/keycloak/keycloak:20.0.3
 
 ENV KC_TRANSACTION_XA_ENABLED 'false'
 
-COPY keycloak/realm/ /opt/keycloak/data/import
+
+COPY keycloak/realm/realm-export.json /opt/keycloak/data/import/realm-export.json
 COPY keycloak/themes /opt/keycloak/themes
 
-ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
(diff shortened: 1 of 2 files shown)


Dockerfile_keycloak
```diff
@@ -2,8 +2,13 @@ FROM quay.io/keycloak/keycloak:20.0.3
...
 COPY keycloak/themes /opt/keycloak/themes
 
-ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
-CMD ["start-dev", "--import-realm", "--spi-theme-static-max-age=-1", "--spi-theme-cache-themes=false", "--spi-theme-cache-templates=false", "--spi-theme-welcome-theme=uid2-theme", "--health-enabled=true", "--metrics-enabled=true", "--features=declarative-user-profile"]
\ No newline at end of file
+USER root
+RUN chmod -R 775 /opt/keycloak/data/import
+USER keycloak
+
+ENTRYPOINT sed -i -E "s/http:\/\/localhost:[0-9]{4}/https:\/\/$SSP_WEB_HOST/g" opt/keycloak/data/import/realm-export.json && \
```
jingyi-gao-ttd, line 13: How about replace all `localhost:3000` in realm-export.json with SSP_WEB_HOST and set SSP_WEB_HOST=localhost:3000 in docker-compose? So that we don't need the regex anymore

lionell-pack-ttd (approved):

Dockerfile_keycloak
Kshitij-Banerjee: Yeah, can't do that because the realm-export is just for import into keycloak.
I'm making use of the env variable replacement being done by docker here instead.

PR #47, opened 2023-05-22
Sharing permission components

jingyi-gao-ttd:
This PR includes the components that are required for sharing permission page:
- The status popup component that display the result of the action
Screenshot 2023-05-23 at 2 32 53 pm

*We don't have design for info and error, the color here could be changed when we using this component in future
Screenshot 2023-05-23 at 2 33 01 pm
Screenshot 2023-05-23 at 2 32 58 pm

- Reusable table components that could display a list of participants, this table could check selected participants by default, and handle selected component change
Screenshot 2023-05-23 at 2 33 11 pm
Screenshot 2023-05-23 at 2 33 17 pm

- Search components that could filter participant by name and type

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/a196b297-a25b-44b3-931d-6eec37b8a6bf


src/web/components/Core/StatusPopup.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+export default {
+  title: 'Shared Components/StatusPopup',
+  component: StatusPopup,
+} as ComponentMeta<typeof StatusPopup>;
+
+const Template: ComponentStory<typeof StatusPopup> = (args) => <StatusPopup {...args} />;
+
+export const Success = Template.bind({});
+Success.args = {
+  message: 'Success! Your action was successful.',
```
Kshitij-Banerjee, line 14: Do we have a place to keep all the strings ? 
(might be helpful if we later want to do translations etc)

Kshitij-Banerjee:
Curious: Do we expect to have more complex use-cases for the table ?
Table implementations can get quite complicated with additional features, so wondering how long our custom implementation can last us

jingyi-gao-ttd:
@Kshitij-Banerjee Refers to the figma design, I can't find any other use cases for the table, I think the other usage for this table would be in the TAMs' approval list, but I don't think we will make it more complex

Kshitij-Banerjee (approved):

PR #48, opened 2023-05-22
Sharing permission table

jingyi-gao-ttd:
This PR creates sharing permission table which displays the checkable participants in table and provides a search bar to filter the participants in the table

-  When there are participants displayed in the table
Screenshot 2023-05-23 at 2 41 04 pm

-  When table is empty
Screenshot 2023-05-23 at 2 40 59 pm


src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+export function SharingPermissionsTable({ sharedParticipants }: SharingPermissionsTableProps) {
+  const [filter, setFilter] = useState('');
+  const [checkedParticipants, setCheckedParticipants] = useState<number[]>([]);
+  const [selectAll, setSelectAll] = useState(false);
+
+  const handleSelectAllChange = () => {
+    setSelectAll(!selectAll);
+    if (selectAll) {
+      setCheckedParticipants([]);
```
Kshitij-Banerjee, line 32: shouldn't this condition be reversed ? check all if selectAll and viceversa ?

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <img src='/group-icon.svg' alt='group-icon' />
+      <div className='no-participants-text'>
+        <h1>No Participants</h1>
+        <span>You don&apos;t have any sharing permissions yet.</span>
+      </div>
+    </div>
+  );
+}
+export function SharingPermissionsTable({ sharedParticipants }: SharingPermissionsTableProps) {
+  const [filter, setFilter] = useState('');
```
Kshitij-Banerjee, line 25: should we call this filterText to make it explicitly a noun ?

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: @Kshitij-Banerjee when the handleSelectAllChange is called, the selectAll is not updated yet. I agree this looks unclear. I will change this to use an useEffect

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: what do you mean by explicitly a noun?

Kshitij-Banerjee: The word filter is typically also used for a action/operation - like `.filter( x -> .. predicate )`
Here, it seems this is the text that needs to be used for a future filter operation - so adding the suffix would make it explicit and easier to read imo.

src/web/components/SharingPermission/SharingPermissionsTable.scss
```diff
@@ -0,0 +1,85 @@
...
+  border-bottom: 2px solid var(--theme-search-border);
+  padding: 6px 2px;
+  justify-content: space-between;
+  ::placeholder {
+    /* Chrome, Firefox, Opera, Safari 10.1+ */
+    color: var(--theme-search-border);
+    opacity: 1; /* Firefox */
+  }
+
+  :-ms-input-placeholder {
```
lionell-pack-ttd, line 27: Why the vendor-specific versions? Caniuse.com suggests this has been in everything for years now.

jingyi-gao-ttd: I will test in different browser and check if we could remove them

jingyi-gao-ttd: @lionell-pack-ttd I've tested it in edge and it works without this.

PR #52, opened 2023-05-24
Kba fix cors issues in keycloak

Kshitij-Banerjee:
### What does this MR do ? 

We are getting some CORS origin issues, because the settings are not properly set in the realm--export files.


Kshitij-Banerjee:
This is not needed, as the problem seems to be related to distributed caches in keycloak cluster mode.

PR #53, opened 2023-05-25
Account information page

jingyi-gao-ttd:
image


src/api/entities/Participant.ts
```diff
@@ -40,12 +40,16 @@ export class Participant extends BaseModel {
   id!: number;
   name!: string;
   status!: ParticipantStatus;
+  allowSharing!: boolean;
+  location?: string;
 }
 
 export const ParticipantSchema = z.object({
-  id: z.number().optional(),
+  id: z.number(),
```
lionell-pack-ttd, line 48: Making this required will break create participant. You've got to be cautious re-using these.

jingyi-gao-ttd: Oh, I didn't realise that! Thanks, I add a new partial schema for creation

src/api/participantsRouter.ts
```diff
@@ -1,21 +1,49 @@
...
+  next: NextFunction
+) => {
+  const { participantId } = idParser.parse(req.params);
+  const participant = await Participant.query().findById(participantId);
+  if (!participant) {
+    return res.status(404).send([{ message: 'The participant cannot be found.' }]);
+  }
+
+  if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, participantId))) {
+    return res.status(401).send([{ message: 'You do not have permission to update participant.' }]);
```
lionell-pack-ttd, line 32: This should be a 403. 401 is if the user has no/invalid credentials, and they should log in and try again. 403 is for a user who's logged in, but doesn't have permission to do the operation.

src/api/tests/participantsRoute.test.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+describe('Participants Route tests', () => {
+  const withToken = useTestServer();
+
+  describe('hasParticipantAccess middleware', () => {
+    afterEach(() => {
+      jest.restoreAllMocks();
+    });
+
+    it('should allow access to an authenticated user with permission', async () => {
```
lionell-pack-ttd, line 46: This test doesn't seem to be named properly, and I'm not sure what it's achieving. It's looking for a status code 400, so I guess it's testing that the validation doesn't allow an empty request, but it's also testing that the participant and user queries were called.

Decide what this should be testing, and then focus on testing the outcome you want to see.

src/api/tests/participantsRoute.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      expect(res.statusCode).toBe(400);
+    });
+
+    it('should deny access to an authenticated user without permission', async () => {
+      mockParticipant();
+      mockUser(null);
+      const req: Request = request(api).post('/api/participants/1/invite');
+      const res = await withToken(req);
+
+      expect(Participant.query).toHaveBeenCalledTimes(1);
```
lionell-pack-ttd, line 64: Why do you check for this? The test is already expecting that the response is a 401. There's no need to check that it happened a specific way.

src/api/tests/participantsRoute.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    it('should throw error when participant not exists', async () => {
+      mockParticipant(null);
+      mockUser();
+      const req: Request = request(api).post('/api/participants/10000/invite');
+      const res = await withToken(req);
+
+      expect(Participant.query).toHaveBeenCalledTimes(1);
+      expect(User.query).toHaveBeenCalledTimes(0);
```
lionell-pack-ttd, line 78: Why does it matter whether the user query was called or not? It doesn't seem important to the test.

src/database/migrations/20230418232949_AddAllowSharingAndLocationColumnsToParticipantsTable.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('participants', (table) => {
+    table.boolean('allowSharing').defaultTo(true);
```
lionell-pack-ttd, line 5: When there's a default column value, you usually also want to set that value for existing rows. I think that's probably a good idea here.

src/web/components/Core/Tooltip.scss
lionell-pack-ttd, line 1: As discussed in person, the styling here needs some updating to match the newest designs. Make sure the font is fixed 🙂

src/api/tests/participantsRoute.test.ts
jingyi-gao-ttd: This test is part of the `hasParticipantAccess` unit test, I want to test the user belongs to participant should go pass middleware. I am using an existing endpoint rather than create a new one. 
I will change 
` expect(res.statusCode).toBe(400);` to ` expect(res.statusCode).not.toBe(403);`

src/web/screens/accountInformation.scss
```diff
@@ -1,5 +1,5 @@
 .account-info-content {
-  margin-top: 3.75rem;
+  margin-top: 2rem;
```
lionell-pack-ttd, line 2: This should be in px.

src/web/components/Core/Tooltip.scss
```diff
@@ -5,9 +5,17 @@
...
-  color: var(--theme-info-content);
-  font-family: 'Inter';
+  color: var(--theme-button-text);
+  border-radius: 4px;
+  background-color: var(--theme-background-sidepanel);
+  padding: 10px;
   font-weight: 400;
-  font-size: 1rem;
-  line-height: 1.18rem;
+  font-size: 10px;
```
lionell-pack-ttd, line 13: Font size and line height should be in rem - we want them to scale if the user zooms or has a larger font set (e.g. if they have impaired vision and need larger text).

lionell-pack-ttd (approved):

PR #54, opened 2023-05-26
UID2 833 setup email service

jingyi-gao-ttd:
image


src/api/services/emailService.ts
```diff
@@ -0,0 +1,17 @@
+import { EmailArgs } from './emailTypes';
+import * as NodemailerService from './nodemailerService';
+import * as SendGridService from './sendGridService';
+
+export function createEmailService() {
+  const isProduction = process.env.NODE_ENV === 'production';
```
Kshitij-Banerjee, line 6: will integration environment still use sendgrid? (as node env is production in integ too)

src/api/services/nodemailerService.ts
```diff
@@ -0,0 +1,47 @@
+import nodemailer from 'nodemailer';
+import { Address } from 'nodemailer/lib/mailer';
+import SMTPTransport from 'nodemailer/lib/smtp-transport';
+import hbs from 'nodemailer-express-handlebars';
+
+import { EmailArgs, EmailData, UID2Sender } from './emailTypes';
+
+const transporter = nodemailer.createTransport({
+  host: 'localhost',
```
Kshitij-Banerjee, line 9: does this have to change in non-local environemnt?

src/api/services/sendGridService.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SSP_SEND_GRID_API_KEY } from '../envars';
+import _templateIdMapping from '../templateIdMapping.json';
+import { EmailArgs, UID2Sender } from './emailTypes';
+
+type TemplateIdMapping = {
+  [key: string]: string;
+};
+const templateIdMapping = _templateIdMapping as TemplateIdMapping;
+
+sgMail.setApiKey(SSP_SEND_GRID_API_KEY);
```
Kshitij-Banerjee, line 12: What happens if this is not set ? If this is mandatory, should we crash the app if this is not provided, or let it start silently, but add a warning?

src/api/envars.ts
```diff
@@ -26,3 +26,7 @@ export const SSP_DB_HOST = process.env.SSP_DB_HOST ?? 'localhost';
 export const SSP_DB_PORT = process.env.SSP_DB_PORT ? +process.env.SSP_DB_PORT : 11433;
 export const SSP_DB_USER = process.env.SSP_DB_USER ?? 'sa';
 export const SSP_DB_PASSWORD = process.env.SSP_DB_PASSWORD ?? 'D3velopmentP0';
+export const SSP_SEND_GRID_API_KEY = process.env.SSP_SEND_GRID_API_KEY ?? '';
```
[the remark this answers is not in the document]
jingyi-gao-ttd: Yes, once this get merged, I will raise another PR fro k8s deployment

src/api/services/emailService.ts
jingyi-gao-ttd: Yes, I think integration should use a different SendGrid account to test the email workflow

src/api/services/nodemailerService.ts
jingyi-gao-ttd: no, this nodemailerService is only used by local env

src/api/services/sendGridService.ts
jingyi-gao-ttd: This is mandatory, I will add the warning to it

src/api/templateIdMapping.json
lionell-pack-ttd, line 1: Why do we have all these template IDs that aren't related to the portal?

jingyi-gao-ttd: I think thoses are the template in the sendgrid, not sure if those are default ones or legacy ones, I don't have access to login and check

lionell-pack-ttd: If our key has access to all of those other templates (which aren't ours), are we certain that the sync job won't delete or change any other templates?

emailTemplates/newParticipantReadyForReview.hbs
```diff
@@ -0,0 +1,55 @@
...
+      height: 30px; } .content { margin-top: 70px; width:500px; display: inline-block; } .button {
+      border-radius: 5px; font-weight: 700; font-size: 18px; padding: 15px 120px 15px 120px; margin:
+      0 auto; background-color: #cdf200; color: #030a40; width: auto; text-align: center;
+      text-decoration: none; } .content-text { text-align: left; }
+    </style>
+
+  </head>
+  <body>
+    <div class='logo-container'>
+      <a href='http://localhost:3000/'><img
```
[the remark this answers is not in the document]
lionell-pack-ttd: If this is going to be addressed in a later PR once we have an image online in the CDN, can you add a comment to that ticket to make sure this spot is updated?

.github/workflows/sync-sendgripd-template.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: SendGrid sync
+        id: sendgrid_sync
+        uses: nanopx/action-sendgrid-sync@0.5.2
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          sendgridApiKey: ${{ secrets.SENDGRID_API_KEY }}
+          templatesDir: 'emailTemplates/'
+          forceSyncAll: '${{ github.event.inputs.forceSyncAll }}'
```
lionell-pack-ttd, line 22: What is forceSyncAll? When would we want to provide it? Could it update templates which are used by other teams?

emailTemplates/newParticipantReadyForReview.hbs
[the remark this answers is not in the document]
jingyi-gao-ttd: @lionell-pack-ttd I think this has been added to that ticket already

src/api/templateIdMapping.json
jingyi-gao-ttd: @lionell-pack-ttd  The sync job will only update the template that has the name that matches the file in our repo. So there is a chance that we overwrite their templates, but this sync job can't delete any templates. I think last time Andrei checked this API key is not used by other app, we could find out why those templates are in our SendGrid account, and if we could remove them?

.github/workflows/sync-sendgripd-template.yaml
jingyi-gao-ttd: We manually set it to true when we add a new template, otherwise this sync job only updates the template with existing name matched in sendgrid

lionell-pack-ttd: Maybe add something to the readme about when to send a forceSyncAll and how to do it.

lionell-pack-ttd (approved):

PR #55, opened 2023-05-26
TAM email template 

jingyi-gao-ttd:
image

image

### DIFF
--- a/emailTemplates/newParticipantReadyForReview.hbs
+++ b/emailTemplates/newParticipantReadyForReview.hbs
@@ -1,54 +1,55 @@
 <html>
   <head>
-    <meta http-equiv='Content-Type' content='text/html; charset=utf-8' />
-    <meta
-      name='viewport'
-      content='width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1'
-    /><!--[if !mso]><!-->
-    <meta http-equiv='X-UA-Compatible' content='IE=Edge' /><!--<![endif]-->
-    <!--[if (gte mso 9)|(IE)]>
-    <xml>
-    <o:OfficeDocumentSettings>
-    <o:AllowPNG/>
-    <o:PixelsPerInch>96</o:PixelsPerInch>
-    </o:OfficeDocumentSettings>
-    </xml>
-    <![endif]-->
-    <!--[if (gte mso 9)|(IE)]>
-    <style type="text/css">
-      body {width: 600px;margin: 0 auto;}
-      table {border-collapse: collapse;}
-      table, td {mso-table-lspace: 0pt;mso-table-rspace: 0pt;}
-      img {-ms-interpolation-mode: bicubic;}
-    </style>
-    <![endif]-->
+    <link rel='preconnect' href='https://fonts.googleapis.com' />
+    <link rel='preconnect' href='https://fonts.gstatic.com' crossorigin />
+    <link href='https://fonts.googleapis.com/css2?family=Inter&display=swap' rel='stylesheet' />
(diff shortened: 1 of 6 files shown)


emailTemplates/newParticipantReadyForReview.hbs
```diff
@@ -1,54 +1,55 @@
...
+      14px; line-height: 34px; } tbody tr td:first-child { width: 250px; min-width: 250px;
+      max-width: 250px; word-break: break-all; } .table-header:nth-child(3) { margin-top: 24px; }
+      .footer { margin-top: 32px; }
     </style>
-
   </head>
   <body>
-    <div class='logo-container'>
+    <div>
       <a href='http://localhost:3000/'><img
```
Kshitij-Banerjee, line 23: how would these localhost references be resolved in integration/prod environemtns?

src/api/participantsRouter.ts
```diff
@@ -23,6 +23,8 @@ participantsRouter.post('/', async (req, res) => {
       relate: true,
     });
 
+    const participantTypes = await ParticipantType.query().findByIds(data.types!.map((t) => t.id));
+    sendNewParticipantEmail(data, participantTypes);
```
Kshitij-Banerjee, line 27: what if the participant creation succeeds, but the email fails ? 
Would the catch block below say Unable to create participant?

src/api/services/participantsService.ts
```diff
@@ -0,0 +1,31 @@
...
+    jobFunction: requestor.role,
+    link: SSP_WEB_BASE_URL,
+  };
+  const emailArgs: EmailArgs = {
+    subject: 'New Participant Request',
+    templateData,
+    template: 'newParticipantReadyForReview',
+    to: { name: SSP_TAM_EMAIL_DISPLAY_NAME, email: SSP_TAM_EMAIL },
+  };
+  emailService.sendEmail(emailArgs);
```
Kshitij-Banerjee, line 30: should we catch any errors here, and send back a standardised error code - so that the caller can distinguish between email errors vs creational/querying errors?

emailTemplates/newParticipantReadyForReview.hbs
jingyi-gao-ttd: We have a separate ticket to update the logo in templates. I've add this file in the ticket description as well. https://atlassian.thetradedesk.com/jira/browse/UID2-985

src/api/services/participantsService.ts
jingyi-gao-ttd: I think this is a good idea. I will do this now. I think if the email does not go through we shouldn't block the user , am I right?

emailTemplates/newParticipantReadyForReview.hbs
Kshitij-Banerjee: I had moved this comment to the other PR.

src/api/services/participantsService.ts
Kshitij-Banerjee: I think so, Maybe we can ask Kimberly on this as well.

IMO From the user's perspective they haven't done anything wrong, and if they repeat - will our database even let them create again ? If they can't create again, then better for us to alert ourselves and take a manual reconciliation.

src/api/participantsRouter.ts
jingyi-gao-ttd: I've added error catch up in the other PR
https://github.com/IABTechLab/uid2-self-serve-portal/pull/54
image

lionell-pack-ttd (approved):

PR #58, opened 2023-06-09
Fix participants search bug

lionell-pack-ttd:
There was a bug in the CSS causing the td padding on the participants search bar to apply to all td elements site-wide. This refactors the file to make better use of SCSS nesting and removes some unnecessary vendor-specific rules.

### DIFF
--- a/src/web/components/SharingPermission/ParticipantSearchBar.scss
+++ b/src/web/components/SharingPermission/ParticipantSearchBar.scss
@@ -1,6 +1,6 @@
 .search-icon {
   color: var(--theme-search-border);
-  height: 23px;
+  height: 1.4375rem;
 }
 
 .search-bar-input-container {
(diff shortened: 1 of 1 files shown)


jingyi-gao-ttd (approved):

src/web/components/SharingPermission/ParticipantSearchBar.scss
```diff
@@ -48,36 +38,36 @@
...
 
-.search-bar-participants th,
-td {
-  background: none !important;
-  padding: 0 0 1rem 0 !important;
-}
+  th,
+  td {
+    background: none !important;
+    padding: 0 0 1rem !important;
```
jingyi-gao-ttd, line 56: I guess we could update padding here to use px?

lionell-pack-ttd: I've updated all the sizing to use px for layout and rem for fonts.

PR #60, opened 2023-06-13
Update sharing permission page styling

jingyi-gao-ttd:
- Update rem to px for padding and margins
- Fix the trigger of tooltip
- Update the button on sharing permission page
sharing_permission
- Update sharing permission table to show the delete permissions button when the participant is selected
Screenshot 2023-06-15 at 10 46 30 pm
Screenshot 2023-06-15 at 10 46 35 pm

- Add intermediate state to selectAll checkbox 

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/0f74436e-24f7-4be1-a61d-f599b3e84a8a

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/eb49864e-c4c9-4faa-8aaa-03c7f73769b8


src/web/components/Core/SelectAllCheckbox.scss
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  // Override the style of .app button
+  padding: 0 !important;
+  border: 0 !important;
+}
+
+.select-all-checkbox.uncheck {
```
lionell-pack-ttd, line 14: I'd like to see this done using some of the available SCSS features - you could nest this inside the previous block, probably using the parent selector: https://sass-lang.com/documentation/style-rules/parent-selector

src/web/components/Core/SelectAllCheckbox.scss
```diff
@@ -0,0 +1,21 @@
+.select-all-checkbox {
+  background-color: var(--theme-radio-background) !important;
```
lionell-pack-ttd, line 2: Try to avoid using !important - sometimes it's needed, but you might just need to up the specificity of the selector. With SCSS sometimes the best way is to just nest the whole thing inside a selector for the class of the component - see e.g. Dialog.scss and its use of `.dialog-container`.

src/web/components/Core/StatusPopup.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .status-popup.Error {
-  border-color: darkred;
+  border-color: red;
```
lionell-pack-ttd, line 19: Make sure all colours are coming from `themes.scss`.

src/web/components/Core/SelectAllCheckbox.scss
jingyi-gao-ttd: I've updated all scss invovled in this PR. will udpate the rest when working on styling change ticket

src/web/components/Core/SelectAllCheckbox.tsx
lionell-pack-ttd, line 1: What about this component makes it a select-all checkbox? It just seems to be a regular checkbox.

lionell-pack-ttd: ... except worse, because it can only operate in controlled mode - it doesn't work unless the parent component handles all the events.

src/web/components/Core/SelectAllCheckbox.stories.tsx
lionell-pack-ttd, line 1: None of these stories actually let you do anything with the components - this is a sign to me that the component itself isn't really providing any value.

src/web/components/SharingPermission/ParticipantSearchBar.spec.tsx
```diff
@@ -19,17 +19,20 @@ describe('ParticipantSearchBar', () => {
...
+    ] = screen.getAllByRole('checkbox');
 
     // The label text assumes that you are using the participant's name as the label for the checkbox.
     // If it's different, please update it accordingly.
-    expect(participant1Checkbox).toBeChecked();
-    expect(participant2Checkbox).not.toBeChecked();
-    expect(participant3Checkbox).toBeChecked();
-    expect(participant4Checkbox).not.toBeChecked();
+
+    expect(participant1Checkbox).toHaveAttribute('data-state', 'checked');
```
lionell-pack-ttd, line 33: Why are you changing from checking the value to checking a particular data-state? Is there some benefit to this?

jingyi-gao-ttd: to support the indeterminate state of the checkbox, I am now using the checkbox from radix-ui, which is implemented by a button rather than input, thus it does not support toBeChecked

src/web/components/Core/SelectAllCheckbox.tsx
jingyi-gao-ttd: I was trying to build a component that can have indeterminate state, maybe selectAllCheckbox is not the current name, the stories is to display the styling of three state

src/web/components/SharingPermission/SharingPermissionsTable.scss
```diff
@@ -1,74 +1,93 @@
...
-  font-size: 12px;
-  line-height: 15px;
-}
+  .sharing-permissions-search-bar {
+    width: 100%;
+    border: none;
+    outline: none;
+    color: var(--theme-search-border);
+    font-weight: 400;
+    font-size: 12px;
```
lionell-pack-ttd, line 47: This is a good chance to clean this font-size and line-height up to rem.

jingyi-gao-ttd: Yea, you are right, I updated height and margin, font-size and line-height to the scss files that involved in this PR

src/web/components/SharingPermission/ParticipantSearchBar.scss
```diff
@@ -1,6 +1,6 @@
 .search-icon {
   color: var(--theme-search-border);
-  height: 1.4375rem;
+  height: 23px;
```
lionell-pack-ttd, line 3: This probably should stay as rem - icons are usually pretty much the same as fonts, they should scale with the base font size. If the font size was double but the icon didn't also get bigger, it would probably look wrong.

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -21,64 +24,69 @@ function NoParticipant() {
...
+export function SharingPermissionsTable({
+  sharedParticipants,
+  onDeleteSharingPermission,
+  children,
+}: SharingPermissionsTableProps) {
   const [filterText, setFilterText] = useState('');
-  const [checkedParticipants, setCheckedParticipants] = useState<number[]>([]);
-  const [selectAll, setSelectAll] = useState(false);
+  const [checkedParticipants, setCheckedParticipants] = useState<Set<number>>(new Set());
+  const hasParticipantSelected = useMemo(() => checkedParticipants.size > 0, [checkedParticipants]);
```
lionell-pack-ttd, line 34: Do you need a memo here? Checking `.size>0` is a very cheap operation, and the overhead of the memo is probably much higher.

lionell-pack-ttd (approved):

lionell-pack-ttd (approved):
Just fix up the build error and I think this is good to go.

PR #62, opened 2023-06-19
UID2-835 Sharing relationship backend

jingyi-gao-ttd:
https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/5319b5ab-6fc1-475a-bc19-39528eefadbe

Screenshot 2023-06-19 at 11 53 49 am


src/web/components/Core/StatusPopup.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .status-popup.Error {
-  border-color: darkred;
+  border-color: red;
```
jingyi-gao-ttd, line 19: we don't have a design for that, I just use red for now

src/api/services/adminServiceClient.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+});
+
+export const getSharingList = async (siteId: number): Promise<SharingListResponse> => {
+  try {
+    const response = await adminServiceClient.get(`/api/sharing/list/${siteId}`);
+    return response.data;
+  } catch (error: unknown) {
+    // if site don't have ACLs, currently it returns 404
+    if (error instanceof AxiosError && error.response?.status === 404) {
```
lionell-pack-ttd, line 27: Probably a better way would be to define 404 as a non-error status, so you only get an exception if something goes wrong.

The default axios `validateStatus` handler is:
```
validateStatus: (status) => {
  return status >= 200 && status < 300; // default
},
```

You could easily update that to also allow 404 as well as anything 2xx. Relevant docs: https://axios-http.com/docs/handling_errors

src/api/participantsRouter.ts
```diff
@@ -39,7 +50,7 @@ export const hasParticipantAccess = async (
 };
 
 participantsRouter.get('/', async (_req, res) => {
-  const participants = await Participant.query().withGraphFetched('types');
+  const participants = await Participant.query().whereNotNull('siteId').withGraphFetched('types');
```
lionell-pack-ttd, line 53: A get to `/participants` should return all participants - this adds an unexpected filter. We still might want to fetch all participants, including those with no site id.

Adding a URL parameter is a common way to filter a get request - so something like /participants/?withSiteId=true

jingyi-gao-ttd: Coz if a participant is without a site id, it means those participants are not yet approved by TAM, which I think the external user should have access to, if we make this endpoint to  /participants/?withSiteId=true, that means theoretically any user is able to retrieve those un-approved participants.

My initial thought was to add a middleware to check the user permission, if a user is a TAM, this endpoint will return all participants, otherwise return the participant with site_id, but since we haven't implement that permission, so I return all participants with site_id for now

src/api/participantsRouter.ts
```diff
@@ -111,3 +122,76 @@ participantsRouter.put(
...
+    if (!participant?.siteId) {
+      return res.status(400).send('Site id is not set');
+    }
+    const { newParticipantSites } = sharingRelationParser.parse(req.body);
+    const sharingParticipants = await addSharingParticipants(
+      participant.siteId,
+      newParticipantSites
+    );
+
+    const currentUser = await findUserByEmail(req.auth?.payload?.email as string);
```
lionell-pack-ttd, line 157: With an audit trail, always insert the audit entry in a way that if the audit insert fails, the operation doesn't happen. So await the insert first, and then submit the update to the admin service.

It's probably a good idea to then update the audit entry afterwards to indicate whether the update succeeded or failed.

src/api/participantsRouter.ts
```diff
@@ -111,3 +122,76 @@ participantsRouter.put(
...
+      return res.status(400).send('Site id is not set');
+    }
+    const { sharingSitesToRemove } = removeSharingRelationParser.parse(req.body);
+    const sharingParticipants = await deleteSharingParticipants(
+      participant.siteId,
+      sharingSitesToRemove
+    );
+
+    const currentUser = await findUserByEmail(req.auth?.payload?.email as string);
+    await insertSharingAuditTrails(
```
lionell-pack-ttd, line 188: Same here - audit entry first.

src/database/migrations/20230613000210_AddSiteIdToParticipantsTableAndCreateSharingAuditTrailTable.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('participants', (table) => {
+    table.integer('siteId').unique();
+  });
+
+  await knex.schema.createTable('sharingAuditTrails', (table) => {
+    table.increments('id').primary();
+    table.integer('participantId').references('participants.id').onDelete('CASCADE');
+    table.integer('userId').references('users.id').onDelete('CASCADE');
```
lionell-pack-ttd, line 11: We can *never* delete an audit entry, so these cascading deletes aren't OK. We probably won't ever delete participants, so that relationship is probably fine. But we might delete users, so I think we should store the email address as well, and not enforce the reference to the users table.

lionell-pack-ttd (approved):

PR #71, opened 2023-06-26
UID2-1348 fix click on tooltip cause error and sharing permission page

jingyi-gao-ttd:
- Fix the form is submitted when clicking tooltip
- Fix the font-size and title of sharing permission table
- Update the style of the disabled button
- Add outlines for search bar
- Remove display of `0 participant` when no participant is selected and captialize the P in `Participants` 
- Hide placeholder when search bar clicked
- Remove "Participant Not Appearing in Search?
- Hide checkbox in the table header when participant is shared
- Fix styling of delete permission popup

### DIFF
--- a/src/web/components/Core/Tooltip.tsx
+++ b/src/web/components/Core/Tooltip.tsx
@@ -13,10 +13,7 @@ export function Tooltip({ children, side, align }: TooltipProps) {
   return (
     <TooltipPrimitive.Provider>
       <TooltipPrimitive.Root>
-        <TooltipPrimitive.Trigger
-          className='tooltip-trigger'
-          onClick={(event) => event.preventDefault()}
-        >
+        <TooltipPrimitive.Trigger className='tooltip-trigger' type='button'>
           <FontAwesomeIcon icon='circle-info' data-testid='tooltip-info-icon' />
         </TooltipPrimitive.Trigger>
         <TooltipPrimitive.Content side={side} align={align} className='tooltip-content'>
--- a/src/web/components/Core/Tooltip.tsx
+++ b/src/web/components/Core/Tooltip.tsx
@@ -13,7 +13,10 @@ export function Tooltip({ children, side, align }: TooltipProps) {
   return (
     <TooltipPrimitive.Provider>
       <TooltipPrimitive.Root>
-        <TooltipPrimitive.Trigger className='tooltip-trigger'>
+        <TooltipPrimitive.Trigger
+          className='tooltip-trigger'
+          onClick={(event) => event.preventDefault()}
+        >
           <FontAwesomeIcon icon='circle-info' data-testid='tooltip-info-icon' />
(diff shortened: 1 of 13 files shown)


src/web/components/Core/Tooltip.tsx
```diff
@@ -13,7 +13,10 @@ export function Tooltip({ children, side, align }: TooltipProps) {
   return (
     <TooltipPrimitive.Provider>
       <TooltipPrimitive.Root>
-        <TooltipPrimitive.Trigger className='tooltip-trigger'>
+        <TooltipPrimitive.Trigger
+          className='tooltip-trigger'
+          onClick={(event) => event.preventDefault()}
```
lionell-pack-ttd, line 18: I think instead of leaving this as a 'submit' button (the default button type) and then preventing the submit, better to just use type='button' so it doesn't try to submit at all.

src/web/components/SharingPermission/SharingPermissionsTable.scss
```diff
@@ -6,12 +6,12 @@
     justify-content: space-between;
     align-items: center;
     margin-bottom: 16px;
-  }
 
-  .sharing-permissions-table h1 {
-    font-weight: 600;
-    font-size: 1.125rem;
-    line-height: 1.375rem;
+    h1 {
```
lionell-pack-ttd, line 10: As discussed - make the element a h2 instead and update the size of h2 across the site (it's only used in one other place which should also be 1.125rem).

src/web/styles/buttons.scss
```diff
@@ -13,8 +13,9 @@ button {
 }
 
 .primary-button:disabled {
-  background-color: var(--theme-background-sidepanel);
-  border-color: var(--theme-background-sidepanel);
+  background-color: var(--theme-table-border);
```
lionell-pack-ttd, line 16: Someone working on a theme would probably be surprised to find that when they change the table border color, it also changes the disabled button color.
I think a `theme-button-disabled` variable would make more sense, even though they're the same color value.

src/web/components/SharingPermission/ParticipantSearchBar.scss
```diff
@@ -19,6 +19,11 @@
 .clicked .search-bar-input-container {
   background: var(--theme-search-bar-background);
   border-radius: 10px 10px 0 0;
+  border: 2px solid var(--theme-search-border);
```
lionell-pack-ttd, line 22: The design looks like it needs the border around the dropdown as well, and Kristen mentioned adding it to both. It's probably good to fix them both together :)

lionell-pack-ttd:
One more thing - the delete permission popup still doesn't look quite right. The X button to close the dialog is uncomfortably close to the title.
image

Take a look at Figma - there needs to be more separation. The same problem is there in the dialog to add sharing permissions.

lionell-pack-ttd (approved):

PR #72, opened 2023-06-27
Remove checkbox and fix the dialog size on large screen

jingyi-gao-ttd:
image

### DIFF
--- a/src/web/screens/addTeamMemberDialog.scss
+++ b/src/web/screens/addTeamMemberDialog.scss
@@ -1,3 +1,0 @@
-.add-team-member-dialog .dialog-content {
-  flex-grow: 1;
-}
--- a/src/web/screens/addTeamMemberDialog.scss
+++ b/src/web/screens/addTeamMemberDialog.scss
@@ -1,9 +1,3 @@
-.addTeamMemberDialog {
-  max-width: 510px;
-}
-
-@media (min-width: 606px) {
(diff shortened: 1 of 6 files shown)


src/web/styles/buttons.scss
```diff
@@ -22,8 +22,8 @@ button {
 }
 
 .small-button {
-  padding: 0.5rem;
-  font-size: 1rem;
+  padding: 0.75rem;
+  font-size: 0.75rem;
```
lionell-pack-ttd, line 26: Kristen provided this CSS, but it looks too small to me and doesn't match Figma. Let's merge it as-is but then ask her to see if it looks right. Based on Figma the font-size should be 0.9375rem.

jingyi-gao-ttd: Yes I agree, the text looks really small. I've changed it back to 1

src/web/screens/addTeamMemberDialog.scss
```diff
@@ -1,3 +1,9 @@
 .addTeamMemberDialog {
   max-width: 510px;
 }
+
+@media (min-width: 606px) {
```
lionell-pack-ttd, line 5: See my PM - this makes the dialog too narrow once the window is less than 606px, we ideally want to make it full-width for any smaller window, and stop it from growing once it reaches the target width.

lionell-pack-ttd (approved):

PR #75, opened 2023-06-30
Llp uid2 1296 terms of service

lionell-pack-ttd:
Removes location from create account.
Removes the global line-height which was causing problems.
Adds terms of service text and removes it from the dialog wrapper.


lionell-pack-ttd:
image

jingyi-gao-ttd (approved):

PR #76, opened 2023-06-30
Llp uid2 1296 terms in create account

lionell-pack-ttd:
Require new users to accept the terms and conditions when creating a new participant.


.eslintrc.js
```diff
@@ -144,6 +144,7 @@ module.exports = {
       files: ['*.tsx', '*.ts'],
       rules: {
         'react/require-default-props': 'off',
+        'react/destructuring-assignment': 'off',
```
lionell-pack-ttd, line 147: I found this doesn't seem to have any quality benefits, and I ended up needing to write less readable code. May as well just disable it.

src/web/components/Core/Form.tsx
```diff
@@ -15,6 +22,11 @@ type FormProps<T extends FieldValues> = {
   id?: string;
 };
 
+export function withoutRef<TFieldName extends string>(props: UseFormRegisterReturn<TFieldName>) {
```
lionell-pack-ttd, line 25: Helper for `register` - it always provides `ref` but not all form-related components need it, and passing it unnecessarily requires forwardRef (which breaks TS types).

src/web/components/Core/TermsAndConditions.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  .terms-container {
+    border: 1px solid var(--theme-border);
+    border-radius: 5px;
+    padding: 16px;
+    height: 468px;
+    overflow: scroll;
+    margin-bottom: 22px;
+    max-width: 910px;
```
lionell-pack-ttd, line 17: This needs to be wider than the Figma design to look right - we might end up narrowing it later but we'll see.

src/web/components/Account/CreateAccountForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { FormStyledCheckbox } from '../Input/StyledCheckbox';
+import { TextInput } from '../Input/TextInput';
+
+export type CreateAccountFormProps = {
+  resolvedParticipantTypes: ParticipantTypeDTO[];
+  onSubmit: (data: CreateParticipantForm) => Promise<string[] | void>;
+};
+export function CreateAccountForm({ resolvedParticipantTypes, onSubmit }: CreateAccountFormProps) {
+  const [showTermsDialog, setShowTermsDialog] = useState(false);
+  const formMethods = useForm<CreateParticipantForm>({
```
lionell-pack-ttd, line 22: I had to switch to the hook-based `useForm` method for a few things, including `watch` (needed to disable the submit button until the terms have been read - see Figma).

src/web/components/Input/Input.scss
```diff
@@ -25,7 +25,6 @@
 }
 
 .option-label {
-  padding-left: 0.5rem;
```
lionell-pack-ttd, line 28: I moved this to right-margin on the checkbox instead so it would work better with custom-provided labels.

src/web/screens/createAccount.tsx
```diff
@@ -20,85 +13,38 @@ import { PortalRoute } from './routeUtils';
...
-    await CreateParticipant(formData, LoggedInUser!.profile);
+  const { participantTypes } = useLoaderData() as { participantTypes: ParticipantType[] };
+  const onSubmit = async (data: CreateParticipantForm) => {
+    const createResult = await CreateParticipant(data, LoggedInUser!.profile);
+    if ('errorStatus' in createResult) {
+      return createResult.messages;
+    }
     await loadUser();
-    navigate('/account/pending');
+    return navigate('/account/pending');
```
jingyi-gao-ttd, line 26: what's the reason of return the navigate here?

jingyi-gao-ttd (approved):

src/web/screens/createAccount.tsx
lionell-pack-ttd: It's just returning void to satisfy the linter - I'll update the lint rule instead, it doesn't really make sense.

PR #79, opened 2023-07-06
Add current participant middleware

jingyi-gao-ttd:
- Add the enrich participant middleware to  `/api/participants/current/*`
- Fix the queryMock so that it will return undefined rather than query db


src/api/services/participantsService.ts
```diff
@@ -75,3 +81,44 @@ export const deleteSharingParticipants = async (
...
+  next: NextFunction
+) => {
+  const { participantId } = idParser.parse(req.params);
+  const participant = await Participant.query().findById(participantId);
+  if (!participant) {
+    return res.status(404).send([{ message: 'The participant cannot be found.' }]);
+  }
+
+  if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, participantId))) {
+    return res.status(403).send([{ message: 'You do not have permission to update participant.' }]);
```
lionell-pack-ttd, line 101: Minor: Now that this is an enricher, the 403 should probably not include the word 'update' because it might just be a read. So maybe change to "You do not have permission to that participant."

src/api/participantsRouter.ts
```diff
@@ -204,3 +183,6 @@ participantsRouter.post(
     return res.status(200).json(sharingParticipants);
   }
 );
+
+export const currentParticipantRouter = express.Router();
+participantsRouter.use('/current', enrichCurrentParticipant, currentParticipantRouter);
```
lionell-pack-ttd, line 188: This breaks the convention we have everywhere else, where each endpoint declares the full route under that router type, and uses route params for parts of the REST path (e.g. the participant to query).

Is there any reason this couldn't follow the same convention (and be part of the same router), but with a route parameter that will parse as a string (and expect 'current')? It would fit in with everything else much better.

jingyi-gao-ttd: Oh I think I misunderstood the conversation before. I thought endpoint like teammember and business contact will have a separate endpoint like `/participants/current/businessContact`. I think what you said make sense, which will make all current endpoint works without adding more endpoints. I will make a change to that

src/api/tests/participantsService.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { participantsRouter } from '../participantsRouter';
+import { ParticipantRequest } from '../services/participantsService';
+import { mockParticipant, mockUser } from './queryMocks';
+import useTestServer, { api } from './utils';
+
+describe('Participant Service Tests', () => {
+  const withToken = useTestServer();
+
+  beforeEach(() => {
```
lionell-pack-ttd, line 11: This should be beforeAll - with beforeEach it adds the route over and over so there end up being many identical routes.

lionell-pack-ttd (approved):

PR #85, opened 2023-07-13
UID2 1296 terms of service links

lionell-pack-ttd:


src/web/components/Core/TermsAndConditions.tsx
```diff
@@ -11,233 +237,16 @@ export type AcceptTermForm = {
   acceptConditions: boolean;
 };
 
-export function TermsAndConditions({ onAccept, onCancel }: TermsAndConditionsProps) {
+export function TermsAndConditionsForm({ onAccept, onCancel }: TermsAndConditionsFormProps) {
```
jingyi-gao-ttd, line 240: why do we call this a Form? I didn't see any submit action?

jingyi-gao-ttd (approved):

src/web/components/Core/TermsAndConditions.tsx
lionell-pack-ttd: Because there's a button at the bottom - but it's called Accept, not Submit.

PR #86, opened 2023-07-14
UID2-1458 Fix home page error when first login

jingyi-gao-ttd:
- Fix backend return 500 when the access token is invalid
- Fix home page send request before token is refreshed

### DIFF
--- a/src/web/screens/home.tsx
+++ b/src/web/screens/home.tsx
@@ -51,8 +51,4 @@ export const HomeRoute: PortalRoute = {
   path: '/',
   description: 'Home',
   element: <Home />,
-  // loader: () => {
-  //   const sharingPermissions = GetSharingParticipants();
-  //   return defer({ sharingPermissions });
-  // },
 };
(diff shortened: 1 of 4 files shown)


src/web/screens/home.tsx
```diff
@@ -1,37 +1,47 @@
...
   const { LoggedInUser } = useContext(CurrentUserContext);
-  const { sharingPermissions } = useLoaderData() as {
-    sharingPermissions: ParticipantResponse[];
-  };
+  const [loading, setIsLoading] = useState<boolean>(false);
+  const [sharingPermissionsCount, setSharingPermissionsCount] = useState<number>(0);
+  const [hasError, setHasError] = useState<boolean>(false);
+  useEffect(() => {
+    const getSharingParticipantsCount = async () => {
+      setIsLoading(true);
```
lionell-pack-ttd, line 20: I can see a flicker of "0" before loading is set to true - wouldn't it make more sense to start loading as true?

lionell-pack-ttd (approved):

PR #87, opened 2023-07-14
Require the user to accept the terms of service before using the site.

lionell-pack-ttd:


src/api/usersRouter.ts
```diff
@@ -29,42 +34,38 @@ usersRouter.post('/', async (req, res) => {
   res.status(201).json(user);
 });
 
-const userIdParser = z.object({
-  userId: z.string(),
+usersRouter.put('/current/acceptTerms', enrichCurrentUser, async (req: UserRequest, res) => {
```
jingyi-gao-ttd, line 37: I think our convention was to to use the same router as /:userId and parse the param to be string `current`?

lionell-pack-ttd: That's what we did for participant, but that wouldn't make sense here - I don't want this route to work for anything except the current user. A user can't accept the terms and conditions for a different user.

jingyi-gao-ttd (approved):

PR #91, opened 2023-07-18
UID2-1343 allow user to edit and delete team members

jingyi-gao-ttd:
1. Handle update and delete error inline

- Business contact table

Screenshot 2023-07-18 at 4 47 24 pm

- Team member table

Screenshot 2023-07-18 at 4 47 34 pm

2. Add endpoints to update and delete team member

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/b763242a-d162-4b9c-93c3-35e5e6b5cea1


src/api/usersRouter.ts
```diff
@@ -62,15 +62,33 @@ usersRouter.post('/:userId/resendInvitation', async (req: UserRequest, res) => {
...
 
     logger.info(`Resending invitation email for ${req.user?.email}, keycloak ID ${user[0].id}`);
     await sendInviteEmail(kcAdminClient, user[0]);
     return res.sendStatus(200);
-});
+  });
+
+  usersRouter.delete('/:userId', async (req: UserRequest, res) => {
+    const { user } = req;
+    await user!.$query().delete();
```
lionell-pack-ttd, line 76: I think this should have some protection to prevent getting an account into a bad state - probably block a user from deleting themselves. That way a participant can never completely lock themselves out.

src/web/components/BusinessContacts/BusinessContactsTable.scss
```diff
@@ -24,9 +24,12 @@
       border: none;
     }
 
-    td.action {
+    th.action {
+      text-align: center;
```
lionell-pack-ttd, line 28: Making the th center and the td (below) flex-end means it gets out of alignment as the browser gets to full-width.
image

src/web/styles/themes.scss
```diff
@@ -35,6 +35,7 @@
   --theme-disabled-button-text: #878787;
   --theme-disabled-button: #d9d9d9;
   --theme-label: rgba(120, 204, 0, 0.4);
+  --theme-error-label: rgba(253, 4, 4, 0.25);
```
lionell-pack-ttd, line 38: This should probably be theme-error-label-background to be consistent with the other background colours.

src/web/components/Core/InlineError.scss
```diff
@@ -0,0 +1,11 @@
+.inline-error {
+  margin: 0 8px;
+}
+
+.inline-error-content {
```
lionell-pack-ttd, line 5: Looking at Figma, the inline error example has a font color (#030A40;) that I don't think is being applied here.

src/web/components/TeamMember/TeamMembersTable.scss
```diff
@@ -21,21 +21,28 @@
     }
 
     .email {
-      width: 60%;
+      width: 40%;
+    }
+
+    th.action {
+      text-align: center;
```
lionell-pack-ttd, line 28: This creates the same alignment problem I mentioned on the email contacts table.

src/api/usersRouter.ts
```diff
@@ -62,15 +62,33 @@ usersRouter.post('/:userId/resendInvitation', async (req: UserRequest, res) => {
...
+    return res.sendStatus(200);
+  });
+
+  const UpdateUserParser = z.object({
+    firstName: z.string(),
+    lastName: z.string(),
+    role: z.nativeEnum(UserRole),
+  });
+
+  usersRouter.put('/:userId', async (req: UserRequest, res) => {
```
lionell-pack-ttd, line 86: This should probably be usersRouter.patch - it's partly updating an entity, but leaving some fields unchanged.

src/web/services/userAccount.ts
```diff
@@ -80,6 +87,22 @@ export async function GetAllUsersOfParticipant(participantId?: number) {
...
+  try {
+    await axios.delete(`/users/${id}`);
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not delete user');
+  }
+}
+
+export async function UpdateUser(id: number, formData: InviteTeamMemberForm) {
+  try {
+    await axios.patch(`/users/${id}`, formData);
```
lionell-pack-ttd, line 100: As discussed - this is sending too much data, limit what's being sent to just the fields that can be updated.

src/api/usersRouter.ts
```diff
@@ -62,15 +62,36 @@ usersRouter.post('/:userId/resendInvitation', async (req: UserRequest, res) => {
...
     return res.sendStatus(200);
-});
+  });
+
+  usersRouter.delete('/:userId', async (req: UserRequest, res) => {
+    const { user } = req;
+    if (req.auth?.payload?.email === user?.email) {
+      return res.status(403).send([{ message: 'You do not have permission to delete yourself.' }]);
+    }
+    await user!.$query().delete();
```
lionell-pack-ttd, line 79: I just realised this will still leave the user with a keycloak account - so they could potentially still get a valid JWT and possibly even make API calls. This should delete the user in keycloak as well.

src/web/components/TeamMember/TeamMember.tsx
lionell-pack-ttd, line 1: I think this makes it too easy to accidentally delete a user - this is a significant action and needs an extra confirmation step.

src/api/services/kcUsersService.ts
```diff
@@ -40,3 +43,35 @@ export const sendInviteEmail = async (
...
+
+export const deleteUser = async (kcAdminClient: KeycloakAdminClient, user: UserRepresentation) => {
+  await kcAdminClient.users.del({
+    id: user.id!,
+  });
+};
+
+export interface KcUserRequest extends UserRequest {
+  kcUser?: UserRepresentation;
+  kcAdminClient?: KeycloakAdminClient;
```
lionell-pack-ttd, line 55: I don't think providing something like a keycloak client is appropriate using an enricher like this. It's nothing to do with the request itself.

lionell-pack-ttd: I see the reasoning - you don't want to auth twice. But putting a rich object like a client on the request seems like a bad idea - the request object is for data related to the request.

jingyi-gao-ttd: Yep, I agree, it doesn't make sense to include that object in the request. Removed

src/web/components/TeamMember/TeamMember.tsx
```diff
@@ -3,18 +3,77 @@ import clsx from 'clsx';
...
+  person: UserResponse;
+  onRemoveTeamMember: () => Promise<void>;
+};
+
+function DeleteConfirmationDialog({ person, onRemoveTeamMember }: DeleteConfirmationDialogProps) {
+  const [openConfirmation, setOpenConfirmation] = useState(false);
+
+  const handleRemove = () => {
+    setOpenConfirmation(false);
+    onRemoveTeamMember();
```
lionell-pack-ttd, line 21: I think you need to await this - otherwise the state refresh at the top level could possibly happen before the API call to delete finishes.

lionell-pack-ttd (approved):

alex-yau-ttd:
I was expecting to see my name change in the top right corner after updating my name

image

Is that worth addressing in this PR?

jingyi-gao-ttd:
> I was expecting to see my name change in the top right corner after updating my name
> 
> image
> 
> Is that worth addressing in this PR? Looks like they might be based off different sources of truth?

Yes, they are based on different sources of truth, but the right top should be updated as well, that's a good catch! I will add that to this PR, thanks!

src/api/services/kcUsersService.ts
```diff
@@ -41,6 +41,27 @@ export const sendInviteEmail = async (
...
+  lastName: string;
+};
+
+export const updateUserProfile = async (
+  kcAdminClient: KeycloakAdminClient,
+  userEmail: string,
+  updateUserPayload: UpdateUserPayload
+) => {
+  const users = await queryUsersByEmail(kcAdminClient, userEmail);
+  if (users.length !== 1) throw Error(`Unable to update  entry for ${userEmail}`);
```
alex-yau-ttd, line 55: nit: there's an extra space in the string

src/api/usersRouter.ts
```diff
@@ -62,15 +67,43 @@ usersRouter.post('/:userId/resendInvitation', async (req: UserRequest, res) => {
...
+  const UpdateUserParser = z.object({
+    firstName: z.string(),
+    lastName: z.string(),
+    role: z.nativeEnum(UserRole),
+  });
+
+  usersRouter.patch('/:userId', async (req: UserRequest, res) => {
+    const { user } = req;
+    const data = UpdateUserParser.parse(req.body);
+    const kcAdminClient = await getKcAdminClient();
```
lionell-pack-ttd, line 99: Minor (doesn't need to be fixed unless you want to!) - the database update and the keycloak update are independent, so you don't need to wait for one to finish before starting the other - you could start them both and then await them both.

The same applies to the delete.

lionell-pack-ttd (approved):

PR #93, opened 2023-07-20
ajy-UID2-1028-List-outstanding-approvals

alex-yau-ttd:
Add the Manage Participants admin page, which lists approvals that are awaiting approvals.

**User is an approver-> show table with relevant participants** 

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/68a9963b-905e-431e-ba3a-c641aeb968d4

**User is an approver but there are no participant requests -> show No Approvals**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/df87862b-db21-4a25-a6f0-789c6acda6d8

**User is not an approver -> don't show the new page**

image


src/api/participantsRouter.ts
```diff
@@ -31,7 +32,10 @@ import {
   getAllUserFromParticipant,
 } from './services/usersService';
 
-export type AvailableParticipantDTO = Pick<ParticipantDTO, 'id' | 'name' | 'siteId' | 'types'>;
```
alex-yau-ttd, line 34: Wonder if this should be a separate DTO rather than just adding to this one?

src/api/participantsRouter.ts
```diff
@@ -49,6 +53,23 @@ export function createParticipantsRouter() {
     return res.status(200).json(participants.map(mapParticipantToAvailableParticipant));
   });
 
+  participantsRouter.get('/awaitingApproval', async (req: ParticipantRequest, res) => {
```
alex-yau-ttd, line 56: This logic is a little messy, so thinking of either trying to simplify it or at least moving to the service? Let me know what you think

src/api/usersRouter.ts
```diff
@@ -35,23 +42,23 @@ export function createUsersRouter() {
     res.status(201).json(user);
   });
 
-usersRouter.put('/current/acceptTerms', enrichCurrentUser, async (req: UserRequest, res) => {
-  await req.user!.$query().patch({ acceptedTerms: true });
-  return res.sendStatus(200);
+  usersRouter.put('/current/acceptTerms', enrichCurrentUser, async (req: UserRequest, res) => {
```
alex-yau-ttd, line 45: A few formatting changes in this file

src/database/seeds/Participants.ts
```diff
@@ -15,27 +15,31 @@ const sampleData: Optional<
     status: ParticipantStatus.AwaitingSigning,
     type: 'Publisher',
     location: 'Sydney',
+    types: [],
```
alex-yau-ttd, line 18: If we do add this, maybe I should populate these?

src/web/components/Approvals/ApprovalItem.scss
```diff
@@ -0,0 +1,31 @@
+.approval-item {
+  line-height: 2rem;
```
alex-yau-ttd, line 2: Probably not an ideal way to get the rows the same height (at at least look the same height) as the other tables. A lot of the scss has been copied from elsewhere then just changed to look how I think it should.

src/web/components/Approvals/ApprovalItem.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  .approval-logo {
+    width: 50px;
+    margin-right: 16px;
+  }
+
+  .approval-types {
+    display: flex;
+    align-items: center;
+    font-size: 0.75rem;
```
alex-yau-ttd, line 18: This was `0.625rem` in ParticipantItem.scss which I thought looked too small here.

src/web/components/Approvals/ApprovalItem.tsx
```diff
@@ -0,0 +1,40 @@
+import { z } from 'zod';
+
+import { ParticipantTypeSchema } from '../../../api/entities/ParticipantType';
+import { AvailableParticipantDTO } from '../../../api/participantsRouter';
+
+import './ApprovalItem.scss';
+
+type ParticipantItemProps = {
+  participant: AvailableParticipantDTO;
```
alex-yau-ttd, line 9: As per previous comment, not sure if I should be using this DTO here or using a new one

src/web/components/Approvals/ApprovalItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export function ApprovalItem({ participant }: ParticipantItemProps) {
+  function getParticipantTypes(participantTypes?: z.infer<typeof ParticipantTypeSchema>[]) {
```
alex-yau-ttd, line 13: Copied from ParticipantItem - perhaps could be refactored?

src/web/components/Approvals/ApprovalsTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+import './ApprovalsTable.scss';
+
+type ApprovalsTableProps = {
+  participants: Participant[];
+};
+
+function NoApprovals() {
+  return (
+    <div className='no-approvals-container'>
+      <img src='/group-icon.svg' alt='group-icon' />
```
alex-yau-ttd, line 13: I guess we want a different icon here, but just putting it here as a placeholder

src/web/screens/approvals.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { Participant } from '../../api/entities/Participant';
+import { ApprovalsTable } from '../components/Approvals/ApprovalsTable';
+import { Loading } from '../components/Core/Loading';
+import { GetParticipantsAwaitingApproval } from '../services/participant';
+import { PortalRoute } from './routeUtils';
+
+function Approvals() {
+  const { participantsAwaitingApproval } = useLoaderData() as {
+    participantsAwaitingApproval: Participant[];
```
alex-yau-ttd, line 12: might rename this to participantRequests or something

src/web/screens/dashboard.tsx
```diff
@@ -34,8 +36,11 @@ function Dashboard() {
   const { LoggedInUser, loadUser } = useContext(CurrentUserContext);
   const [showMustAccept, setShowMustAccept] = useState(false);
   const navigate = useNavigate();
+  const filteredMenu = LoggedInUser?.user?.isApprover
```
alex-yau-ttd, line 39: Not sure if this is the best way to do this

src/api/participantsRouter.ts
lionell-pack-ttd: I like the idea of moving it into a service.

src/api/services/usersService.ts
```diff
@@ -4,9 +4,17 @@ import { z } from 'zod';
 import { Participant } from '../entities/Participant';
 import { User, UserDTO } from '../entities/User';
 import { getLoggers } from '../helpers/loggingHelpers';
+import { isUserAnApprover } from './approversService';
+
+export type UserWithIsApprover = User & { isApprover: false };
 
 export const findUserByEmail = async (email: string) => {
-  return User.query().findOne('email', email);
+  const user = await User.query().findOne('email', email);
```
lionell-pack-ttd, line 12: This is causing a bunch of typescript errors for me - before, the return type was `User | undefined`, but because this is using the spread operator (below) to combine user with isApprover, it changes it to always returning something, but all of the properties are now `string | undefined`, 'number | undefined`, etc.

src/database/seeds/Users.ts
```diff
@@ -13,6 +13,7 @@ const sampleParticipant = {
   status: ParticipantStatus.AwaitingSigning,
   allowSharing: true,
   location: 'Sydney',
+  types: [],
```
lionell-pack-ttd, line 16: Why this? It's optional and the thing it's used for uses a separate parameter for type.

src/web/screens/dashboard.tsx
lionell-pack-ttd: I think this is OK for just one thing like this - probably the next time we add a conditional display menu item we should come up with something better.

src/api/participantsRouter.ts
alex-yau-ttd: yep, have added `ParticipantRequestDTO`

src/api/participantsRouter.ts
alex-yau-ttd: done!

src/web/components/Approvals/ApprovalItem.scss
alex-yau-ttd: Changed to just be 2, instead of 2rem.

src/web/components/Approvals/ApprovalItem.tsx
alex-yau-ttd: Did not end up extracting, but changed the signature since we don't need infer

src/web/components/Approvals/ApprovalItem.tsx
alex-yau-ttd: ParticipantRequestItem is now using ParticipantRequestDTO

src/web/screens/approvals.tsx
alex-yau-ttd: done

src/database/seeds/Users.ts
alex-yau-ttd: removed

src/api/services/usersService.ts
alex-yau-ttd: have tried to address - let me know if that works

src/web/screens/manageParticipants.tsx
```diff
@@ -0,0 +1,41 @@
...
+    </div>
+  );
+}
+
+export const ManageParticipantsRoute: PortalRoute = {
+  path: '/dashboard/manageParticipants',
+  description: 'Manage Participants',
+  element: <ManageParticipants />,
+  loader: async () => {
+    const participantsAwaitingApproval = await GetParticipantsAwaitingApproval();
```
jingyi-gao-ttd, line 36: could you remove the await from here, otherwise this will wait for the request to resolve and then render the page

jingyi-gao-ttd: here is the documentation: https://reactrouter.com/en/main/route/loader

src/api/services/participantsService.ts
```diff
@@ -53,6 +58,21 @@ export const getSharingParticipants = async (participantSiteId: number): Promise
   return fetchSharingParticipants(sharingListResponse);
 };
 
+export const getParticipantsAwaitingApproval = async (email: string): Promise<Participant[]> => {
+  const approvers = await Approver.query().distinct('participantTypeId').where('email', email);
+  const approvableParticipantTypeIds = approvers.map((approver) => approver.participantTypeId);
```
jingyi-gao-ttd, line 63: Could we move the logic of getting participantTypeIds by email to the approver service?

src/web/screens/manageParticipants.tsx
alex-yau-ttd: done! so I guess it only makes sense to await subsequent calls (rather than the first). e.g. sharingPermissions has:
```
  loader: async () => {
    const participants = GetAllAvailableParticipants();
    const participantTypes = await GetAllParticipantTypes();
    return defer({ participants, participantTypes });
  },
```
so the page loads after participants comes back, then we async load participantTypes.
Is that right?

src/api/services/participantsService.ts

jingyi-gao-ttd (approved):

PR #98, opened 2023-07-24
UID2 1473 Update keycloak

lionell-pack-ttd:
Tweak some CSS and always show "Sign out from other devices" when changing password.


lionell-pack-ttd:
image

jingyi-gao-ttd (approved):

PR #100, opened 2023-07-25
UID2-1516 setup role check on endpoint

jingyi-gao-ttd:
- Add claimCheck to check if user has api-participant-user role
- Update keycloak to assign api-user by default
- Update bypassHandlerForPaths to be able to check method
- Update getParticipantByUserId to getCurrentUsersParticipant, which use `users/current/participant instead`

Required steps
- [ ] Import realm settings to keycloak
- [ ] Manually assign api-participant-user to existing users with approved participant
- [ ] Before we have approval in portal, we need to manually assign roles every time we approve a participant

### DIFF
--- a/src/api/configureApi.ts
+++ b/src/api/configureApi.ts
@@ -107,6 +107,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
         const roles = claim.resource_access?.self_serve_portal_apis?.roles || [];
         return roles.includes('api-participant-member');
       }),
+      { url: `${BASE_REQUEST_PATH}/participantTypes`, method: 'GET' },
       { url: `${BASE_REQUEST_PATH}/participants`, method: 'POST' },
       { url: `${BASE_REQUEST_PATH}/users/current`, method: 'GET' },
       { url: `${BASE_REQUEST_PATH}/users/current/participant`, method: 'GET' },
--- a/src/api/configureApi.ts
+++ b/src/api/configureApi.ts
@@ -86,7 +86,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
         audience: SSP_KK_AUDIENCE,
         issuerBaseURL: SSP_KK_ISSUER_BASE_URL,
       }),
-      BYPASS_PATHS
+      ...BYPASS_PATHS
     )
   );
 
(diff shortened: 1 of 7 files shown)


src/api/usersRouter.ts
```diff
@@ -19,27 +19,28 @@ import {
...
     return res.json(userWithIsApprover);
   });
 
-  usersRouter.put('/current/acceptTerms', enrichCurrentUser, async (req: UserRequest, res) => {
+  usersRouter.put('/current/acceptTerms', async (req: UserRequest, res) => {
     await req.user!.$query().patch({ acceptedTerms: true });
     return res.sendStatus(200);
   });
 
+  usersRouter.get('/current/participant', async (req: UserRequest, res) => {
```
jingyi-gao-ttd, line 33: It needs exact work for the middleware to be able to match the path with params. For now, this endpoint is only used by get current user's participant. I think it is fine to use `/current` pattern.

src/api/configureApi.ts
```diff
@@ -26,11 +26,23 @@ import { createParticipantsRouter } from './participantsRouter';
...
-function bypassHandlerForPaths(middleware: express.Handler, ...paths: string[]) {
+const BYPASS_PATHS = [
+  `/favicon.ico`,
+  `${BASE_REQUEST_PATH}/`,
+  `${BASE_REQUEST_PATH}/metrics`,
+  `${BASE_REQUEST_PATH}/health`,
+  `${BASE_REQUEST_PATH}/keycloak-config`,
+];
+
+function bypassHandlerForPaths(middleware: express.Handler, ...paths: (string | string[])[]) {
```
lionell-pack-ttd, line 37: Instead of using a string array for [path, method], it would be better to use an object. `path.url` and `path.method` looks better than `path[0]` and `path[1]`.

src/api/configureApi.ts
```diff
@@ -26,11 +26,23 @@ import { createParticipantsRouter } from './participantsRouter';
...
   return function (req, res, next) {
-    const pathCheck = paths.some((path) => path === req.path);
-    if (pathCheck) {
-      next();
+    const bypassPath = paths.find((path) =>
+      Array.isArray(path) ? path[0] === req.path : path === req.path
```
lionell-pack-ttd, line 40: If you map all the paths which are just strings to objects and give them a default `.method='GET'` then this logic will become simpler and you won't have to check the type multiple times.

lionell-pack-ttd (approved):

PR #102, opened 2023-07-26
navigate user to the next screen

jingyi-gao-ttd:
- Updated email sender in keycloak
- Navigate user to action page and redirect user to portal after they update their account


keycloak/themes/uid2-theme/login/info.ftl
```diff
@@ -0,0 +1,25 @@
...
+    <#elseif section = "form">
+    <div id="kc-info-message">
+        <p class="instruction">${message.summary}<#if requiredActions??><#list requiredActions>: <b><#items as reqActionItem>${kcSanitize(msg("requiredAction.${reqActionItem}"))?no_esc}<#sep>, </#items></b></#list><#else></#if></p>
+        <#if skipLink??>
+        <#else>
+            <#if pageRedirectUri?has_content>
+              <script>location.href = '${pageRedirectUri}';</script>
+            <#elseif actionUri?has_content>
+              <p><a id="proceedWithAction" href="${actionUri}">${kcSanitize(msg("proceedWithAction"))?no_esc}</a></p>
+              <script>document.getElementById('proceedWithAction').click()</script>
```
jingyi-gao-ttd, line 18: keycloak is setting cookie and proceeding action token at this page, using location.href will cause an infinity loop. So we have to use click instead.

keycloak/realm/realm-export.json
```diff
@@ -1465,8 +1465,8 @@
     "port": "1025",
     "host": "mailhog",
     "replyTo": "",
-    "from": "test@self-serve-portal.com",
-    "fromDisplayName": "",
+    "from": "noreply@unifiedid.com",
+    "fromDisplayName": "UID2 Service (do not reply)",
```
alex-yau-ttd, line 1469: is it standard practice to have the "do not reply" in the display name? I feel like the usual pattern would be the display name is just `UID Service` and the "no reply" is in the email address like you already have.

alex-yau-ttd:
is it possible to include a video that showcases this new functionality? I tried to manually test but not sure if I got it working right

keycloak/realm/realm-export.json
jingyi-gao-ttd: I think we are asked to use this display name, can't find the slack message for that anymore, but this is consistent with the email we currently send out from portal (the email send to approver)

jingyi-gao-ttd:
Sure, @alex-yau-ttd  here is the screen recording

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/6ab24b89-77a4-4352-be50-3be3bfe4e013

alex-yau-ttd (approved):

PR #103, opened 2023-07-26
ajy-UID2-1443-Bulk-add-permissions-component

alex-yau-ttd:
**Build bulk add permissions component.**

This is not fully functional yet (doesn't call any APIs). 
However, it should populate the recommendation message and select default checkboxes based on the recommended types listed in the ticket. Clicking the "View" button will just change the text to "Hide", and vice versa.

Here is a screen recording from storybook (minor css adjustments have been made since):

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/a0b4037d-bb6a-4b1c-9207-d1f57e488f3f

Screenshot with updated padding:

image

### DIFF
--- a/src/web/components/SharingPermission/BulkAddPermissions.scss
+++ b/src/web/components/SharingPermission/BulkAddPermissions.scss
@@ -1,0 +1,49 @@
+.bulk-add-permissions {
+  display: flex;
+  flex-direction: column;
+  border: 2px solid var(--theme-action);
+  border-radius: 5px;
+  line-height: 1;
+  box-shadow: 0px 4px 14px var(--theme-box-shadow);
+
+  .bulk-add-permissions-header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 0 20px;
+    flex-direction: row;
+
+    .recommendation-label {
+      background: var(--theme-label-background);
+      border-radius: 3px;
+      padding: 4px 8px;
+      margin-right: 8px;
+      display: flex;
+      align-items: center;
+      font-size: 0.75rem;
+      width: 115px;
+      height: 18px;
+    }
+  }
+
+  .bulk-add-permissions-body {
+    padding: 0 40px;
+
+    .participant-type-checkbox-section {
+      display: flex;
+      align-items: center;
+
+      .checkbox-label {
+        padding-right: 25px;
+      }
+    }
+
+    .bulk-add-permissions-view-recommendations-button {
+      padding-top: 20px;
+    }
+  }
+
+  .bulk-add-permissions-footer {
+    padding: 20px 20px;
+  }
+}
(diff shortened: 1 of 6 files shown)


src/web/components/SharingPermission/BulkAddPermissions.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  .bulk-add-permissions-header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 0 20px;
+    flex-direction: row;
+
+    h2 {
+      margin-top: 20px;
```
alex-yau-ttd, line 17: thought about adding this to the generic styling in `text.scss` but figured this was saver for now. Happy to change the generic one from 30 to 20 though if other people are happy

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
```diff
@@ -0,0 +1,57 @@
+import { formatStringsWithSeparator, getArticle } from '../../utils/textHelpers';
```
alex-yau-ttd, line 1: Not sure how we feel about helper files, but this seemed like stuff that could be moved out of the component. Let me know what you think

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const getDefaultPublisherCheckboxState = (participantTypeNames: string[]) => {
+  if (participantTypeNames?.includes('Data Provider')) return true;
```
jingyi-gao-ttd, line 4: I found types are used a lot in this file, could we change them to constant or enum? Just in case we rename or add a new participant type.

NIT:not sure if something like 
```
const types = ['Data Prodiver', 'DSP] 
participantTypesNames.some(type =>types.includes(type))
```
make sense to you, I think then this could be reused rather than multiple 
`||participantTypeNames?.includes()`

alex-yau-ttd: ✔ good suggestion! I was struggling to think of a way to make this code better, so that definitely helps.

src/web/components/SharingPermission/BulkAddPermissions.scss
```diff
@@ -0,0 +1,53 @@
...
+
+    .recommendation-label {
+      background: var(--theme-label-background);
+      border-radius: 3px;
+      padding: 4px 8px;
+      margin-right: 8px;
+      display: flex;
+      align-items: center;
+      font-size: 0.75rem;
+      width: 115px;
```
lionell-pack-ttd, line 28: Setting an explicit width is probably going to be a bit flakey - it looks like it sizes basically the same without this.

jingyi-gao-ttd (approved):

src/web/components/SharingPermission/BulkAddPermissions.scss
alex-yau-ttd: ✔ removed

PR #105, opened 2023-07-26
Fix possible issue with Keycloak sometimes using profile username ins…

lionell-pack-ttd:
…tead of email.


jingyi-gao-ttd:
Tested locally with upgraded keycloak version, cannot reproduce this issue, will test this in integ first

lionell-pack-ttd:
Converting to a draft - maybe this problem doesn't exist? I'll close it in a few days without merging if we don't see it happen.

PR #106, opened 2023-07-26
ajy-UID2-1443-Accordion-component

alex-yau-ttd:
Add accordion component from radix.

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/c50bd65f-8ae7-4e9a-8929-cd0f56c525bd

### DIFF
--- a/src/web/components/Core/Accordion.tsx
+++ b/src/web/components/Core/Accordion.tsx
@@ -1,0 +1,50 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import * as RadixAccordion from '@radix-ui/react-accordion';
+import { ReactNode } from 'react';
+
+import './Accordion.scss';
+
+export type AccordionProps = {
+  title: string;
+  content: ReactNode;
+  defaultExpanded: boolean;
+  isCollapsible: boolean;
+  label: ReactNode;
+};
+
+export function Accordion({
+  title,
+  content,
+  defaultExpanded,
+  isCollapsible,
+  label,
+}: AccordionProps) {
+  const itemName = 'item-1';
+  const defaultValue = defaultExpanded ? itemName : '';
+
+  return (
+    <RadixAccordion.Root
+      className='accordion-root'
+      type='single'
+      defaultValue={defaultValue}
+      collapsible
+    >
+      <RadixAccordion.Item className='accordion-item' value={itemName}>
+        <div className='accordion-header'>
+          <h2>{title}</h2>
+          {isCollapsible && (
+            <RadixAccordion.AccordionTrigger className='accordion-trigger'>
+              <FontAwesomeIcon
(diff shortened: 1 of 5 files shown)


src/web/components/Core/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export function Accordion({
+  title,
+  content,
+  defaultExpanded,
+  isCollapsible,
+  label,
+}: AccordionProps) {
+  const itemName = 'item-1';
```
alex-yau-ttd, line 22: the defaultValue on the Root and the value of the item need to be the same for it to show by default. let me know if you can think of a better way.

src/web/components/Core/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+import * as RadixAccordion from '@radix-ui/react-accordion';
+import { ReactNode } from 'react';
+
+import './Accordion.scss';
+
+export type AccordionProps = {
+  title: string;
+  content: ReactNode;
+  defaultExpanded: boolean;
+  isCollapsible: boolean;
```
alex-yau-ttd, line 11: I thought about maybe creating a separate component to support the case where isCollapsible is false, or maybe making these props optional since some combinations don't make sense. e.g. label should only be defined when isCollapsible is false. let me know what you think

src/web/components/Core/Accordion.tsx
```diff
@@ -0,0 +1,50 @@
...
+  content,
+  defaultExpanded,
+  isCollapsible,
+  label,
+}: AccordionProps) {
+  const itemName = 'item-1';
+  const defaultValue = defaultExpanded ? itemName : '';
+
+  return (
+    <RadixAccordion.Root
```
alex-yau-ttd, line 26: I also thought about allowing multiple items, but in my testing it seems to be fine to have multiple roots

src/web/components/Core/Accordion.tsx
alex-yau-ttd: for instance, wondering if these two should be different components or the same component with different props:

image

image

alex-yau-ttd: differences of note:
- border styling
- label
- collapsibility
- Behaviour of deselecting the recommendations

jingyi-gao-ttd: For me it makes sense to be two components, there is no reason to use Accordion if the component is not Collapsible

alex-yau-ttd:
After talking with Jingyi, it probably makes more sense to use the Collapsible component rather than the accordion.

src/web/components/Core/Accordion.tsx
alex-yau-ttd: agreed, will do

PR #107, opened 2023-07-27
ajy-UID2-1443-Collapsible-component

alex-yau-ttd:
Add collapsible component from Radix

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/09eb0f20-1660-4aee-a2f7-64d0a4cf1c74

### DIFF
--- a/src/web/components/Core/Collapsible.tsx
+++ b/src/web/components/Core/Collapsible.tsx
@@ -1,0 +1,31 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import * as RadixCollapsible from '@radix-ui/react-collapsible';
+import { ReactNode } from 'react';
+
+import './Collapsible.scss';
+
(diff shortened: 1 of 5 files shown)


src/web/components/Core/Collapsible.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import './Collapsible.scss';
+
+export type CollapsibleProps = {
+  title: string;
+  content: ReactNode;
+  defaultOpen: boolean;
+};
+
+export function Collapsible({ title, content, defaultOpen }: CollapsibleProps) {
```
alex-yau-ttd, line 13: I've left this as uncontrolled for now, but we can easily make it controlled later per the example here: https://www.radix-ui.com/docs/primitives/components/collapsible

jingyi-gao-ttd (approved):

PR #108, opened 2023-07-27
Jyg UI d2 1474 change search sharing permission workflow

jingyi-gao-ttd:
- image

- on sharing permission page

image

image


src/web/components/SharingPermission/ParticipantSearchBar.scss
```diff
@@ -35,8 +34,8 @@
   outline: none;
   color: var(--theme-button-text);
   font-weight: 400;
-  font-size: 1.875rem;
-  line-height: 2.25rem;
+  font-size: 1.1875rem;
+  line-height: 1.4375rem;
```
alex-yau-ttd, line 38: Lionell recently suggested I not use rem for `line-height`, since it can start acting weirdly when the font size changes. e.g. with a really big font and a rem `line-height`, the text cuts off. The solution is to just use it without units, where it acts as a scalar, though you might need to play with the values to make it look the same. Is that okay?

jingyi-gao-ttd: Good catch! I ran into the same problem before, will remove all units

src/web/screens/sharingPermissions.tsx
```diff
@@ -27,6 +30,10 @@ function SharingPermissions() {
   const { participant } = useContext(ParticipantContext);
   const [sharingParticipants, setSharingParticipants] = useState<AvailableParticipantDTO[]>([]);
   const [statusPopup, setStatusPopup] = useState<StatusPopupType>();
+  const { participants, participantTypes } = useLoaderData() as {
+    participants: AvailableParticipantDTO[];
+    participantTypes: ParticipantType[];
```
alex-yau-ttd, line 35: maybe we can use ParticipantTypeDTO instead of ParticipantType here?

src/web/components/SharingPermission/searchAndAddParticipants.tsx
```diff
@@ -1,35 +1,31 @@
-import { Suspense, useContext, useMemo, useState } from 'react';
```
alex-yau-ttd, line 1: Could we possibly rename these three files to be PascalCase, for consistency?

src/web/screens/sharingPermissions.tsx
jingyi-gao-ttd: done

src/web/components/SharingPermission/searchAndAddParticipants.tsx
jingyi-gao-ttd: done

src/web/components/SharingPermission/ParticipantSearchBar.tsx
```diff
@@ -43,19 +48,19 @@ export function ParticipantSearchBar({
...
+    <div className={clsx('search-bar', { clicked: open })}>
       <div className='search-bar-input-container'>
         <input
           type='text'
           className='search-bar-input'
-          onClick={() => setDropdownOpen(true)}
+          onClick={() => onToggle(true)}
           onChange={handleFilterTextChange}
-          placeholder='Search and Add Participants'
+          placeholder='Search Participants'
```
alex-yau-ttd, line 58: Is it possible to also display this placeholder when the the search is `open`? That's what the design shows: 
image

jingyi-gao-ttd: Yes, we used to get feedback about should not show it, so I set it to transparent. I removed css for that now

alex-yau-ttd:
If I were to nit the styling, I think the padding on left of the participants is a bit bigger in the design.
Current: 

image


Design:

image

jingyi-gao-ttd:
> If I were to nit the styling, I think the padding on left of the participants is a bit bigger in the design. Current:
> 
> image
> 
> Design:
> 
> image

done

src/web/components/SharingPermission/ParticipantSearchBar.tsx
```diff
@@ -2,28 +2,31 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
 import './ParticipantSearchBar.scss';
 
 type ParticipantSearchBarProps = {
   participants: AvailableParticipantDTO[];
-  participantTypes: ParticipantTypeResponse[];
+  participantTypes: ParticipantTypeDTO[];
   selectedParticipantIds?: Set<number>;
   onSelectedChange: (selectedItems: Set<number>) => void;
+  open: boolean;
+  onToggle: (open: boolean) => void;
```
alex-yau-ttd, line 18: I wonder if we can rename this prop to be more clear - what is is toggling? maybe `onToggleOpen`?

jingyi-gao-ttd: agree, changed

alex-yau-ttd:
> > If I were to nit the styling, I think the padding on left of the participants is a bit bigger in the design. Current:
> > image
> > Design:
> > image
> 
> done

hmm I think it looks worse now? unless something weird with the scss files going on (though I think I fixed them locally)
image

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "strict": true,
-    "forceConsistentCasingInFileNames": true,
+    "forceConsistentCasingInFileNames": false,
```
alex-yau-ttd, line 10: Can we turn this back to true now that it seems fixed?

alex-yau-ttd (approved):

PR #109, opened 2023-07-28
ajy-UID2-1537-Implement-styling-changes

alex-yau-ttd:
Affected areas include:
- Participant tables
- ParticipantRequestsTable (in Managing Participants)
- h2 (trimmed 10px)

**Participant table**

Before

image

After

image

**Participant table empty**

Before

image

After

image

**Participant request table**

Before

image

After

image

**Other changes**
- Added storybook for participants request table.

### DIFF
--- a/src/web/components/Core/Collapsible.scss
+++ b/src/web/components/Core/Collapsible.scss
@@ -10,10 +10,6 @@
     justify-content: space-between;
     padding-left: 20px;
 
-    h2 {
-      margin: 20px 0;
-    }
-
     .collapsible-trigger {
       border: none;
       border-radius: 5px;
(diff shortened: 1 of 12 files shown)


jingyi-gao-ttd:
I found your sharing permission table has a box-shadow on it, could you please add that?
nit: I guess we could remove the No Participant now, coz we will hide this table in that case

alex-yau-ttd:
> I found your sharing permission table has a box-shadow on it, could you please add that? 
done!

> nit: I guess we could remove the No Participant now, coz we will hide this table in that case
I'd like to do this in a separate PR if that is okay with you. Reason is that I'd like to add some logic to conditionally show/hide in the same PR, so that we don't end up in a weird in-between state.

jingyi-gao-ttd (approved):

PR #111, opened 2023-07-31
ajy-UID2-1443-Warning-and-error-banners

alex-yau-ttd:
Note that info and error banners are not currently used (even in designs), hence they don't exactly match the style of the Warning banner which uses an svg icon rather than font awesome.

Info

image

Warning

image

Error

image


src/web/components/Core/Banner.tsx
```diff
@@ -1,17 +1,28 @@
...
 import './Banner.scss';
 
 type BannerProps = {
   message: string;
-  className?: string;
+  type: 'info' | 'warning' | 'error';
 };
 
-export function Banner({ message, className }: BannerProps) {
+export function Banner({ message, type }: BannerProps) {
```
alex-yau-ttd, line 12: warning uses svg while the other two use font awesome. Not ideal but only the warning banner actually has a design for it, so we can always change this in future if we have a need for the other two.

jingyi-gao-ttd (approved):

PR #112, opened 2023-07-31
ajy-UID2-1443-Collapsible-component-improvements

alex-yau-ttd:
Now the whole header is clickable to collapse/expand the Collapsible component. Before, the trigger was just a small box around the chevron icon.

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/aca24764-3842-4558-b6ae-d11b8b27376e

Also, a label string can be passed to the collapsible component:

image

### DIFF
--- a/src/web/components/Core/Collapsible.tsx
+++ b/src/web/components/Core/Collapsible.tsx
@@ -13,16 +13,16 @@ export type CollapsibleProps = {
 export function Collapsible({ title, content, defaultOpen }: CollapsibleProps) {
   return (
     <RadixCollapsible.Root className='collapsible-root' defaultOpen={defaultOpen}>
-      <div className='collapsible-header'>
-        <h2>{title}</h2>
-        <RadixCollapsible.Trigger className='collapsible-trigger'>
+      <RadixCollapsible.Trigger className='collapsible-trigger'>
+        <div className='collapsible-header'>
+          <h2>{title}</h2>
           <FontAwesomeIcon
             icon='chevron-down'
             data-testid='chevron-icon'
             className='chevron-icon'
           />
-        </RadixCollapsible.Trigger>
-      </div>
+        </div>
+      </RadixCollapsible.Trigger>
       <RadixCollapsible.Content>
         <div className='collapsible-content'>{content}</div>
       </RadixCollapsible.Content>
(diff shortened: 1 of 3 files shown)


jingyi-gao-ttd (approved):

src/web/components/Core/Collapsible.tsx
```diff
@@ -8,21 +8,25 @@ export type CollapsibleProps = {
...
-          <FontAwesomeIcon
-            icon='chevron-down'
-            data-testid='chevron-icon'
-            className='chevron-icon'
-          />
-        </RadixCollapsible.Trigger>
-      </div>
+      <RadixCollapsible.Trigger className='collapsible-trigger'>
+        <div className='collapsible-header'>
+          <h2>{title}</h2>
```
jingyi-gao-ttd, line 19: How about taking a header prop which is a ReactNode? So that if we need to make the component looks different when collapsed we could just pass into this prop?

alex-yau-ttd: ah good idea. to clarify, do you propose that the title and the icon would be part of this header too?

jingyi-gao-ttd: Yea, I am thinking something like
```
<Collapsible header={
 <div className='collapsible-header'>
      <h2>{title}</h2>
      <div className='collapsible-header-label-and-icon'>
        {label && <span className='label'>{label}</span>}
        <FontAwesomeIcon
          icon='chevron-down'
          data-testid='chevron-icon'
          className='chevron-icon'
        />
      </div>
    </div>
}/>
```

Just for discussion if you think this makes sense?

alex-yau-ttd: definitely a discussion worth having. I would have thought that all the uses of the collapsible would use the same h2 for the title, and the same chevron for the icon. So the text for the title as well as the label are the only things that would differ, which would lead more to just having title and label props, so that the consumer doesn't have to repeat that pattern on every usage.

One alternative could be for the label prop to be a ReactNode? Though I would've thought only the styling (e.g. background colour) might change, so perhaps not the best idea.

Does that sound reasonable?

jingyi-gao-ttd: I see, yes that's reasonable to me. Let's just leave label in this component

alex-yau-ttd: great, thanks for advising!

PR #113, opened 2023-07-31
ajy-UID2-1443-Add-logic-to-bulk-add-permissions-component

alex-yau-ttd:
- If there are no sharing participants, show the expanded recommendation collapsible that features different styling, as well as the view/hide participants button
- If there are sharing participants, show the collapsed collapsible that lacks the view/hide participants button but displays the save button when the form is dirty
- Call `onBulkAddSharingPermission` on-save. This method calls getCheckedParticipantTypeIds which converts the form data into a list of list of selected participant ids.  These ids are printed to console in the storybook.
- Add className to collapsible to allow override of the border styling

**When there are no sharing participants**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/6863d1fd-9aa1-440d-819c-86ff10c446a9

**When there are sharing participants** (note: no warning banner since it is hardcoded to false)

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/ca84bfef-b19c-4260-95a5-7c08333b2f86


src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -2,31 +2,35 @@ import { useState } from 'react';
...
 import './BulkAddPermissions.scss';
 
 type BulkAddPermissionsProps = {
   participant: ParticipantDTO | null;
+  participantTypes: ParticipantTypeDTO[];
+  sharingParticipants: AvailableParticipantDTO[];
   onBulkAddSharingPermission: (selectedTypes: number[]) => Promise<void>;
 };
 
-export type BulkAddPermissionsForm = {
```
alex-yau-ttd, line 21: Moved this to the helper file to avoid a circular dependency. Happy to change if there is a better way

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -42,70 +46,136 @@ export function BulkAddPermissions({
...
 
-  const { register, handleSubmit } = formMethods;
+  const {
+    register,
+    handleSubmit,
+    watch,
+    formState: { isDirty },
+  } = formMethods;
+
+  const watchPublisherChecked = watch('publisherChecked');
```
alex-yau-ttd, line 56: Haven't used react-hook-form much before, so not sure if this is the best way to get the form values.

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -42,70 +46,136 @@ export function BulkAddPermissions({
...
-                : 'View Recommended Participants'}
-            </button>
-          </div>
-          <div className='bulk-add-permissions-footer'>
-            <button type='submit' className='primary-button bulk-add-permissions-submit-button'>
-              Save Recommendations
-            </button>
-          </div>
-        </div>
+        {!sharingParticipants && (
```
alex-yau-ttd, line 163: This probably isn't the best way to implement this - I'm open to suggestions!

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -2,31 +2,35 @@ import { useState } from 'react';
...
   getRecommendationMessageFromTypeNames,
+  hasUncheckedDefaultParticipantTypeRecommendation,
 } from './bulkAddPermissionsHelpers';
 
 import './BulkAddPermissions.scss';
 
 type BulkAddPermissionsProps = {
   participant: ParticipantDTO | null;
+  participantTypes: ParticipantTypeDTO[];
+  sharingParticipants: AvailableParticipantDTO[];
```
jingyi-gao-ttd, line 26: Do we need the value of sharingParticipants? could we just pass a boolean instead?

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -42,70 +46,135 @@ export function BulkAddPermissions({
...
+        ) && (
+          <div className='remove-recommended-type-warning'>
+            <Banner
+              type='warning'
+              message='If you remove the sharing permissions for a participant type, all sharing permissions of that type are removed, including future participants of that type.'
+            />
+          </div>
+        )}
+      </div>
+      {isDirty && saveRecommendationsButton}
```
jingyi-gao-ttd, line 154: Do we need to share the button if no changes is made?

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
```diff
@@ -40,6 +48,36 @@ export const getRecommendedTypeFromParticipant = (participantTypeNames: string[]
   return formatStringsWithSeparator(result);
 };
 
+export const hasUncheckedDefaultParticipantTypeRecommendation = (
```
jingyi-gao-ttd, line 51: In my understanding, we only show the warning if user is sharing with the type and now they uncheck any of the value?

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -42,70 +46,135 @@ export function BulkAddPermissions({
...
+      <button type='submit' className='primary-button bulk-add-permissions-submit-button'>
+        Save Recommendations
+      </button>
+    </div>
+  );
+
+  const recommendationContent = (
+    <div className='bulk-add-permissions'>
+      <div className='bulk-add-permissions-body'>
+        {bulkAddPermissionsMessage}
```
jingyi-gao-ttd, line 126: nit: I could see the common components used between recommendationContent and collapsibleContent. I wonder in this case if we change `content` props to become `children`
then we could do 
```
<Collapsible>
    <div className='bulk-add-permissions'>
      <div className='bulk-add-permissions-body'>
        {bulkAddPermissionsMessage}
        {participantTypeCheckboxes}
       {sharingParticipant && <Warning>}
{saveRecommendationsButton}
      </div>
</Collapsible>
```
is it easier to mantain?

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: The design shows a state in which the button isn't shown if no changes are made, unless I am misinterpreting something.

image

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
alex-yau-ttd: yes, with the assumption that the ones that are checked are checked because they align with the recommended type given their participant type. So essentially only if they are already sharing with the recommended types and uncheck one of those recommended types.

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: yeah it's a good question, I was wondering if there was a better way. I like the idea, though in principle it would be a bit more complicated, something like 
```
    <div className='bulk-add-permissions'>
      <div className='bulk-add-permissions-body'>
        {bulkAddPermissionsMessage}
        {participantTypeCheckboxes}
        {hasSharingParticipants &&
          hasUncheckedDefaultParticipantTypeRecommendation(
            currentParticipantTypeNames,
            watchPublisherChecked,
            watchAdvertiserChecked,
            watchDSPChecked,
            watchDataProviderChecked
          ) && (
            <div className='remove-recommended-type-warning'>
              <Banner
                type='warning'
                message='If you remove the sharing permissions for a participant type, all sharing permissions of that type are removed, including future participants of that type.'
              />
            </div>
          )}
        {viewRecommendedParticipantsButton}
      </div>
      {!hasSharingParticipants || (hasSharingParticipants && isDirty && saveRecommendationsButton)}
    </div>
```

Plus the need to alter the styling based on the two states and the `defaultOpen` prop. 

It's not clear to me which one would be better - what do you think?

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: good idea - done!

src/web/components/SharingPermission/BulkAddPermissions.tsx
jingyi-gao-ttd: Oh I see, I didn't find this design before

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
jingyi-gao-ttd: Hmmm could we double check with Kimberly about this? So if we are having a participant currently sharing with a type which is not a recommanded type. and later, they deceided to uncheck the checkbox, are we showing this or not?

src/web/components/SharingPermission/BulkAddPermissions.tsx
jingyi-gao-ttd: Ah, I see your point. Yep, passing with component is better this time. But could we still rename the `content` to `children` I think in that case it will easier to use later

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
alex-yau-ttd: yep I will double check!

alex-yau-ttd: The more I think about it I think I have probably misinterpreted the design - I think we probably just want to show the warning whenever we remove any type, but I have asked on Figma anyway.

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: cool - so to clarify - leave the implementation basically as is, but pass it in as `children` instead of the `content` prop?

jingyi-gao-ttd: That's right

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
alex-yau-ttd: This brings up another question which I have asked in Figma:
let's say this participant is a publisher and has shared with Advertisers and DSPs (despite DSPs being the only recommended type). When the user sees this screen, will DSPs AND Advertisers be checked here? Or just DSPs since that is the recommended type?

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: done!

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
jingyi-gao-ttd: In my understanding, the recommended component only shows when this participant hasn't shared with any specific participant or type. 
When a participant shares with a specific participant or type (which means now we have that sharing permission table and the top 2 components get collapsed), this bulk add component should only show the current type they are sharing with, in your case both DSPs and Advertisers should be checked

src/web/components/SharingPermission/BulkAddPermissions.tsx
lionell-pack-ttd: Yeah - I'm still not sold on the pattern, but it's the way react-hook-form provides 🤷

src/web/components/SharingPermission/BulkAddPermissions.tsx
lionell-pack-ttd: It's a trade-off - on one hand, we save some vertical space if they haven't changed anything. On the other hand, this means the content below will all shift up and down as the user makes changes/reverts their changes. 🤷

I think wait and see if the reflow becomes annoying - if it is annoying, there might be a case for always having the button, but leaving it disabled if there are no changes.

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -42,70 +44,129 @@ export function BulkAddPermissions({
...
+      {saveRecommendationsButton}
+    </div>
+  );
+
+  const collapsibleContent = (
+    <div className='bulk-add-permissions'>
+      <div className='bulk-add-permissions-body'>
+        {bulkAddPermissionsMessage}
+        {participantTypeCheckboxes}
+        {/* Change this condition once we have APIs to get the which types have been shared with */}
```
alex-yau-ttd, line 137: hey @jingyi-gao-ttd, in an attempt to merge these changes in while blocked on the APIs, I have hardcoded the warning banner to never show, since the logic I had before was incorrect. The correct logic requires the APIs. Is that okay with you?

jingyi-gao-ttd: That sounds reasonable to me

jingyi-gao-ttd (approved):

src/web/components/SharingPermission/bulkAddPermissionsHelpers.ts
alex-yau-ttd: yes you are right. I will address this in a future MR.

PR #123, opened 2023-08-03
UID2-1029 Create Audit trail table and migrate existing sharing audit trail

jingyi-gao-ttd:
- Create a new Audit trail table which could pass in a json as event data
- Migrate all existing sharing Audit trails into new table (notice: the siteId will be missing for those legacy data)
- Update sharing permission endpoints to record audit trail in new table


src/api/entities/AuditTrail.ts
```diff
@@ -0,0 +1,55 @@
...
+export enum SharingAction {
+  Add = 'add',
+  Delete = 'delete',
+}
+
+export enum AuditTrailEvents {
+  UpdateSharingPermissions = 'UpdateSharingPermissions',
+}
+
+export type AuditTrailEventData = UpdateSharingPermissionEventData | string;
```
jingyi-gao-ttd, line 15: I realise that when we retrieve json column from db, it will not convert it into an object, but keep it in string. Maybe provide a get method here?

src/database/migrations/20230803014224_CreateAuditTrailsTableAndMigrateData.ts
```diff
@@ -0,0 +1,120 @@
...
+    .select(knex.raw(`string_agg(sharingParticipantSiteId, ',') siteIds`), ...groupByColumns)
+    .groupBy(groupByColumns);
+  const auditTrails = sharingAuditTrails.map(mapSharingAuditTrailToAuditTrails);
+  await knex('auditTrails').insert(auditTrails);
+}
+
+function mapAuditTrailsToSharingAuditTrails(
+  auditTrail: AuditTrailDTO & { created_at: Date; updated_at: Date }
+): SharingAuditTrail[] {
+  const eventData = JSON.parse(auditTrail.eventData as string) as UpdateSharingPermissionEventData;
```
jingyi-gao-ttd, line 65: Feel a bit annoying if we always need to do JSON.parse on auditTrail.eventData. Open to suggestions

src/api/entities/AuditTrail.ts
```diff
@@ -0,0 +1,59 @@
...
+      relation: Model.BelongsToOneRelation,
+      modelClass: 'User',
+      join: {
+        from: 'auditTrails.userId',
+        to: 'users.id',
+      },
+    },
+  };
+
+  static get jsonAttributes() {
```
jingyi-gao-ttd, line 46: Ref: https://vincit.github.io/objection.js/api/model/static-properties.html#static-jsonattributes

src/api/entities/AuditTrail.ts
```diff
@@ -0,0 +1,59 @@
...
+    return ['eventData'];
+  }
+
+  id!: number;
+  userId!: number;
+  participantId!: number;
+  userEmail!: string;
+  proceed!: boolean;
```
lionell-pack-ttd, line 54: I don't think this really describes what the field is for. Maybe 'succeeded' or 'completed' would be more descriptive?

src/api/services/auditTrailService.ts
```diff
@@ -0,0 +1,35 @@
...
+      participantId: participant.id,
+      userId,
+      userEmail,
+      event: AuditTrailEvents.UpdateSharingPermissions,
+      eventData: {
+        siteId: participant.siteId,
+        action,
+        sharingPermissions: siteIds,
+      },
+    } as AuditTrail;
```
lionell-pack-ttd, line 23: Try to avoid `as` whenever possible - in this case, it's masking a possible bug (you can pass in a participant which doesn't have a siteId). When you use `as` you're telling TypeScript to skip checking the types (although sometimes the linter still finds things).

Better to declare the variable with a type - e.g. `const sharingAuditTrail: AuditTrail = {`. That way TypeScript knows you expect something of type AuditTrail and it'll check that what you assign matches.

lionell-pack-ttd (approved):

PR #126, opened 2023-08-06
UID2-1029 Create audit trail when account is approved

jingyi-gao-ttd:
image

### DIFF
--- a/src/api/entities/AuditTrail.ts
+++ b/src/api/entities/AuditTrail.ts
@@ -23,6 +23,10 @@ export type UpdateSharingPermissionEventData = {
 
 export type ApproveAccountEventData = {
   siteId: number;
+  oldName?: string;
+  newName?: string;
+  oldTypeIds?: number[];
+  newTypeIds?: number[];
 };
 
 export class AuditTrail extends BaseModel {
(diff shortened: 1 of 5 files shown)


src/api/entities/AuditTrail.ts
```diff
@@ -10,16 +10,21 @@ export enum SharingAction {
...
+export type AuditTrailEventData = UpdateSharingPermissionEventData | ApproveAccountEventData;
 
 export type UpdateSharingPermissionEventData = {
   siteId: number;
   action: SharingAction;
   sharingPermissions: number[];
 };
 
+export type ApproveAccountEventData = {
+  siteId: number;
```
lionell-pack-ttd, line 25: From the parser on the API endpoint, it looks like a few things can be changed as well - name and types. I think it would be good to record any changes to those as well as the site ID being chosen.

jingyi-gao-ttd: That's a good point. Should we record both? or just the original name and type? as the new name and type will be stored in the participant

lionell-pack-ttd: I'd probably prefer to see both what the values were, and the new ones - so if someone gets approved with changed details and there's a query raised about it, we can find out how it was changed and what it was originally.

jingyi-gao-ttd: Name and typeIds are now added to the event data

lionell-pack-ttd (approved):

PR #137, opened 2023-08-10
Llp uid2 1030 refactor search bar

lionell-pack-ttd:


src/web/components/Search/SearchBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+        {...rest}
+      />
+    </div>
+  );
+}
+
+type SearchBarResultsProps = React.PropsWithChildren<{ className?: string }>;
+export function SearchBarResults({ children, className }: SearchBarResultsProps) {
+  return <div className={clsx('search-bar-results', className)}>{children}</div>;
+}
```
jingyi-gao-ttd, line 37: May I ask why this need to be wrap into a component?

jingyi-gao-ttd (approved):

src/web/components/Search/SearchBar.tsx
lionell-pack-ttd: You can provide it or not. You can style it separately. You can nest it inside something else if you want to. This way just makes it nice and flexible - it's the same pattern that Radix uses a lot.

jingyi-gao-ttd: I see, thanks.

PR #140, opened 2023-08-11
UID2-1566 add sorting to sharing permission table

jingyi-gao-ttd:
This PR introduces two main components: `SortableProvider` and `SortableTableHeader`

SortableProvider: A context provider that wraps around a table, allowing headers to tap into sorting logic and state.

SortableTableHeader: A header component that, when provided a sortKey, allows users to toggle the sorting order of its associated column. You can also provide your own sorting logic via the customSortFn 

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/89bb350d-911f-4dfb-a779-ba01a836db8d


alex-yau-ttd:
wondering if we could possibly adjust the styling so that the up and down arrows are still aligned with the text 
image

jingyi-gao-ttd:
image

@alex-yau-ttd  That's a good point. I've added align center to it, but as you can see, I think those icons are designed to be the same height, the `sort-up` and `sort-down` arrows are aligned with arrows in `sort` icon,

alex-yau-ttd:
> Screen.Recording.2023-08-21.at.4.08.38.pm.mov 
> @alex-yau-ttd That's a good point. I've added align center to it, but as you can see, I think those icons are designed to be the same height, the `sort-up` and `sort-down` arrows are aligned with arrows in `sort` icon,

thanks for explaining - seems reasonable to me for now, we can always play with styling later if necessary.

src/web/components/SharingPermission/ParticipantsTable.tsx
```diff
@@ -115,3 +121,14 @@ export function ParticipantsTable({
     </table>
   );
 }
+
+export function ParticipantsTable(props: ParticipantsTableProps) {
+  return (
+    <SortableProvider>
+      <ParticipantsTableContent
+        // eslint-disable-next-line react/jsx-props-no-spreading
```
alex-yau-ttd, line 129: If you merge in `main` you should be able to get rid of this comment - we removed these warnings last week

alex-yau-ttd (approved):

PR #146, opened 2023-08-21
ajy-UID2-1596-Make-handleOnBlur-optional-in-SearchBar

alex-yau-ttd:
Managed to break main (https://github.com/IABTechLab/uid2-self-serve-portal/actions/runs/5922755162/job/16057208608) by having a `SearchBarContainer` that did not provide a `handleOnBlur` method. It probably makes the most sense to set this as optional for now.

In this version, the event listener is added only if `handleOnBlur` is provided and is used as a dependency in the `useEffect` hook. This ensures that the event listener is added and removed properly when `handleOnBlur` changes.

### DIFF
--- a/src/web/components/Search/SearchBar.tsx
+++ b/src/web/components/Search/SearchBar.tsx
@@ -6,24 +6,26 @@ import './SearchBar.scss';
 
 type SearchBarContainerProps = React.PropsWithChildren<{
   className?: string;
-  handleOnBlur: () => void;
+  handleOnBlur?: () => void;
 }>;
 export function SearchBarContainer({ children, className, handleOnBlur }: SearchBarContainerProps) {
   const componentRef = useRef<HTMLDivElement>(null);
-  const handleClick = (event: MouseEvent) => {
-    if (componentRef.current && !componentRef.current.contains(event.target as Node)) {
-      handleOnBlur();
-    }
-  };
 
   useEffect(() => {
+    const handleClick = (event: MouseEvent) => {
+      if (componentRef.current && !componentRef.current.contains(event.target as Node)) {
+        if (handleOnBlur) {
+          handleOnBlur();
+        }
+      }
+    };
+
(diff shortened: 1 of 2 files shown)


src/web/components/Search/SearchBar.tsx
```diff
@@ -6,24 +6,26 @@ import './SearchBar.scss';
...
+    };
+
     document.addEventListener('click', handleClick);
 
     return () => {
       document.removeEventListener('click', handleClick);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [handleOnBlur]);
```
jingyi-gao-ttd, line 28: I found this useEffect is triggered every time you have a click, not sure if this the correct way to do it

jingyi-gao-ttd: I tested that wrapped by useCallback could fix this
handleOnBlur={useCallback(() => onToggleOpen(false), [onToggleOpen])}

otherwise we are having unnecessary  re-attachments of the event listener

src/web/components/Search/SearchBar.tsx
```diff
@@ -6,24 +6,26 @@ import './SearchBar.scss';
...
-  const handleClick = (event: MouseEvent) => {
-    if (componentRef.current && !componentRef.current.contains(event.target as Node)) {
-      handleOnBlur();
-    }
-  };
 
   useEffect(() => {
+    const handleClick = (event: MouseEvent) => {
+      if (componentRef.current && !componentRef.current.contains(event.target as Node)) {
+        if (handleOnBlur) {
```
jingyi-gao-ttd, line 17: could we check this before we add the event listener?

alex-yau-ttd: great catch - does that look right now?

jingyi-gao-ttd (approved):

PR #151, opened 2023-08-23
ajy-UID2-1617-Participant-item-simple

alex-yau-ttd:
* ParticipantItem currently has a checkbox as well as an `addedBy` prop that is not needed in all use cases (e.g. Bulk Add Permissions control). 
* The new ParticipantItemSimple component just has the logo, name and associated types, and ParticipantItem now has a ParticipantItemSimple.

image

### DIFF
--- a/src/web/components/SharingPermission/ParticipantItem.tsx
+++ b/src/web/components/SharingPermission/ParticipantItem.tsx
@@ -6,35 +6,24 @@ import { TriStateCheckbox } from '../Core/TriStateCheckbox';
 
 import './ParticipantItem.scss';
 
-type ParticipantItemProps = {
+function getParticipantTypes(participantTypes?: z.infer<typeof ParticipantTypeSchema>[]) {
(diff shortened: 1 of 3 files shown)


src/web/components/SharingPermission/ParticipantItem.tsx
```diff
@@ -6,27 +6,51 @@ import { TriStateCheckbox } from '../Core/TriStateCheckbox';
...
+
+type ParticipantItemSimpleProps = {
   participant: AvailableParticipantDTO;
+};
+
+export function ParticipantItemSimple({ participant }: ParticipantItemSimpleProps) {
+  const logo = '/default-logo.svg';
+
+  return (
+    <>
```
alex-yau-ttd, line 26: ideally we'd stick a classname here but we can't since we can't put a <div> here. this means that we lose some structure in the scss

jingyi-gao-ttd (approved):

PR #152, opened 2023-08-23
ajy-UID2-List-participants-by-type-in-bulk-add

alex-yau-ttd:
* Show participants according to checked types
* Show the "View Participants" button when a type is checked
* Some copy changes

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/53754bd1-c519-4abc-bea4-cf42ef9a7801

**Future work**
* Pass the participants into this component, will need to come from the Sites from Admin. We will need to use siteId to map from a SiteDTO to an AvailableParticipantDTO.

### DIFF
--- a/src/web/components/SharingPermission/BulkAddPermissions.tsx
+++ b/src/web/components/SharingPermission/BulkAddPermissions.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 
 import { ParticipantDTO } from '../../../api/entities/Participant';
@@ -39,7 +39,6 @@ export function BulkAddPermissions({
   onBulkAddSharingPermission,
 }: BulkAddPermissionsProps) {
   const [showRecommendedParticipants, setShowRecommendedParticipants] = useState(false);
-  const [filteredParticipants, setFilteredParticipants] = useState(availableParticipants);
   const currentParticipantTypeNames = participant?.types
     ? participant.types.map((p) => p.typeName ?? '')
     : [];
@@ -83,15 +82,13 @@ export function BulkAddPermissions({
     );
   }, [watchPublisherChecked, watchAdvertiserChecked, watchDSPChecked, watchDataProviderChecked]);
 
-  useEffect(() => {
-    setFilteredParticipants(
-      getFilteredParticipantsByType(
-        availableParticipants,
-        watchPublisherChecked,
-        watchAdvertiserChecked,
-        watchDSPChecked,
-        watchDataProviderChecked
-      )
+  const filteredParticipants = useMemo(() => {
+    return getFilteredParticipantsByType(
+      availableParticipants,
+      watchPublisherChecked,
+      watchAdvertiserChecked,
+      watchDSPChecked,
+      watchDataProviderChecked
     );
   }, [
     availableParticipants,
--- a/src/web/components/SharingPermission/BulkAddPermissions.tsx
+++ b/src/web/components/SharingPermission/BulkAddPermissions.tsx
@@ -1,9 +1,11 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 
 import { ParticipantDTO } from '../../../api/entities/Participant';
+import { AvailableParticipantDTO } from '../../../api/routers/participantsRouter';
 import { Banner } from '../Core/Banner';
 import { Collapsible } from '../Core/Collapsible';
+import { withoutRef } from '../Core/Form';
 import { FormStyledCheckbox } from '../Input/StyledCheckbox';
 import {
   BulkAddPermissionsForm,
@@ -12,16 +14,19 @@ import {
   getDefaultDataProviderCheckboxState,
   getDefaultDSPCheckboxState,
   getDefaultPublisherCheckboxState,
+  getFilteredParticipantsByType,
   getRecommendationMessageFromTypeNames,
   getRecommendedTypeFromParticipant,
   hasUncheckedASharedType,
 } from './bulkAddPermissionsHelpers';
+import { ParticipantItemSimple } from './ParticipantItem';
 
 import './BulkAddPermissions.scss';
 
 type BulkAddPermissionsProps = {
   participant: ParticipantDTO | null;
   hasSharingParticipants: boolean;
+  availableParticipants: AvailableParticipantDTO[];
   onBulkAddSharingPermission: (selectedTypes: string[]) => Promise<void>;
   sharedTypes: string[];
 };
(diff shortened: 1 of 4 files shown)


src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -12,28 +14,32 @@ import {
...
   hasUncheckedASharedType,
 } from './bulkAddPermissionsHelpers';
+import { ParticipantItemSimple } from './ParticipantItem';
 
 import './BulkAddPermissions.scss';
 
 type BulkAddPermissionsProps = {
   participant: ParticipantDTO | null;
   hasSharingParticipants: boolean;
+  availableParticipants: AvailableParticipantDTO[];
```
jingyi-gao-ttd, line 29: I guess these participants here are getting back from admin service, do they AvailableParticipantDTO type?

alex-yau-ttd: good question - these participants need to be driven from the sites from admin service. my plan is for sharingPermissions.tsx (or elsewhere) to map these sites to AvailableParticipantDTOs (or similar), then pass them into this component. We need to do this because the source of truth for the SiteId comes from admin, but the display name for the participant lives in our db.  So admin doesn't need to have a notion of AvailableParticipantDTO. Does that sound reasonable?

jingyi-gao-ttd: I understand there will eventually be 1 to 1 relationship between the site and the participant. But if a participant owner didn't create an account in self-serve portal, I think we probably still want to display that site in this list. Because technically when current participant share with this type, they will share with all sites no matter they have account in self-serve portal?

alex-yau-ttd: that is correct. do you see any issue with this implementation to achieve that? I understand that it may not necessarily use `AvailableParticipantDTO` type but it would be something similar. Or maybe could have a different name for the prop for clarity?

jingyi-gao-ttd: What I mean is if we don't have such site lives in our db, how are we going to map it to the AvailableParticipantDTOs? do you reformat the object that we get from admin service to AvailableParticipantDTO type

alex-yau-ttd: maybe my use of AvailableParticipantDTO is confusing. My understanding is all we need to do once we have the SiteDTO is to retrieve the participant name & logo by joining to our Participants on SiteId. If there is no associated siteId in our db, then we just display the name of the site (I suppose). I was thinking this could reuse the AvailableParticipantDTO, essentially what we need to end up with is:
- Participant Name
- Types
- Logo
- Id (not strictly necessary but I think makes sense).

fwiw by the time we get to BulkAddPermissions I don't think we even need SiteId, since this control is type-driven rather than siteId-driven.

How's that sound?

jingyi-gao-ttd: Oh I see, thanks for the explanation, re-use AvailableParticipantDTO make sense in this case

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -12,28 +14,32 @@ import {
...
   onBulkAddSharingPermission: (selectedTypes: string[]) => Promise<void>;
   sharedTypes: string[];
 };
 
 export function BulkAddPermissions({
   participant,
   hasSharingParticipants,
+  availableParticipants,
```
jingyi-gao-ttd, line 37: We are retrieving all sites from the admin service? and based on the selected type to filter on them?

alex-yau-ttd: that is the plan yes! load all of them in and then just filter by type within this component

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -71,19 +77,42 @@ export function BulkAddPermissions({
...
 
+  const hasCheckedType = useCallback(() => {
+    return (
+      watchPublisherChecked || watchAdvertiserChecked || watchDSPChecked || watchDataProviderChecked
+    );
+  }, [watchPublisherChecked, watchAdvertiserChecked, watchDSPChecked, watchDataProviderChecked]);
+
+  useEffect(() => {
+    setFilteredParticipants(
+      getFilteredParticipantsByType(
```
jingyi-gao-ttd, line 88: nit: I guess filteredParticipant could managed by useMemo hook? As  we are memoizing the result of an expensive computation

jingyi-gao-ttd (approved):

src/web/components/SharingPermission/BulkAddPermissions.tsx
alex-yau-ttd: hm what would that look like exactly? also where do we draw the line for expensive computation? the `getFilteredParticipantsByType` has linear time complexity, and will comprised of all siteIds which will be quite a long list (at least in future if it isn't already)

alex-yau-ttd: I've tried:
```
  const filteredParticipantsByType = useMemo(() => {
    return getFilteredParticipantsByType(
      availableParticipants,
      watchPublisherChecked,
      watchAdvertiserChecked,
      watchDSPChecked,
      watchDataProviderChecked
    );
  }, [
    availableParticipants,
    watchPublisherChecked,
    watchAdvertiserChecked,
    watchDSPChecked,
    watchDataProviderChecked,
  ]);

  useEffect(() => {
    console.log('filteredParticipantsByType', filteredParticipantsByType);
    setFilteredParticipants(filteredParticipantsByType);
  }, [filteredParticipantsByType]);
```
but it still seems to hit the useEffect even if `filteredParticipantsByType` returns the same participants (maybe because of a deep equality check?).

tbh I haven't used `useMemo` or `useEffect` that much so open to ideas :)

jingyi-gao-ttd: When we have useMemo, we no longer need to manage the state filteredParticipantsByType. So the following could be removed
```
const [filteredParticipants, setFilteredParticipants] = useState(availableParticipants);

useEffect(() => {
    console.log('filteredParticipantsByType', filteredParticipantsByType);
    setFilteredParticipants(filteredParticipantsByType);
  }, [filteredParticipantsByType]);
```

and it will be 
```
  const filteredParticipants = useMemo(() => {
    return getFilteredParticipantsByType(
      availableParticipants,
      watchPublisherChecked,
      watchAdvertiserChecked,
      watchDSPChecked,
      watchDataProviderChecked
    );
  }, [
    availableParticipants,
    watchPublisherChecked,
    watchAdvertiserChecked,
    watchDSPChecked,
    watchDataProviderChecked,
  ]);

          {filteredParticipants.map((p) => (
            <tr key={p.id}>
              <ParticipantItemSimple participant={p} />
            </tr>
          ))}
```

https://stackoverflow.com/questions/56347639/react-useeffect-vs-usememo-vs-usestate

alex-yau-ttd: ahh gotcha! so this will ensure that React will return the same value on subsequent renders given that the dependencies haven't changed. though as I understand it, this same behaviour can be obtained by using `useState` and `useEffect` together, although using `useMemo` makes it easier to understand. Also `useMemo` runs during the render while `useState` causes a re-render and `useEffect` occurs after each render.

done

PR #153, opened 2023-08-24
Sas UI d2 1579 log out on accepted invite

ssundahlTTD:
This bug presents itself when testing the application.  When a user invites a new Team Member, then proceeds to accept that invitation themselves, they are still logged in as the original user when they accept the invite.  This change will log out the current user when an invitation is accepted.

1. From the team members page, add a team member or resend the invite to an existing member.
2. In MailHog, click the Accept Invitation button.
3. Create a new password and save
4. If you hover over the "Back to Application" link, you will notice it now goes to a /Logout route.
5. Click "Back to Application" and verify that you are logged out.

The new /logout route should be added to SSP_WEB_LOGOUT_URL in the .env file


src/web/screens/logout.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import './home.scss';
+
+function Logout() {
+  const { keycloak } = useKeycloak();
+  keycloak?.logout();
+  return <h1>Logging out...</h1>;
+}
+export const LogoutRoute: PortalRoute = {
+  path: '/Logout',
```
jingyi-gao-ttd, line 13: nit: could we change it to lowercase? just to be consistent with other path

jingyi-gao-ttd (commented):
Could you please add `Add SSP_WEB_LOGOUT_URL Env` into  DeploymentSteps.md?

src/web/screens/logout.tsx
ssundahlTTD: changed

ssundahlTTD: Added instruction to add the route to the .env file

jingyi-gao-ttd (approved):

PR #155, opened 2023-08-28
jyg-UID2-1475 Multi select dropdown filter

jingyi-gao-ttd:
https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/89fc0911-d303-409c-98f7-d332cc30953c

### DIFF
--- a/src/web/components/Core/MultiSelectDropdown.scss
+++ b/src/web/components/Core/MultiSelectDropdown.scss
@@ -1,7 +1,7 @@
 .multi-select-dropdown {
   & > button {
     border-radius: 16px;
-    border: 1px solid #0300a5;
+    border: 1px solid var(--theme-radio-background);
     padding: 5px 12px;
   }
 
--- a/src/web/components/Core/MultiSelectDropdown.scss
+++ b/src/web/components/Core/MultiSelectDropdown.scss
@@ -5,6 +5,11 @@
     padding: 5px 12px;
   }
 
+  &.active > button {
+    background-color: var(--theme-radio-background);
+    color: var(--theme-background-content);
+  }
+
   &-trigger svg {
     margin-left: 10px;
   }
--- a/src/web/components/Core/MultiSelectDropdown.scss
+++ b/src/web/components/Core/MultiSelectDropdown.scss
@@ -1,0 +1,52 @@
+.multi-select-dropdown {
+  & > button {
+    border-radius: 16px;
+    border: 1px solid #0300a5;
+    padding: 5px 12px;
+  }
+
+  &-trigger svg {
+    margin-left: 10px;
+  }
+
+  &-content {
+    width: var(--radix-dropdown-menu-trigger-width);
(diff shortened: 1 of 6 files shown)


src/web/components/Core/MultiSelectDropdown.scss
```diff
@@ -0,0 +1,52 @@
+.multi-select-dropdown {
+  & > button {
+    border-radius: 16px;
+    border: 1px solid #0300a5;
```
alex-yau-ttd, line 4: I notice this colour matches `--theme-radio-background`. Would it be appropriate to use that variable instead? Or introduce a new one?

src/web/components/Core/MultiSelectDropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+      );
+    },
+    [onOptionToggle, selectedItems]
+  );
+
+  return (
+    <div className='multi-select-dropdown'>
+      <DropdownMenu.Root open={open} onOpenChange={setOpen}>
+        <DropdownMenu.Trigger className='multi-select-dropdown-trigger'>
+          {title}: {selectedCount}
```
alex-yau-ttd, line 74: I see you asked about `selectedCount` in Figma - I also had the same question: should we show the names of the selected options or count of how many. In any case we can merge like this and change later though.

src/web/components/Core/MultiSelectDropdown.scss
jingyi-gao-ttd: good catch, I totally forgot this! Updated

src/web/components/Core/MultiSelectDropdown.tsx
jingyi-gao-ttd: yea, I think list all of them will make this really long/ do you remember in this morning's discussion, are we want to remove the `selectedCount` or not?

alex-yau-ttd: yeah I think we decided to remove `selectedCount`. As for displaying the type names, tbh I don't think it's a problem if it is really long, I think we have plenty of horizontal screen space. In any case, I think it would be a good idea to make some styling change when filters are applied, maybe something like go from white background colour and black text (no selection) to blue background colour and white text (with selection)?

jingyi-gao-ttd: Thanks, Got both title and styling updated

alex-yau-ttd: looks good! one nit: could we possibly add a space after the commas when listing the options? Perhaps `formatStringsWithSeparator` could help, though I know the implementation is a bit different here.

jingyi-gao-ttd: Actually I think it makes sense to reuse that formatStringsWithSeparator! Updated

alex-yau-ttd (approved):

PR #163, opened 2023-08-29
ajy-UID2-1617-Implement-bulk-add-functionality

alex-yau-ttd:
**This branch currently breaks the Sharing Permissions Table by sending it an empty list of participants. Jingyi is going to fix it in a separate PR based on these changes**

**Summary**
- Use the SharingListResponse all the way to the frontend, so we can just have one endpoint that has all the shared site ids and shared types in one hit
- Add `allowed_types` to the response and request to & from Admin
- Call admin api for saving by type and adding/deleting by siteId with correct data
- Pass updated state as props into children (rather than passing `availableParticipants`, we pass `sharedSiteIds` and/or `sharedTypes`
- Minor copy changes
- New tests and storybook data adjustments

**Testing**
- Ensure admin is running locally and setup properly

**Evidence**
- Shows the response after calling save using Bulk Add (adding by type aka `allowed_types`) and Search and Add (adding by siteid aka `allowed_sites`)

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/2029ee2c-22a9-4ddc-8cbe-35d1f82c8dfa


src/api/services/participantsService.ts
```diff
@@ -88,30 +71,30 @@ export const getAttachedSiteIDs = async (): Promise<SiteIdType[]> => {
 
 export const addSharingParticipants = async (
   participantSiteId: number,
-  siteIds: number[]
-): Promise<Participant[]> => {
+  siteIds: number[],
+  types: string[]
```
jingyi-gao-ttd, line 75: nit: should we use clientType[]?

src/api/routers/participantsRouter.ts
```diff
@@ -240,16 +241,18 @@ export function createParticipantsRouter() {
 
       const sharingParticipants = await addSharingParticipants(
         participant.siteId,
-        newParticipantSites
+        newParticipantSites,
+        newTypes
       );
 
       await updateAuditTrailToProceed(auditTrail.id);
-      return res.status(200).json(sharingParticipants.map(mapParticipantToAvailableParticipant));
```
jingyi-gao-ttd, line 247: could we now remove `mapParticipantToAvailableParticipant`?

src/web/components/SharingPermission/BulkAddPermissions.tsx
```diff
@@ -203,7 +200,7 @@ export function BulkAddPermissions({
             {recommendationContent}
           </Collapsible>
         )}
-        {hasSharingParticipants && (
+        {(hasSharedSiteIds || sharedTypes.length > 0) && (
```
jingyi-gao-ttd, line 203: I think the publisher will by default share with DSPs? and I remember we want to add a flag to check if the user has gone through the recommendation? Are you going to do it in a separate PR?

alex-yau-ttd: Yeah those are separate tickets, see: https://atlassian.thetradedesk.com/jira/browse/UID2-1735 and https://atlassian.thetradedesk.com/jira/browse/UID2-1736

src/api/services/participantsService.ts
alex-yau-ttd: To use ClientType in the service, I believe I would need to re-define the client types in the router as an enum like this:
```
  const ClientType = z.enum(['DSP', 'ADVERTISER', 'DATA_PROVIDER', 'PUBLISHER']);

  const sharingRelationParser = z.object({
    newParticipantSites: z.array(z.number()),
    newTypes: z.array(ClientType),
  });
```

Does that sound right to you? Or is there a better way?

Source: https://zod.dev/

jingyi-gao-ttd: I think this looks better than string[]

jingyi-gao-ttd (approved):

src/api/services/participantsService.ts
alex-yau-ttd: cool will do. There is also a bunch of uses of `string[]` in the components which I will update to use `ClientType[]`

src/api/routers/participantsRouter.ts
alex-yau-ttd: done!

src/api/services/participantsService.ts

PR #164, opened 2023-08-30
jyg-UID2-1475 fetch available sites and convert it to available participant dto

jingyi-gao-ttd:
Add a new endpoint for fetching sites from admin and filter the ones that don't have type

To make the new endpoint `/api/sites/available` to return list of participants, make sure 

- [ ] Admin service is running
- [ ] Some sites have types set ( You can set it using`http://localhost:8089/adm/site.html#`)


lionell-pack-ttd:
See my Slack message - I want to make sure we have a clear difference between Site and Participant.

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
```diff
@@ -1,14 +1,40 @@
...
+  },
+  {
+    name: 'Site 4',
+    types: [{ id: 4, typeName: 'Type 4' }],
+    siteId: 13,
+  },
+];
 const Template: ComponentStory<typeof SharingPermissionsTable> = (args) => (
-  <SharingPermissionsTable {...args} />
+  <TestAvailableSiteListProvider response={response}>
```
alex-yau-ttd, line 35: Given we can provide test sites for this storybook using this Provider, I wonder if it is worth removing the storybook `ParticipantSearchBar.stories.tsx` since it is essentially the same thing, but prop driven. I was confused as to what this line of code did because I was accidentally looking at `Sharing Permissions/ParticipantSearchBar` which kept working after commenting out this line 😅 What do you think?

alex-yau-ttd:
LGTM, though I guess we'll wait for Kimberly to confirm as above :)

alex-yau-ttd (approved):

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
jingyi-gao-ttd: I assume you mean `ParticipantSearchBar.stories` and `SearchAndAddParticipants.stories` are the same thing? Happy to remove one of them, but as we have the tests on  ParticipantSearchBar, how about removing `SearchAndAddParticipants.stories` instead?

alex-yau-ttd: hmm actually I was thinking of removing `ParticipantSearchBar.stories` but you are right in that there are tests that depend on it. Let's keep both for now since they are ever so slightly different, we can always revisit later if it is still confusing

alex-yau-ttd (approved):

PR #165, opened 2023-08-30
Sas UI d2 1719 banner toastie message design changes

ssundahlTTD:
Update messages and notifications to match new designs.

image

Add consistent themes and styling to StatusPopup, Banner and InlineMessage (formerly InlineError)

image

image


src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
+
+    svg {
+      color: #000;
+      background: transparent;
+    }
+  }
 }
 
 .status-popup-message {
   margin-left: 22px;
```
alex-yau-ttd, line 47: nit: the message doesn't exactly line up with the title. I think it should be 24px here

src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
   font-weight: 400;
-  font-size: 0.75rem;
+  font-size: 1rem;
   line-height: 1rem;
```
alex-yau-ttd, line 50: nit: the line height looks a bit small in the screenshot where the message is two lines. Maybe change to `line-height: 1.3;`?

src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
+  border-radius: 10px;
 }
 
 .status-popup-title-container {
   margin-bottom: 4px;
-  &.Success {
-    color: var(--theme-success);
-  }
+  display: flex;
+  justify-content: space-between;
```
alex-yau-ttd, line 17: nit: I think we could also do with a 
```
align-items: center;
```
to ensure the close button lines up with the title

src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
 
-  &.Info {
-    color: var(--theme-action);
+  .status-popup-title {
+    margin-left: 8px;
+    font-weight: 700;
+    font-size: 1rem;
+    line-height: 1rem;
+
+    .status-popup-title-left {
```
alex-yau-ttd, line 29: I don't think this css is actually being applied: the DOM has `status-popup-title` inside `status-popup-title-left` rather than the other way around. Perhaps we could get rid of this className?

src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
+    font-size: 1rem;
   }
 
-  &.Info {
-    color: var(--theme-action);
+  .status-popup-title {
+    margin-left: 8px;
+    font-weight: 700;
+    font-size: 1rem;
+    line-height: 1rem;
```
alex-yau-ttd, line 27: nit: I know you didn't introduce this line, but might be a good time to set a unitless line-height rather than using rem:
```
line-height: 1;
```

src/web/components/Core/StatusPopup.scss
```diff
@@ -2,64 +2,50 @@
...
-  font-weight: 700;
-  font-size: 0.8125rem;
-  line-height: 1rem;
+  .popup-close-button {
+    background: none;
+    border: none;
+    padding: 0;
+
+    svg {
+      color: #000;
```
alex-yau-ttd, line 40: nit: we may want to use 
```
color: var(--theme-primary);
```
instead here. Will help for when we implement dark mode too

src/web/components/Core/StatusPopup.scss
ssundahlTTD: fixed

src/web/components/Core/StatusPopup.scss
ssundahlTTD: This doesn't alight it the same way:  
image

src/web/components/Core/StatusPopup.scss
ssundahlTTD: good call.  i got carried away with keeping things symmetric.

src/web/components/Core/StatusPopup.scss
ssundahlTTD: Fixed.

src/web/components/Core/StatusPopup.scss
ssundahlTTD: changed.

src/web/components/Core/StatusPopup.tsx
```diff
@@ -2,15 +2,21 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { useEffect } from 'react';
```
alex-yau-ttd, line 2: I've noticed that if the user triggers a second Popup before the first one goes away, the second one goes away very quickly because the timer hasn't reset. Could you please add `message` and `status` as dependencies to the timer `useEffect` in this file so the timer resets when the message & status change?

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/629e8511-8de5-4929-a6a7-8e9c2ed49f19

src/web/components/Core/StatusPopup.scss
ssundahlTTD: changed

src/web/components/Core/StatusPopup.scss
alex-yau-ttd: hmm that's weird, mine looks like this

image

ssundahlTTD: My bad.  I replaced one rule with the other rather than having them both.  I changed this.

src/web/components/Core/StatusPopup.tsx
ssundahlTTD: In the case of your example, neither status nor message will change since you're deleting twice.  I'll try to think of something else on addition to those.

alex-yau-ttd: yeah good point, I made that change locally then tested with two different actions 🤦 .  maybe some sort of id or something

src/web/styles/themes.scss
```diff
@@ -25,19 +25,23 @@
   --theme-info-icon: #b7c2ea;
   --theme-info-content: #035959;
```
alex-yau-ttd, line 26: looks like we can delete these two vars. `--theme-info-icon` is overridden by your new one on line 33 and `--theme-info-content` is unused

ssundahlTTD: removed.

src/web/components/Core/StatusPopup.tsx
ssundahlTTD: I'm going to address this in a separate PR.

alex-yau-ttd (approved):

PR #166, opened 2023-08-31
Uid2 1030 approval search

lionell-pack-ttd:
The layout isn't ideal (height of the dialog) but I've tried a bunch of things and everything is some kind of compromise - given this is an internal screen, I don't think it's worth spending more time on.

chrome_RvglU8bCKJ


src/web/components/ParticipantRequests/ParticipantApprovalForm.tsx
```diff
@@ -94,6 +145,23 @@ function ParticipantApprovalForm({
...
               }))}
             />
+            <Input inputName='readonly-requestor-name' label='Requestor Name'>
+              <input
+                className='input-container'
+                disabled
+                value={participant.requestingUser.fullName}
+              />
+            </Input>
+            <Input inputName='readonly-requestor-email' label='Requestor Email'>
```
jingyi-gao-ttd, line 155: Can we use `<TextInput inputName='readonly-requestor-email' label='Requestor Email' value={participant.requestingUser.email} disabled/>` instead?

src/api/services/participantsService.ts
```diff
@@ -67,10 +73,20 @@ export const getParticipantsAwaitingApproval = async (email: string): Promise<Pa
         .select('participantId')
     )
     .withGraphFetched('types')
+    .withGraphFetched('users')
```
jingyi-gao-ttd, line 76: nit: I think we could do `.withGraphFetched('[types, users]')`, maybe they have performance as chain it up

src/database/seeds/Participants.ts
```diff
@@ -7,7 +7,7 @@ import { Participant, ParticipantStatus } from '../../api/entities/Participant';
 type ParticipantsType = ModelObject<Participant>;
 const sampleData: Optional<
```
jingyi-gao-ttd, line 8: I wonder if we need to insert a user for the `AwaitingApproval example` participant. Otherwise, we will have a participantRequest without a user. I think this `const firstUser = participant.users.sort((a, b) => a.id - b.id)[0];` will break?

src/api/routers/participantsRouter.ts
```diff
@@ -41,7 +41,9 @@ export type AvailableParticipantDTO = Required<Pick<ParticipantDTO, 'name' | 'si
 export type ParticipantRequestDTO = Pick<
   ParticipantDTO,
   'id' | 'name' | 'siteId' | 'types' | 'status'
->;
+> & {
+  requestingUser: Pick<UserDTO, 'email' | 'role'> & { fullName: string };
```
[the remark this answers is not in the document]
lionell-pack-ttd: I thought about it, but it's already on the User entity type - but as a function, not a property (i.e. `fullName()`), so I think we should come up with a pattern around how to use those, and keep this change just on the specific endpoint for now.

src/api/services/participantsService.ts
```diff
@@ -67,10 +73,20 @@ export const getParticipantsAwaitingApproval = async (email: string): Promise<Pa
...
     .where('status', ParticipantStatus.AwaitingApproval);
   return participantsAwaitingApproval;
 };
 
+type SiteIdType = NonNullable<ParticipantDTO['siteId']>;
+export const GetAttachedSiteIDs = async (): Promise<SiteIdType[]> => {
```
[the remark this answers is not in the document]
lionell-pack-ttd: No reason - I'll fix it.

src/web/components/ParticipantRequests/ParticipantApprovalForm.tsx
lionell-pack-ttd: TextInput attaches a controller and tries to make it part of the form, but this is only for display purposes - I didn't want to use `Input` either but it gives me the label and some other things.

src/database/seeds/Participants.ts
lionell-pack-ttd: Good catch - I just removed the awaiting approval seed and changed the participant for the user seed to be awaitingApproval. It was set to awaitingSigning but we're not using that at all at the moment.

src/web/components/ParticipantRequests/ParticipantApprovalForm.tsx
```diff
@@ -13,6 +15,46 @@ import { SearchBarContainer, SearchBarInput, SearchBarResults } from '../Search/
 
 import './ParticipantApprovalForm.scss';
 
+const getSpans = (matches: [number, number][], totalLength: number) => {
```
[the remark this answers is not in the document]
lionell-pack-ttd: Done :)

src/api/services/participantsService.ts
lionell-pack-ttd: Done.

PR #168, opened 2023-09-01
Show added by column and make it sortable

jingyi-gao-ttd:
**This branch is on working with uid2-admin service with latest [cbc-UID2-1177-keyset-share-types](https://github.com/IABTechLab/uid2-admin/pull/130).  You will need to make changes to sites.json and clients.json, here is the example: https://github.com/IABTechLab/uid2-admin/pull/133/files

- Use sharedSiteIds and sharedTypeIds to figure out the addedBy column
- Add sorting to it

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/24f338ae-05c8-4f7f-b47d-558a1d56991b


src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -81,6 +101,7 @@ export function SharingPermissionsTable({
     }
   }, [checkedParticipants.size, isSelectedAll]);
 
+  if (isLoading) return <Loading />;
   const deletePermissionBtn = (
```
alex-yau-ttd, line 105: Should we also include some conditional copy in this `dialog-body-section` to inform the user that if the participant has been shared by type, then it will continue to be shared with despite deletion?

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -3,41 +3,55 @@ import { CheckedState } from '@radix-ui/react-checkbox';
 import { useEffect, useMemo, useState } from 'react';
```
alex-yau-ttd, line 3: I believe the plan is to also rename the 'Added By' column to `Source`, and then the rows will be something like 'DSP', 'Manually Added', 'DSP and Advertiser' and 'DSP and Manually Added'. Is the plan to do this in a future PR?

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -3,41 +3,55 @@ import { CheckedState } from '@radix-ui/react-checkbox';
...
+  sharedSiteIds,
+  sharedTypes,
   onDeleteSharingPermission,
   participantTypes,
 }: SharingPermissionsTableProps) {
+  const { sites: availableParticipants, isLoading } = useAvailableSiteList();
   const [filterText, setFilterText] = useState('');
   const [checkedParticipants, setCheckedParticipants] = useState<Set<number>>(new Set());
   const [openConfirmation, setOpenConfirmation] = useState(false);
+  const siteIds = useMemo(() => new Set(sharedSiteIds), [sharedSiteIds]);
```
alex-yau-ttd, line 33: What's the benefit of using this rather than the `sharedSiteIds` prop directly? I see you utilise some of the Set methods (e.g. `.has`) - does this also provide a performance benefit?

src/web/screens/sharingPermissions.tsx
```diff
@@ -99,11 +99,15 @@ function SharingPermissions() {
...
               </div>
-              <SharingPermissionsTable
-                sharingParticipants={sharingParticipants}
-                onDeleteSharingPermission={handleDeleteSharingPermission}
-                participantTypes={resolvedParticipantTypes}
-              />
+              {/* TODO: not show this table if sharedSiteIds and sharedTypes are empty */}
+              {false && (
+                <SharingPermissionsTable
+                  sharedSiteIds={[]}
```
alex-yau-ttd, line 105: For any other reviewers, these two props will be populated in this PR: https://github.com/IABTechLab/uid2-self-serve-portal/pull/163

src/web/components/SharingPermission/ParticipantsTable.tsx
```diff
@@ -9,20 +9,34 @@ import { ParticipantItem } from './ParticipantItem';
...
-  tableHeader: (filteredParticipants: AvailableParticipantDTO[]) => ReactNode;
+  tableHeader: (filteredParticipants: T[]) => ReactNode;
   className?: string;
   hideSelectAllCheckbox?: boolean;
   showAddedByColumn?: boolean;
-  onFilteredParticipantChanged?: (filteredParticipants: AvailableParticipantDTO[]) => void;
+  onFilteredParticipantChanged?: (filteredParticipants: T[]) => void;
+};
+
+type DetermineParticipantsType<ShowColumn extends boolean> = ShowColumn extends true
```
alex-yau-ttd, line 29: I can see how the name of this type could be confusing, since participants themselves have a `type`, whereas this is referring to a typescript type. Not sure if something like `DetermineParticipantTSType` is any better...

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
```diff
@@ -12,24 +12,40 @@ export default {
 const response: AvailableParticipantDTO[] = [
   {
     name: 'Site 1',
```
alex-yau-ttd, line 14: While we do have a 1:1 mapping between Participant and Sites, I sorta think we should use "Participant 1" as a placeholder since we should be showing the participant name rather than the site name in the UI here.

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: Yes, I remember we have that conversation before, but I couldn't find the copy for that anymore. I will come up with something and let's leave it for Gen to review the copy

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: I could do it in this PR

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: Yes, I found we had a few place that we need to do `siteIds.includes` which is O(n), and `.has` is O(1)

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
jingyi-gao-ttd: I think as we use admin service as source of choice, the name here is actually the site name in admin service

src/web/components/SharingPermission/SharingPermissionsTable.tsx
alex-yau-ttd: good to know! ta

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
alex-yau-ttd: oh perhaps we need to change the source as well then, see this message from Kimberly: https://thetradedesk.slack.com/archives/C04SFSESXQF/p1692803331960089?thread_ts=1692754725.148259&cid=C04SFSESXQF.

We can always do in a separate PR if easier

jingyi-gao-ttd: I see! I think it make sense to create a new PR for that

jingyi-gao-ttd:
https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/0c53d55b-b1ca-44bc-8d7a-3c03d46fd921

- Show source ('Manually Added', 'DSP') rather than Added By ('Manual', 'Auto')
- Show notice when user is trying to delete a site which is shared by Manually added and types

src/web/components/SharingPermission/SharingPermissionsTable.stories.tsx
alex-yau-ttd: I just made a separate ticket: https://atlassian.thetradedesk.com/jira/browse/UID2-1799

src/web/components/SharingPermission/ParticipantItem.tsx
```diff
@@ -51,7 +51,7 @@ export function ParticipantItem({ participant, onClick, checked, addedBy }: Part
           onClick={onClick}
           status={checked}
           className='participant-checkbox'
-          disabled={addedBy === 'Auto'} // addedBy is currently hardcoded to 'Manual'
+          disabled={!!addedBy && !addedBy.includes('Manually Added')} // addedBy is currently hardcoded to 'Manual'
```
alex-yau-ttd, line 54: nit: Let's remove this comment? :D

jingyi-gao-ttd: done!

alex-yau-ttd (approved):

src/web/components/SharingPermission/ParticipantItem.tsx
```diff
@@ -51,7 +51,7 @@ export function ParticipantItem({ participant, onClick, checked, addedBy }: Part
           onClick={onClick}
           status={checked}
           className='participant-checkbox'
-          disabled={addedBy === 'Auto'} // addedBy is currently hardcoded to 'Manual'
+          disabled={!!addedBy && !addedBy.includes('Manually Added')}
```
lionell-pack-ttd, line 54: Doing a string comparison here seems like a code smell - I'm not sure about it. At least there should be a `const` somewhere but it makes me think there might be something else to look into.

src/web/components/SharingPermission/SharingPermissionsTable.tsx
lionell-pack-ttd: `useMemo` is probably unnecessary here - it has an overhead, and the cost of the operation (creating a set from an array) is trivial compared to the cost of rendering the component, so avoiding the set creation every time the component is re-rendered is unnecessary.

You should use `useMemo` if the operation is expensive, or if keeping the same reference will let you avoid re-rendering another expensive component (i.e. you pass the memoized value as a prop to another component which is expensive to re-render).

lionell-pack-ttd: There's a good article about when to *not* use `useMemo` here: https://blog.logrocket.com/when-not-to-use-usememo-react-hook/

src/web/components/SharingPermission/ParticipantsTable.tsx
```diff
@@ -112,7 +127,11 @@ function ParticipantsTableContent({
       <tbody>
         {sortedData.map((participant) => (
           <ParticipantItem
-            addedBy={showAddedByColumn ? 'Manual' : undefined} // TODO: Update this once we have auto add functionality
+            addedBy={
+              showAddedByColumn
+                ? formatStringsWithSeparator((participant as SharingParticipant).addedBy)
```
lionell-pack-ttd, line 129: Nit: I think the formatting should be done in the item instead - that way the checkbox can look for the existence of 'Manually Added' in the array, instead of searching the string for the substring 'Manually Added'. It feels like a cleaner approach.

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: This siteIds is used in the other `useMemo` hook, which is used to populate the addedBy field to the participant and get the sharingParticipants list. This siteIds without the useMemo hook will cause re-render of that hook.

I've noticed we have lots of redundant renders on this page, I would like to create a ticket for it and look into that once we have some spare time in the sprint. Hopefully we could clean up the usage of unnecessary hooks, is that okay?

lionell-pack-ttd: I'm trying to work out why the other `useMemo` exists - what was the actual problem?

src/web/components/SharingPermission/ParticipantsTable.tsx
jingyi-gao-ttd: I think this is a really good idea. I've now moved formatStringsWithSeparator into the participant item, so that I could remove `showAddedBycolumn` from the participant table and make the type check much easier. 
I've also moved the `isAddedByManual` check into a helper to avoid string comparison in the component

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -3,41 +3,59 @@ import { CheckedState } from '@radix-ui/react-checkbox';
...
+
+  const sharingParticipants: SharingParticipant[] = useMemo(() => {
```
lionell-pack-ttd, line 37: Looking at all of the `useMemo` hooks we have in this function, there's a lot of cascading complexity - different `useMemo` results depending on each other. It's a sign to me the whole state needs to be re-thought. This is adding more complexity on top of something that already looks over-optimised.

src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -3,41 +3,59 @@ import { CheckedState } from '@radix-ui/react-checkbox';
...
 
 type SharingPermissionsTableProps = {
-  sharingParticipants: AvailableParticipantDTO[];
+  sharedSiteIds: number[];
+  sharedTypes: string[];
   onDeleteSharingPermission: (siteIds: number[]) => Promise<void>;
   participantTypes: ParticipantTypeDTO[];
 };
 
-function NoParticipant() {
```
alex-yau-ttd, line 21: Was this actually being used before? I know we have conditions to hide the whole collapsible, but I think it would be weird for the whole table to disappear if a user removed all their existing sharing permissions, so I'd argue for keeping this function for that case. Also, this is was Kristen said in Figma:
image

alex-yau-ttd:
The gap between search & add and the sharing permissions table looks a bit big now, I think it was from my PR. 

image

Could you please remove `margin-top: 30px` from .sharing-permissions-table in `SharingPermissionsTable.scss`?

thanks

src/web/components/SharingPermission/SharingPermissionsTable.tsx
jingyi-gao-ttd: I saw we have condition  `sharedSiteIds.length > 0 || sharedTypes.length > 0` on the table, so I think in this case, the table will always have participants showing? If both sharedSiteIds and sharedTypes I think we are showing recommendation again? should we show this section as well?

alex-yau-ttd: I'm actually not sure of the best way to handle this. I think maybe the ideal state is the we hide the table when the user first sees this screen (e.g. before they have gone through the recommendations workflow), but then always show this section after that. That way, if the user removes all existing permissions then the section won't just disappear. 

> if both sharedSiteIds and sharedTypes I think we are showing recommendation again?
Actually I am going to change the condition for recommendation to just be `sharedTypes.length === 0`, if that is any help. Otherwise the checkbox was misleading as it looks like "DSP" is already saved, even though it is just recommended. I think it makes more sense to just be `sharedTypes.length === 0`.

jingyi-gao-ttd: image
I added this back, but the user won't to be able to see this state for now. I think once we have that recommendation flag, we could then hide the table from user who haven't gone through recommendations before.

alex-yau-ttd: yep good idea!

src/web/components/SharingPermission/ParticipantsTable.spec.tsx
jingyi-gao-ttd, line 1: As the filtering and selectAll are now on the parent, we could remove the participantTable spec and story. And we already have story and spec for sharing permission table and search table

lionell-pack-ttd (approved):

PR #171, opened 2023-09-05
ajy-UID2-1735-Create-bulk-add-recommendations-flag

alex-yau-ttd:
- Create migration script to add non-nullable bit column to participants table called "completedRecommendations"
- Add field to seed files
- Update this flag when hitting the button to save types
- Use this new flag instead of `hasSharedSiteIds` and `sharedTypes.length` in the bulk add component and sharingPermissions screen

### Manual tests

**Change from recommendation state and show permissions table when saving by type**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/2f40c45c-9b3b-4bb6-b93f-8111de863de7

**Show permissions table when participant has manually shared with sites, despite not completing recommendations workflow**

image


src/api/routers/participantsRouter.ts
```diff
@@ -277,6 +277,15 @@ export function createParticipantsRouter() {
     }
   );
 
+  participantsRouter.post(
+    '/:participantId/completedRecommendationsWorkflow',
```
alex-yau-ttd, line 281: I thought about just adding this functionality to the `/:participantId/sharingPermission/add` endpoint and checking if `newTypes` was defined, but that would not actually confirm if the user had gone through the recommendations because we may populate that field for them.

src/database/migrations/20230905013335_AddCompletedRecommendationsWorkflowColumnToParticipantTable copy 2.ts
```diff
@@ -0,0 +1,16 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('participants', (table) => {
+    table.boolean('completedRecommendationsWorkflow').defaultTo(false).notNullable();
```
alex-yau-ttd, line 5: Open to suggestions on the column name!

src/api/routers/participantsRouter.ts
```diff
@@ -277,6 +277,15 @@ export function createParticipantsRouter() {
...
+    async (req: ParticipantRequest, res: Response) => {
+      const { participant } = req;
+      await participant!.$query().update({ completedRecommendationsWorkflow: true });
+      return res.status(200).json(participant);
```
alex-yau-ttd, line 285: Wasn't sure what to return here, so just returned the whole participant. Happy to change to something simpler: maybe just `participant.completedRecommendationsWorkflow`?

alex-yau-ttd: Changed to just return `participant.completedRecommendationsWorkflow`

src/web/screens/sharingPermissions.tsx
```diff
@@ -41,6 +42,7 @@ function SharingPermissions() {
         } saved to Your Sharing Permissions`,
       });
       setSharedTypes(response.allowed_types);
+      if (!participant?.completedRecommendations) await CompleteRecommendations(participant!.id);
```
jingyi-gao-ttd, line 45: do we need to reload participant after this?

src/api/entities/Participant.ts
```diff
@@ -69,6 +70,7 @@ export const ParticipantSchema = z.object({
   types: z.array(ParticipantTypeSchema).optional(),
   users: z.array(UserSchema).optional(),
   allowSharing: z.boolean(),
+  completedRecommendations: z.boolean(),
```
jingyi-gao-ttd, line 73: nit: I think this is unnecessary

src/api/routers/participantsRouter.ts
```diff
@@ -277,6 +277,15 @@ export function createParticipantsRouter() {
     }
   );
 
+  participantsRouter.put(
+    '/:participantId/completeRecommendations',
+    async (req: ParticipantRequest, res: Response) => {
+      const { participant } = req;
+      await participant!.$query().update({ completedRecommendations: true });
+      return res.status(200).json(participant?.completedRecommendations);
```
jingyi-gao-ttd, line 285: I think it's a bit weird to return a boolean, maybe better to return the updated participant or just 200?

src/web/screens/sharingPermissions.tsx
alex-yau-ttd: I was initially thinking no, but I think it is probably a good idea. How would I trigger a reload given the participant is using `useContext`?
```
```

src/api/routers/participantsRouter.ts
alex-yau-ttd: fair enough, just returning 200 now

src/api/entities/Participant.ts
alex-yau-ttd: good catch!

src/web/screens/sharingPermissions.tsx
jingyi-gao-ttd: I think we could use loaded the updated participant and use setParticipant to set the latest participant

alex-yau-ttd: ahh right, didn't realise we could just do the following to get the setter:
```
const { participant, setParticipant } = useContext(ParticipantContext);
```

alex-yau-ttd: how's that look now?

alex-yau-ttd: actually maybe it makes more sense for `CompleteRecommendations()` to just return the updated participant, hence I won't need the extra call to `GetCurrentUsersParticipant()`

jingyi-gao-ttd: Yea, I think let CompleteRecommendations() return the updated participant, and then call setParticipant make more sense.

jingyi-gao-ttd: I realised that we have a `useEffect` that has a dependency on participants. However, participant.id is optional to GetSharingList. Could we move this call into the router loader? So that the participant updates will not make this unnecessary call
```
  const loadSharingList = useCallback(async () => {
    const response = await GetSharingList(participant!.id);
    setSharedSiteIds(response.allowed_sites);
    setSharedTypes(response.allowed_types ?? []);
  }, [participant]);

  useEffect(() => {
    loadSharingList();
  }, [loadSharingList]);
```

src/api/routers/participantsRouter.ts
alex-yau-ttd: returning the updated participant now :)

src/api/routers/participantsRouter.ts
alex-yau-ttd: Changed again to return the whole participant :)

jingyi-gao-ttd (approved):

src/web/screens/sharingPermissions.tsx
alex-yau-ttd: **tl;dr**: I think it is better to just remove the optional parameter on the `GetSharingList` call and remove the `participant` dependency on the callback.

I tried putting this into the router loader but here are my findings:
- Since <BulkAddPermissions>, <SearchAndAddParticipants> and <SharingPermissionsTable> all need the resolvedSharingList, I believe they have to be wrapped into one big <Await>, which means resolvedParticipantTypes has to be roped in too. This means that each time the useRevalidator is called, it also needs to run the api calls to participantTypes, available and sharingPermission again. Whereas the original solution does not need to make these three calls, as the updated participant comes back from the add endpoint anyway. If there is a way to avoid having all three components within the same <Await> then useLoaderData might seem like a more viable solution.
- Per the [docs](https://reactrouter.com/en/main/hooks/use-revalidator), useRevalidator should not really be used for normal CRUD operations on data in response to user interactions, instead, should probably use <Form>, useSubmit or useFetcher. I'm not sure if those alternatives are appropriate here, but thought it was an interesting point anyway.
- It could just be the way I have implemented it, but it also seems to cause a flicker in the UI where the components within the <Await> disappear then quickly reappear 
- I'm thinking I'll just merge in the above solution which is to remove the participant as a dependency for the useEffect , and just call GetSharingList() without the optional participantId parameter. I think if all the three components needed those two variables to be resolved every time, then the useLoaderData pattern may be appropriate, but given that is not the case, I think keeping the two local state variables and the useEffect is a better solution.

PR #173, opened 2023-09-05
Use participant name instead of site name

jingyi-gao-ttd:
Use Participant Name in DB instead of site name
image

### DIFF
--- a/src/api/helpers/siteConvertingHelpers.ts
+++ b/src/api/helpers/siteConvertingHelpers.ts
@@ -1,3 +1,4 @@
+import { Participant } from '../entities/Participant';
 import { ParticipantTypeDTO } from '../entities/ParticipantType';
 import { AvailableParticipantDTO } from '../routers/participantsRouter';
 import { ClientType, SiteDTO } from '../services/adminServiceHelpers';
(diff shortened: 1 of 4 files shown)


alex-yau-ttd:
I think this also fixes a bug (which I hadn't noticed before) in which the user would see their own site come up as a sharable participant!

jingyi-gao-ttd:
> I think this also fixes a bug (which I hadn't noticed before) in which the user would see their own site come up as a sharable participant!
> 
> e.g. my participant had the siteId set to the id of `Special`. Before, I would still see `Special` show up, but now it doesn't
This is the code we used to filter out our own site. it is not fixed by this PR, hmmm not sure why that happens before
```
  const getSearchableParticipants = (resolvedParticipants: AvailableParticipantDTO[]) => {
    return resolvedParticipants.filter(
      (p) => p.siteId !== participant?.siteId && !sharingParticipantsSiteIds.has(p.siteId)
    );
  };
```

alex-yau-ttd:
> > I think this also fixes a bug (which I hadn't noticed before) in which the user would see their own site come up as a sharable participant!
> > e.g. my participant had the siteId set to the id of `Special`. Before, I would still see `Special` show up, but now it doesn't
> > This is the code we used to filter out our own site. it is not fixed by this PR, hmmm not sure why that happens before
> 
> ```
>   const getSearchableParticipants = (resolvedParticipants: AvailableParticipantDTO[]) => {
>     return resolvedParticipants.filter(
>       (p) => p.siteId !== participant?.siteId && !sharingParticipantsSiteIds.has(p.siteId)
>     );
>   };
> ```

I believe that code fixes it for the SearchAndAddParticipants component, but the issue was still there for Bulk Add and the table:

image

alex-yau-ttd (approved):

PR #174, opened 2023-09-07
confirmations work

ssundahlTTD:
Add action confirmations via StatusPopup.

invite error

Email contact success

### DIFF
--- a/src/web/components/BusinessContacts/BusinessContact.tsx
+++ b/src/web/components/BusinessContacts/BusinessContact.tsx
@@ -59,11 +59,9 @@ function BusinessContact({
   onRemoveEmailContact,
   onUpdateEmailContact,
 }: BusinessContactProps) {
-  const [hasError, setHasError] = useState<boolean>(false);
-  const [errorMessage, setErrorMessage] = useState<string>('An error has ocurred');
+  const [errorMessage, setErrorMessage] = useState<string>();
 
   const setErrorInfo = (e: Error) => {
-    setHasError(true);
     setErrorMessage(e.message);
   };
 
@@ -90,7 +88,7 @@ function BusinessContact({
       <td>{contact.contactType}</td>
       <td className='action'>
         <div className='action-cell'>
-          {hasError && <InlineMessage message={errorMessage} type='Error' />}
+          {!!errorMessage && <InlineMessage message={errorMessage} type='Error' />}
           <div>
             <BusinessContactDialog
               onFormSubmit={handleUpdateEmailContact}
--- a/src/web/components/BusinessContacts/BusinessContact.tsx
+++ b/src/web/components/BusinessContacts/BusinessContact.tsx
@@ -60,20 +60,26 @@ function BusinessContact({
   onUpdateEmailContact,
 }: BusinessContactProps) {
   const [hasError, setHasError] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<string>('An error has ocurred');
+
+  const setErrorInfo = (e: Error) => {
+    setHasError(true);
+    setErrorMessage(e.message);
+  };
 
   const handleRemoveEmailContact = async () => {
     try {
       await onRemoveEmailContact(contact.id);
-    } catch {
-      setHasError(true);
+    } catch (e) {
+      setErrorInfo(e as Error);
     }
   };
 
   const handleUpdateEmailContact = async (formData: BusinessContactForm) => {
     try {
       await onUpdateEmailContact(contact.id, formData);
-    } catch {
-      setHasError(true);
+    } catch (e) {
+      setErrorInfo(e as Error);
     }
   };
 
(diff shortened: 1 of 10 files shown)


src/web/components/BusinessContacts/BusinessContact.tsx
```diff
@@ -84,7 +90,7 @@ function BusinessContact({
       <td>{contact.contactType}</td>
       <td className='action'>
         <div className='action-cell'>
-          {hasError && <InlineMessage message='An error has occurred' type='Error' />}
+          {hasError && <InlineMessage message={errorMessage} type='Error' />}
```
jingyi-gao-ttd, line 93: I wonder if we still need `hasError` state? could we just use `!!errorMessage`?

src/web/screens/teamMembers.tsx
```diff
@@ -28,21 +30,61 @@ function TeamMembers() {
...
     onTeamMembersUpdated();
   };
 
   const handleRemoveTeamMember = async (userId: number) => {
-    await RemoveUser(userId);
-    onTeamMembersUpdated();
+    try {
+      const response = await RemoveUser(userId);
+      if (response.statusText === 'OK') {
+        handleSuccessPopup('Team member removedfdg dfghdfg hdfgh dfgh drf hdf gh dfg hdf gh.');
```
jingyi-gao-ttd, line 69: I guess this is just placeholder?

src/web/components/TeamMember/TeamMember.tsx
```diff
@@ -118,7 +133,7 @@ function TeamMember({
       <td>{person.role}</td>
       <td className='action'>
         <div className='action-cell'>
```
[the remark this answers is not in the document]
alex-yau-ttd: after: 
image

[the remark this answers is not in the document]
alex-yau-ttd: Or `align-items: center`, whatever you prefer

src/web/screens/teamMembers.tsx
```diff
@@ -28,21 +30,61 @@ function TeamMembers() {
...
+  const [popupDuration, setPopupDuration] = useState<number>(3000);
+
+  const handleErrorPopup = (e: Error) => {
+    const hash = Object.hasOwn(e, 'errorHash') ? `: (${(e as ApiError).errorHash})` : '';
+    setStatusPopup({
+      type: 'Error',
+      message: `${e.message}${hash}`,
+    });
+    setShowStatusPopup(true);
+    setPopupDuration(10000);
```
alex-yau-ttd, line 44: maybe we could change the `<StatusPopup>` component to decide the displayDuration based on the type? If we say Error should be 10000 but everything else is 3000? What do you think?

src/web/screens/teamMembers.tsx
```diff
@@ -28,21 +30,61 @@ function TeamMembers() {
...
+    }
     onTeamMembersUpdated();
   };
 
   const handleRemoveTeamMember = async (userId: number) => {
-    await RemoveUser(userId);
-    onTeamMembersUpdated();
+    try {
+      const response = await RemoveUser(userId);
+      if (response.statusText === 'OK') {
```
alex-yau-ttd, line 68: is there a good reason to use `response.statusText` rather than `response.status`?

ssundahlTTD: I was experimenting with sending bad requests to the API and found that if I send a request to an invalid endpoint, it returned a 200.  So I wanted to be (somewhat) sure that I was getting a successful response.  I agree this seems a little brittle and I don't love it.  I also don't love that all the statusText is 'OK' except addTeamMember which is 'Created.'  Maybe we can change that to also be 'OK'?

src/web/components/TeamMember/TeamMember.tsx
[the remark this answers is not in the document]
ssundahlTTD: Done.  Looks better now.

image

src/web/screens/teamMembers.tsx
ssundahlTTD: Oops.  That's embarrassing.  I was testing long text alignment.  This is fixed.

src/web/components/BusinessContacts/BusinessContact.tsx
ssundahlTTD: Good idea.  I changed it.

src/web/screens/teamMembers.tsx
ssundahlTTD: I like this idea as a default behavior if no duration was passed to the component.  I changed it.

src/web/screens/teamMembers.tsx
lionell-pack-ttd: I'm OK with this as-is for now, but I think we should use status codes properly - @ssundahlTTD can you write up a ticket to fix it so that we can rely on 200s to really mean success? Add details on how to reproduce the 200 on an invalid endpoint call.

lionell-pack-ttd (approved):

PR #177, opened 2023-09-12
Sas UI d2 1793 delete user with audit trail

ssundahlTTD:
We have a foreign key relationship from the auditTrails table to users.  This prevents us from deleting a user that has anything on auditTrails.  So rather than deleting, we will remove the participant relationship and add some randomization to the email address to prevent future duplication.

### DIFF
--- a/src/api/services/usersService.ts
+++ b/src/api/services/usersService.ts
@@ -67,7 +67,9 @@ export const enrichWithUserFromParams = async (
   if (!user) {
     return res.status(404).send([{ message: 'The user cannot be found.' }]);
   }
-  if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, user.participantId))) {
+  if (
+    !(await isUserBelongsToParticipant(req.auth?.payload?.email as string, user.participantId!))
+  ) {
     return res.status(403).send([{ message: 'You do not have permission to that user account.' }]);
   }
 
--- a/src/api/services/usersService.ts
+++ b/src/api/services/usersService.ts
@@ -9,7 +9,7 @@ import { isUserAnApprover } from './approversService';
 export type UserWithIsApprover = User & { isApprover: boolean };
 
 export const findUserByEmail = async (email: string) => {
-  return User.query().findOne('email', email);
+  return User.query().findOne('email', email).where('deleted', 0);
 };
 
 export const enrichUserWithIsApprover = async (user: User) => {
@@ -29,6 +29,7 @@ export const createUserInPortal = async (user: Omit<UserDTO, 'id' | 'acceptedTer
 export const isUserBelongsToParticipant = async (email: string, participantId: number) => {
   const user = await User.query()
     .where('email', email)
+    .where('deleted', 0)
     .andWhere('participantId', participantId)
     .first();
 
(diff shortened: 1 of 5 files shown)


src/api/services/usersService.ts
```diff
@@ -67,7 +67,9 @@ export const enrichWithUserFromParams = async (
   if (!user) {
     return res.status(404).send([{ message: 'The user cannot be found.' }]);
   }
-  if (!(await isUserBelongsToParticipant(req.auth?.payload?.email as string, user.participantId))) {
+  if (
+    !(await isUserBelongsToParticipant(req.auth?.payload?.email as string, user.participantId!))
```
jingyi-gao-ttd, line 71: Just wondering if someone trying to hack and send request to a deleteUserId, will this crash the backend?
For example sending `api/users/:userId/resendInvitation` via postman?

lionell-pack-ttd: It _shouldn't_ be possible to crash the back-end by causing an exception in a route handler - we should have it set up so that all exceptions are intercepted by a top-level handler. If anyone notices a backend crash - please let everyone know and we'll fix it!

src/database/migrations/20230912185913_AddDeletedFlagToUsersTable.ts
```diff
@@ -0,0 +1,16 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('users', (table) => {
+    table.boolean('deleted').defaultTo(false).notNullable();
+  });
+
+  // Update existing rows to have the default value for 'completedRecommendations'
```
alex-yau-ttd, line 8: nit: update comment :)

ssundahlTTD: I fixed the comment and filtered deleted users where necessary.

src/api/services/usersService.ts
```diff
@@ -78,5 +79,5 @@ export const enrichWithUserFromParams = async (
 };
 
 export const getAllUserFromParticipant = async (participant: Participant) => {
-  return participant!.$relatedQuery('users').castTo<User[]>();
+  return participant!.$relatedQuery('users').where('deleted', 0).castTo<User[]>();
```
jingyi-gao-ttd, line 82: I think we remove the association between participant and user when we delete a user, is this still necessary?

src/api/services/usersService.ts
```diff
@@ -29,6 +29,7 @@ export const createUserInPortal = async (user: Omit<UserDTO, 'id' | 'acceptedTer
 export const isUserBelongsToParticipant = async (email: string, participantId: number) => {
   const user = await User.query()
     .where('email', email)
+    .where('deleted', 0)
```
alex-yau-ttd, line 32: nit: this works, but I think using `.andWhere` is a bit clearer here, given it it used on the line below. Otherwise drop the `andWhere` on the below line, I'm pretty sure they behave the same.

alex-yau-ttd (approved):

src/api/services/usersService.ts
ssundahlTTD: It's not necessary.  I removed it.

src/api/services/usersService.ts
ssundahlTTD: changed this to .andWhere()

PR #182, opened 2023-09-13
Update hasSharerRole with new logic - you can share with anyone who i…

lionell-pack-ttd:
…s a DSP, or has ID_READER or SHARER (but we'll exclude the OPTOUT site as that's an internal account that shouldn't be shared with).

### DIFF
--- a/src/api/helpers/siteConvertingHelpers.ts
+++ b/src/api/helpers/siteConvertingHelpers.ts
@@ -26,15 +26,17 @@ export const convertSiteToAvailableParticipantDTO = (
   return {
     name: matchedParticipant ? matchedParticipant.name : site.name,
     siteId: site.id,
-    types:
-      matchedParticipant && matchedParticipant.types
-        ? matchedParticipant.types
-        : mapClientTypeToParticipantType(site.clientTypes ?? [], participantTypes),
+    types: mapClientTypeToParticipantType(site.clientTypes ?? [], participantTypes),
   };
 };
 
-export const hasSharerRole = (site: SiteDTO): boolean => {
-  if (!site.clientTypes || !site.clientTypes.length) return false;
-  if (site.roles.includes('SHARER') || site.clientTypes?.includes('DSP')) return true;
+export const canBeSharedWith = (site: SiteDTO): boolean => {
+  if (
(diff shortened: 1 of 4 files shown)


src/api/helpers/siteConvertingHelpers.ts
```diff
@@ -26,15 +26,17 @@ export const convertSiteToAvailableParticipantDTO = (
...
 };
 
-export const hasSharerRole = (site: SiteDTO): boolean => {
-  if (!site.clientTypes || !site.clientTypes.length) return false;
-  if (site.roles.includes('SHARER') || site.clientTypes?.includes('DSP')) return true;
+export const canBeSharedWith = (site: SiteDTO): boolean => {
+  if (
+    (site.roles.includes('SHARER') ||
+      site.roles.includes('ID_READER') ||
+      site.clientTypes?.includes('DSP')) &&
```
lionell-pack-ttd, line 37: In theory we could skip the DSP check altogether because all DSPs should have ID_READER, but just in case that changes in future this should be the actual logic we need.

alex-yau-ttd (approved):

src/api/helpers/siteConvertingHelpers.ts
```diff
@@ -26,15 +26,17 @@ export const convertSiteToAvailableParticipantDTO = (
   return {
     name: matchedParticipant ? matchedParticipant.name : site.name,
     siteId: site.id,
-    types:
-      matchedParticipant && matchedParticipant.types
```
jingyi-gao-ttd, line 30: I agree to use site.clientTypes as a source of truth. Just wondering currently some of the participants in integ are just using random site ids, Are we going to fix those data? otherwise, their type will not matches the type they have in portal

jingyi-gao-ttd (approved):

src/api/helpers/siteConvertingHelpers.ts
lionell-pack-ttd: If the type in the portal doesn't match admin, then the type in admin should win for deciding who can be shared with. Recommendations it doesn't matter so much, because the participant can still decide what to do.

PR #183, opened 2023-09-13
Fix showDeletionNotice to check current manual added string

jingyi-gao-ttd:
image


alex-yau-ttd (approved):

lionell-pack-ttd (changes requested):
This is just perpetuating the problem.

PR #184, opened 2023-09-13
Add a type for addedBy and fix type issues.

lionell-pack-ttd:


src/web/components/SharingPermission/SharingPermissionsTable.tsx
```diff
@@ -49,7 +49,9 @@ function DeletePermissionDialog({
   };
 
   const showDeletionNotice = (participant: SharingParticipant) => {
-    const remainSources = participant.addedBy.filter((source) => source !== 'Manually Added');
+    const remainSources = (
+      participant.addedBy.filter((source) => source !== 'Manual') as ParticipantTypeDTO[]
```
jingyi-gao-ttd, line 53: nit: is there a reason we still keep the string here rather than constant? is that breaking type check?

lionell-pack-ttd: Because we've got type checking now it doesn't really matter - if you get the string wrong it's a type error. I've updated it anyway 🙂

jingyi-gao-ttd (approved):

src/api/entities/Participant.ts
```diff
@@ -59,7 +59,7 @@ export class Participant extends BaseModel {
 
 // TODO: Can ModelObjectOpt do relationships automatically?
 export type ParticipantDTO = Omit<ModelObjectOpt<Participant>, 'types' | 'users'> & {
-  types?: ModelObjectOpt<ParticipantType>[];
+  types?: ParticipantTypeDTO[];
```
alex-yau-ttd, line 62: what's the rationale behind this change?

lionell-pack-ttd: Just because we're already doing the ModelObjectOpt<> to create ParticipantTypeDTO anyway - I think the types would have just "matched", but I'm used to C# where types which are just the same still won't match unless they're the exact same type.

PR #185, opened 2023-09-13
Sas UI d2 1862 api status code bug2

ssundahlTTD:
Correctly return 404 on incorrect routes and use status rather than statusText.

### DIFF
--- a/src/web/screens/emailContacts.tsx
+++ b/src/web/screens/emailContacts.tsx
@@ -75,7 +75,7 @@ export function BusinessContacts() {
   const handleAddEmailContact = async (formData: BusinessContactForm) => {
     try {
       const response = await AddEmailContact(formData);
-      if (response.status === 200) {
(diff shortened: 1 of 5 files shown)


src/web/screens/emailContacts.tsx
```diff
@@ -75,7 +75,7 @@ export function BusinessContacts() {
   const handleAddEmailContact = async (formData: BusinessContactForm) => {
     try {
       const response = await AddEmailContact(formData);
-      if (response.statusText === 'OK') {
+      if (response.status === 200) {
```
jingyi-gao-ttd, line 78: nit: I wonder if we should change the route to return 201

jingyi-gao-ttd (approved):

src/web/screens/emailContacts.tsx
ssundahlTTD: i now return and look for a 201

PR #186, opened 2023-09-14
jyg-UID2-1791 Build self-serve portal, db migration and keycloak in one step

jingyi-gao-ttd:
- Checks email templates folder and sync to SendGrid before build portal
- Remove redundant jobs, and make workflow re-usable
- Build SSportal, Keycloak and DB migration with the same image tage


.github/workflows/ssp-ci.yml
```diff
@@ -0,0 +1,32 @@
+name: Self-serve Portal CI
```
alex-yau-ttd, line 1: nit: could we possibly call this file `ssportal-ci.yml`? my head went to Supply-Side-Platform for a sec 😆

.github/workflows/ssp-ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: Sync handlebars template with SendGrid
+    if: needs.check-email-templates.outputs.emailTemplatesUpdated == 'true'
+    uses: ./.github/workflows/sync-sendgrid-template-action.yml
+    with:
+      FORCE_SYNC_ALL: 'true'
+    secrets:
+      SENDGRID_API_KEY: ${{ secrets.SENDGRID_API_KEY }}
+  build-and-publish-ssportal:
+    needs: [sync-email-templates]
+    if: always() && (needs.sync-email-templates.result == 'success' || needs.sync-email-templates.result == 'skipped') # Run build and publish ssportal when sync email template is ran or skipped
```
alex-yau-ttd, line 31: what's the point of this conditional if there is an `always()`? won't it just run every time anyway?

jingyi-gao-ttd: Coz I use needs here to make sure `build-and-publish-ssportal` runs after `sync-email-templates`, without always(), `build and publish ssportal` will be skipped if sync is skipped

.github/workflows/ssp-ci.yml
jingyi-gao-ttd: Yeah, makes sense, haha I will update the file name

.github/workflows/ssp-ci.yml
alex-yau-ttd: okay - so does that mean that `needs.sync-email-templates.result == 'skipped'` does not cover the case if the sync is skipped? I still can't see the point of `(needs.sync-email-templates.result == 'success' || needs.sync-email-templates.result == 'skipped')` if we have an `always` - I must be missing something

jingyi-gao-ttd: so I use `always()` to make sure `build-and-publish-ssportal`should always run after there is result of sync-email-templates. And Use `needs.sync-email-templates.result == 'success' || needs.sync-email-templates.result == 'skipped'` to make sure `build-and-publish-ssportal` should not run if sync-email-templates is failed

jingyi-gao-ttd: Without always(), both skipped and failed will skip the `build-and-publish-ssportal` and without the result check, build and publish ssportal will still run after sync-email-templates failed

alex-yau-ttd: ahh in my head I was reading it as `always() ||` not `always() &&` for some reason 🤦  thanks for explaining!

.github/workflows/docker-publish-action.yml
```diff
@@ -1,19 +1,22 @@
-name: Publish API/UI Docker Image
-on: workflow_dispatch
+on:
+  workflow_call:
+    inputs:
+      IMAGE_NAME:
+        required: true
+        type: string
+      DOCKERFILE:
+        required: false
```
alex-yau-ttd, line 8: naïve question: why is `DOCKERFILE` not required?

.github/workflows/docker-publish-action.yml
```diff
@@ -1,19 +1,22 @@
...
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME_UID2: iabtechlab/uid2-ssportal
 
 jobs:
   build-publish-docker:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
```
alex-yau-ttd, line 13: another naïve question: why don't we need this anymore? Are we just using defaults instead and prefer that?

.github/workflows/docker-publish-action.yml
```diff
@@ -1,19 +1,22 @@
...
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
```
alex-yau-ttd, line 19: what does the version here actually mean? I see v2, v3 and v4 used throughout our workflows for ` actions/checkout`

.github/workflows/docker-publish-action.yml
alex-yau-ttd: or is it because the permissions are specified now in `docker-publish-ssportal.yml`?

.github/workflows/docker-publish-action.yml
jingyi-gao-ttd: Good catch! It should be required

.github/workflows/docker-publish-action.yml
jingyi-gao-ttd: Yes, I put it in there

.github/workflows/docker-publish-action.yml
jingyi-gao-ttd: https://github.com/actions/checkout I just make sure it is uisng the latest version

.github/workflows/sync-sendgrid-template.yml
```diff
@@ -0,0 +1,14 @@
+on:
+  workflow_dispatch:
+    inputs:
+      forceSyncAll:
```
alex-yau-ttd, line 4: Is there a good reason for having this input use camelCase while the other inputs (in different files) use `FORCE_SYNC_ALL`. Is it purely to separate between `FORCE_SYNC_ALL` and `forceSyncAll`?

.github/workflows/docker-publish-action.yml
alex-yau-ttd: is it worth upgrading all of our workflows to use v4 then?

.github/workflows/sync-sendgrid-template.yml
jingyi-gao-ttd: No need to separate them. I've updated them to be consistent

.github/workflows/docker-publish-action.yml
jingyi-gao-ttd: I've upgraded all of them

.github/workflows/docker-publish-ssportal.yml
```diff
@@ -0,0 +1,22 @@
+on:
+  workflow_call:
+
+jobs:
+  build-ssportal:
+    name: Publish API/UI Docker Image
+    uses: ./.github/workflows/docker-publish-action.yml
+    with:
+      IMAGE_NAME: 'iabtechlab/uid2-ssportal' # env context is not available to job.if
```
alex-yau-ttd, line 9: what's this comment mean?

jingyi-gao-ttd: Removed

alex-yau-ttd (approved):

PR #190, opened 2023-09-18
Remove template check from CI

jingyi-gao-ttd:


.github/workflows/docker-publish-ssportal.yml
```diff
@@ -1,22 +0,0 @@
-on:
```
jingyi-gao-ttd, line 1: This reusable workflow is redundant, moved everything to ssportal-ci.yml

alex-yau-ttd (approved):

PR #194, opened 2023-09-19
jyg-UID2-1734 cstg domain mangement table

jingyi-gao-ttd:
- List domains
image

- Bulk delete domains 
image
image

- Add new domain

image

- No domains
image


src/web/components/ClientSideTokenGeneration/CstgDomainsTable.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import { Meta, Story } from '@storybook/react';
+
+import { CstgDomainsTable, CstgDomainsTableProps } from './CstgDomainsTable';
+
+export default {
+  title: 'CSTG/CstgDomainsTable',
+  component: CstgDomainsTable,
+} as Meta<typeof CstgDomainsTable>;
+
+const Template: Story<CstgDomainsTableProps> = (args) => <CstgDomainsTable {...args} />;
```
jingyi-gao-ttd, line 10: Using `Story<Props>` instead of `ComponentStory<typeof Component>` Give us type check for args:

Since v7 ComponentStory and ComponentMeta are deprecated, maybe we should upgrade to v7
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#componentstory-componentstoryobj-componentstoryfn-and-componentmeta-types-are-deprecated

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -0,0 +1,59 @@
...
+        width: 21px;
+        height: 21px;
+      }
+      .input-container {
+        height: 26px;
+      }
+    }
+
+    .domain {
+      min-width: 220px;
```
lionell-pack-ttd, line 28: I'm always cautious of using larger pixel values like this (and also the `th.action { padding-left: 140px; }` above). Could you achieve the effect you're after by using flex-grow?

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -0,0 +1,59 @@
...
+  .cstg-domains-management-actions {
+    display: flex;
+    align-items: center;
+    gap: 16px;
+    padding-left: 11px;
+    margin-bottom: 10px;
+  }
+
+  .cstg-domains-management-delete-button {
+    color: var(--theme-radio-background);
```
lionell-pack-ttd, line 52: radio-background doesn't sound right for a delete button - is there a more appropriate theme value for this?

jingyi-gao-ttd: I've add new theme value for it

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
jingyi-gao-ttd: I think the table does not support flex-grow. We used to use percentages to control the width of the column. 

In this case, what I want to achieve is to make sure the domain column by default is wider than the text input. Maybe give the action column  a percentage is a better approach to set min-width?

jingyi-gao-ttd: I've updated it to percentage width and remove the padding-left as I found in our case we don't need that padding

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+            <th className='action'>Actions</th>
+          </tr>
+        </thead>
+        <tbody>
+          {domains.map((domain) => (
+            <CstgDomainItem
+              key={domain}
+              domain={domain}
+              onClick={() => handleSelectDomain(domain)}
+              onDelete={() => handleBulkDeleteDomains([domain])}
```
alex-yau-ttd, line 206: I'm thinking this should probably open the delete dialog (like the top level delete button), rather than just deleting straight away

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -0,0 +1,47 @@
+.cstg-domains-management {
```
alex-yau-ttd, line 1: I'm wondering if we could potentially get rid of the black border when certain components are in focus. In this video I am using `TAB` and `SHIFT+TAB` to hop around
https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/053afe50-2f9e-4548-a444-8671c1bd33b6

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <input
+          className='input-container'
+          value={newDomain}
+          onChange={(e) => setNewDomain(e.target.value)}
+        />
+      </td>
+      <td className='action'>
+        <div className='action-cell'>
+          <button type='button' className='transparent-button' onClick={() => onAdd(newDomain)}>
+            Save
```
alex-yau-ttd, line 40: Are you planning for hitting "Save" to be the only way to save a new row? I am thinking even just hitting enter or `onBlur` could possibly make the UX more smooth?

jingyi-gao-ttd: In the next PR, I are going to introduce the validation and show a warning like "It does not look like, do you mean '....' " which the save button will be disabled and the user will need to click the correct value, I feel like it is better to let them click save button afterwards? Cause they can see the button is disabled and then turns into able after they correct the value. If using hitting enter, we maybe need to do more things to emphasis this action didn't go through?

alex-yau-ttd: ahh good point, I forgot about validation. Sounds reasonable to leave as is for now then.

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
jingyi-gao-ttd: Good catch, I've removed the outline now

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -0,0 +1,47 @@
...
+    th.action {
+      text-align: end;
+    }
+
+    tr {
+      .tristate-checkbox {
+        margin-right: 16px;
+      }
+
+      button[role='checkbox'] {
```
alex-yau-ttd, line 20: I've seen this styling before. Is it worth looking into extracting this out to somewhere more global?

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
jingyi-gao-ttd: Yea you're right, confirmation dialog added

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const WithDomains = Template.bind({});
+
+WithDomains.args = {
+  domains: ['test.com', 'abc.com', '123.com'],
+  onUpdateDomains: (domains) => Promise.resolve(console.log('update domains to:', domains)),
```
alex-yau-ttd: note that this is a bit confusing when deleting 'test.com', since it will say `update domains to: test.com`. Though I'm not sure of the best way to address that given we don't have separate separate add and delete methods.

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
jingyi-gao-ttd: make sense to be global, updated

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <thead>
+          <tr>
+            <th> </th>
+            <th className='domain'>Domain</th>
+            <th className='action'>Actions</th>
+          </tr>
+        </thead>
+        <tbody>
+          {domains.map((domain) => (
+            <CstgDomainItem
```
alex-yau-ttd, line 178: Do we plan to have existing domain items as un-editable? So only add and delete?

jingyi-gao-ttd: We don't really have an ID on domains, so I am sure if we need un-editable or not.

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.stories.tsx
jingyi-gao-ttd: Ah.... I think I've tested this one... updated

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
alex-yau-ttd: yeah I guess it's just a string. seems reasonable

src/web/styles/tables.scss
```diff
@@ -36,4 +36,50 @@ table {
...
+  gap: 20px;
+
+  .table-data-text {
+    margin-left: 33px;
+
+    h2 {
+      margin-top: 0;
+    }
+  }
+}
```
alex-yau-ttd, line 72: should this css block definitely be in `tables.scss`? Should SharingPermissionsTable, BusinessContactsTable and ParticipantRequestsTable use this same className?

alex-yau-ttd (approved):

PR #198, opened 2023-09-21
jyg-UID2-1734 domain normalization

jingyi-gao-ttd:
### Changes
- Add cancel button to remove the input row
- Add error message and warning message 
- Support normalizing top-level domain
- Support country code top-level domains
- Support suffix check

### Unit Tests
image

### Manual Tests
**Remove row when click cancel**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/470737ae-ac95-4fc1-9478-66bd29d0d203

**Show recommended domain when given url**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/48e8f38c-7998-4693-be9b-d90c3146e251

**Show recommended domain when given domain includes country code**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/1e7b6f08-eda5-411a-86dc-2be9dd46cf7e

**Show error when given input with invalid suffix**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/d56a94b6-355e-47a8-acb2-8d56da5f2751

**Show recommended domain when given domain includes country code**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/bb56c8d6-552e-47cd-aaa8-1b996fb74d3c


src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+import './CstgDomainInputRow.scss';
+
+type CstgDomainInputRowProps = {
+  onAdd: (newDomain: string) => Promise<void>;
+  onCancel: () => void;
+};
+
+type DomainProps = {
+  isIcann: boolean | null;
```
alex-yau-ttd, line 17: what's Icann?

alex-yau-ttd: quick google says it stands for Internet Corporation for Assigned Names and Numbers but what's that mean here?

jingyi-gao-ttd: So the lib has a public suffix list built-in.  isICANN means the TLD come from ICANN part of the list

The full list is here: https://publicsuffix.org/list/public_suffix_list.dat

alex-yau-ttd: thanks! makes sense

src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+            </div>
+          ),
+        });
+    } else {
+      setValidationResult({
+        type: 'Error',
+        message: <span data-testid='domain-input-error-message'>Invalid domain</span>,
+      });
+    }
+  }, 500);
```
alex-yau-ttd, line 80: maybe this could be put in a variable somewhere to avoid magic number?

src/web/components/Input/Input.scss
```diff
@@ -3,25 +3,24 @@
...
-}
 
-.input-container {
-  height: 45px;
-  box-sizing: border-box;
-  border: 1px solid var(--theme-input-border) !important;
-  border-radius: 5px;
-  padding: 0.5rem;
-}
+  .input-container {
```
alex-yau-ttd, line 7: nit: it feels a bit unintuitive to me that an `input-container` is inside an `input-field`. But at the same time, I'm not sure what the best way to address that would be. e.g. I'm not sure if just flipping the names (i.e. a container has a field instead) would solve the problem. 

What do you think? Happy to leave as-is if no easy fix.

src/web/styles/themes.scss
```diff
@@ -30,8 +30,10 @@
   --theme-warning-background: #fefcd3;
   --theme-info-icon: #7085d4;
   --theme-error-icon: #e52900;
+  --theme-error-border: #e52900;
```
alex-yau-ttd, line 33: I do wonder if this colour would ever realistically differ from `--theme-error-icon`? 
Maybe we could simplify and just have `theme-error: #e52900;` (i.e. change from `red`) and `--theme-error-background: #facec5;` and remove the specific `icon` and `border` ones.

What do you think?

src/web/styles/themes.scss
```diff
@@ -30,8 +30,10 @@
...
   --theme-success-icon: #035959;
   --theme-warning-icon: #ffdb1d;
+  --theme-warning-border: #ffdb1d;
```
alex-yau-ttd, line 36: as above, maybe we don't need this and can just use a more generic `--theme-warning`

src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  return (
+    <tr>
+      <td>
+        <button
+          className='icon-button domain-input-remove-button'
+          type='button'
+          onClick={handleCancel}
+        >
+          <FontAwesomeIcon icon='minus' />
```
alex-yau-ttd, line 97: I found this minus icon wasn't very intuitive, what do you think about using `xmark` instead?

src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.spec.tsx
```diff
@@ -0,0 +1,60 @@
+import { composeStories } from '@storybook/testing-react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import * as stories from './CstgDomainInputRow.stories';
+
+const { Default } = composeStories(stories);
+
+const validationInlineBanner = async (messageTestId: string) => {
+  expect(await screen.findByTestId(messageTestId)).toBeInTheDocument();
```
alex-yau-ttd, line 10: I'm a bit concerned that we have to rely on test ids so much for our testing, given it is the last resort for [testing priorities](https://testing-library.com/docs/queries/about/#priority). 

Is there a way we can alter the component to reduce our reliance on test ids? Happy to discuss, it's not exactly my strong point 😊

jingyi-gao-ttd: The main reason I rely on test id rather than text is that I feel we might get a couple of rounds of text changes after this feature is realised, and it easily breaks the tests.

src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.tsx
jingyi-gao-ttd: make sense to me, changed it to xmark instead

src/web/styles/themes.scss
jingyi-gao-ttd: changed

src/web/styles/themes.scss
jingyi-gao-ttd: make sense, changed

src/web/components/Input/Input.scss
jingyi-gao-ttd: Yes, I agree. But I find we have a few pages that are using input-container, flipping names needs to be careful. I will think if I can come up something with less file changes

jingyi-gao-ttd: I change the `input-field` to `input-wrapper` I hope this make sense?

alex-yau-ttd (approved):

src/web/components/ClientSideTokenGeneration/CstgDomainInputRow.spec.tsx
alex-yau-ttd: I see, I guess we don't export strings currently so it would mean double the changes when updating strings.

src/web/components/Input/Input.scss
alex-yau-ttd: makes sense to me!

PR #200, opened 2023-09-25
jyg-UID2-1736 Pretends sharing with DPS when allow_sites is null

jingyi-gao-ttd:
### Changes

- Decoupling updating sharing type from adding sharing permission endpoint 
- Add new audit trail event for sharing types
- Set sharedTypes to ['DSP'] to SharingPermissionsTable when allow_sites is null

### Manual Tests

**Show share with DSP when the participant has completedRecommendations set to true and allow_sites is null**

Screenshot 2023-09-25 at 5 23 48 pm
Screenshot 2023-09-25 at 5 24 01 pm

**Show Recommendation and hide sharingPermissionTable when the participant has completedRecommendations set to false  and allow_sites is null**

image

**When save bulk add permissions should set allow_sites to []**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/7a369a38-364d-4f06-874f-6b12ac07266f

image


src/api/services/participantsService.ts
```diff
@@ -75,30 +75,48 @@ export const getParticipantsBySiteIds = async (siteIds: number[]) => {
...
 };
 
 export const deleteSharingParticipants = async (
   participantSiteId: number,
-  siteIds: number[],
-  types: ClientType[]
+  siteIds: number[]
 ): Promise<SharingListResponse> => {
   const sharingListResponse = await getSharingList(participantSiteId);
+  if (!sharingListResponse.allowed_sites) throw Error('The site is not shared with any sites');
```
alex-yau-ttd, line 97: I'm wondering if it would be clearer if the error message was something like:
"There are no sharing sites to delete"

jingyi-gao-ttd: yep I agree, I think this error message is more clear to me

src/web/screens/sharingPermissions.tsx
```diff
@@ -105,6 +103,19 @@ function SharingPermissions() {
     loadSharingList();
   }, [loadSharingList]);
 
+  const renderSharingPermissionTable = () => {
+    if (!participant?.completedRecommendations) return;
+    const tableSharedSiteIds = sharedSiteIds ?? [];
+    const tableSharedTypes = sharedSiteIds === null ? ['DSP' as ClientType] : sharedTypes;
```
lionell-pack-ttd, line 109: This breaks type safety - you can put anything in that string. 
image
Please don't use `as` unless you really have to.

jingyi-gao-ttd: Thanks! I removed the type assertion and add type annotation to tableSharedTypes

PR #201, opened 2023-09-25
Sas UI d2 1844 new cstg key pair

ssundahlTTD:
Allow the user to create a new CSTG key pair.  Add the creation dialog to the key pair screen.

image

image


src/api/routers/participantsRouter.ts
```diff
@@ -243,6 +249,44 @@ export function createParticipantsRouter() {
...
+      );
+
+      const keyPairs = await addKeyPair(participant.siteId, name, disabled);
+
+      await updateAuditTrailToProceed(auditTrail.id);
+      return res.status(201).json(keyPairs);
+    }
+  );
+
+  participantsRouter.get(
```
jingyi-gao-ttd, line 277: nit: May I ask for the reason for moving this from site routes? For me, I feel as the participant route keeps growing, it makes sense to have those site-oriented APIs to be part of sites routes? (Key pair management, sharing permission management and domain management)

src/web/components/KeyPairs/KeyPairDialog.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { ComponentMeta, ComponentStory } from '@storybook/react';
```
jingyi-gao-ttd, line 1: nit: for the new story could we use CSF 3.0

here is the doc: https://storybook.js.org/docs/react/writing-stories/introduction

src/web/components/KeyPairs/KeyPairDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <Form<AddKeyPairFormProps>
+        onSubmit={onSubmit}
+        submitButtonText='Create Key Pair'
+        defaultValues={keyPair}
+      >
+        <TextInput inputName='name' label='Name' />
+        <CheckboxInput
+          inputName='disabled'
+          label='Disabled'
```
jingyi-gao-ttd, line 44: I think it looks a bit off to have Disabled and checkbox in two lines. could we try
```
<CheckboxInput
          inputName='disabled'
          options={[{ optionLabel: 'Disabled', value: false }]}
        />
```

src/api/routers/participantsRouter.ts
ssundahlTTD: The main reason is that I wanted this behind participant validation and didn't want to recreate that in another router.  The way I wrote this, the participant will only see their own sites (once that's done on the back end).  As it was implemented in the site router, anyone could craft a request to see another site's keys without any sort of permission checks.

src/web/components/KeyPairs/KeyPairDialog.tsx
ssundahlTTD: Good idea.  I changed it.

src/web/components/KeyPairs/KeyPairDialog.stories.tsx
ssundahlTTD: I changed it.

PR #207, opened 2023-10-06
Show participant is sharing with DSP when allow_sites is null

jingyi-gao-ttd:
### Description:
- Show participant is sharing with DSP when allow_sites is null
- Throw error when sharing participant `/api/sharing/list/` return 404  (**I believe we should hide sharing permission from menu for this case, but I think there is a separate ticket for this already**.)

As discussed, when a participant is GENERATOR or SHARER or has a client_keypair, admin service creates a keyset for it. For now we don't allow user create keyset from portal

### Required PR:

- Admin service: [jyg-UID2-1736 return null instead of [] when allow_sites is empty](https://github.com/IABTechLab/uid2-admin/pull/164)

### Manual tests
1. Approve participant with a siteId that has SHAERE role, it should get empty allow_sites, and allow_types, it should show recommended bulk permission based on it's clientTypes
Screenshot 2023-10-06 at 3 56 56 pm

2. Approve participant with a siteId with _GENERATOR_ role, it should get null allow_sites and empty allow_types. 
- In the portal, it should show recommended bulk permission based on it's clientTypes
Screenshot 2023-10-06 at 3 59 58 pm

- It should has `allow_sites` set to `[]` **after click save in bulk permission**
Screenshot 2023-10-06 at 4 00 38 pm

3. Approve participant with a siteId with **GENERATOR** role, and set`completedRecommendations` to true from DB, it should show shared with DSP
Screenshot 2023-10-06 at 4 05 25 pm

4. Approve participant with a siteId that has **MAPPER** role, it should error dialog when user goes to the sharing permission page

image

### DIFF
--- a/src/web/screens/sharingPermissions.tsx
+++ b/src/web/screens/sharingPermissions.tsx
@@ -1,9 +1,7 @@
-import { useContext, useEffect, useState } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { ReactNode, useContext, useEffect, useState } from 'react';
 
 import { ClientType } from '../../api/services/adminServiceHelpers';
 import { Collapsible } from '../components/Core/Collapsible';
-import { Dialog } from '../components/Core/Dialog';
 import { StatusNotificationType, StatusPopup } from '../components/Core/StatusPopup';
 import { BulkAddPermissions } from '../components/SharingPermission/BulkAddPermissions';
 import { SearchAndAddParticipants } from '../components/SharingPermission/SearchAndAddParticipants';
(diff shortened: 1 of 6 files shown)


src/web/screens/sharingPermissions.tsx
```diff
@@ -91,15 +94,18 @@ function SharingPermissions() {
...
+        const response = await GetSharingList();
+        setSharedSiteIds(response.allowed_sites);
+        setSharedTypes(response.allowed_types ?? []);
+      } catch (e: unknown) {
+        if (e instanceof ApiError) throwError(e);
+      }
+    };
     loadSharingList();
-  }, [loadSharingList]);
+  }, [throwError]);
```
lionell-pack-ttd, line 108: I think we should provide a better error experience here - if there's a 404 from the admin service, we know exactly what's wrong. Just dumping the user to an error where they can't even see the menu any more isn't the right move.

src/web/screens/sharingPermissions.tsx
```diff
@@ -91,15 +125,24 @@ function SharingPermissions() {
...
+        const response = await GetSharingList();
+        setSharedSiteIds(response.allowed_sites);
+        setSharedTypes(response.allowed_types ?? []);
+      } catch (e: unknown) {
+        if (e instanceof ApiError) {
+          if (e.statusCode === 404) {
+            setShowErrorDialog(true);
+            return;
+          }
+          throwError(e);
```
jingyi-gao-ttd, line 140: if there is any other ApiError for this get request, I think we should still show error page

lionell-pack-ttd (approved):

lionell-pack-ttd:
I'm merging this as it's approved now, but Jingyi is unavailable at the moment.

PR #209, opened 2023-10-06
show automatically granted sharing permissions on dashboard

ssundahlTTD:
Show both manually added and automatically granted sharing permissions on the dashboard,

image

image

### DIFF
--- a/src/web/screens/home.tsx
+++ b/src/web/screens/home.tsx
@@ -7,7 +7,7 @@ import DocumentationCard from '../components/Home/DocumentationCard';
 import SharingPermissionCard from '../components/Home/SharingPermissionCard';
 import { CurrentUserContext } from '../contexts/CurrentUserProvider';
 import { GetSharingList } from '../services/participant';
-import { preloadAvailableSiteList } from '../services/site';
+import { preloadAllSitesList } from '../services/site';
 import { PortalRoute } from './routeUtils';
 
 import './home.scss';
@@ -26,21 +26,14 @@ function Home() {
         const sharingList = await GetSharingList();
         const manualSites = sharingList.allowed_sites;
         const allowedTypes = sharingList.allowed_types;
-        const allowedTypeMap = new Map<ClientType, boolean>();
+        const allowedTypeSet = new Set<ClientType>();
         allowedTypes.forEach((item) => {
-          allowedTypeMap.set(item, true);
+          allowedTypeSet.add(item);
         });
-        const siteList = await preloadAvailableSiteList();
+        const siteList = await preloadAllSitesList();
         const bulkSites = siteList.filter((item) => {
-          let found = false;
           const clientTypes = item.clientTypes || [];
-          for (let i = 0; i < clientTypes.length; i += 1) {
-            if (allowedTypeMap.get(clientTypes[i])) {
-              found = true;
-              break;
-            }
-          }
-          return found;
+          return clientTypes.some((clientType) => allowedTypeSet.has(clientType));
         });
         setSharingPermissionsCount(manualSites.length);
         setBulkPermissionsCount(bulkSites.length);
--- a/src/web/screens/home.tsx
+++ b/src/web/screens/home.tsx
@@ -26,10 +26,7 @@ function Home() {
         const sharingList = await GetSharingList();
         const manualSites = sharingList.allowed_sites;
         const allowedTypes = sharingList.allowed_types;
-        const allowedTypeSet = new Set<ClientType>();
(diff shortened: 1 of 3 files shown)


src/web/screens/home.tsx
```diff
@@ -14,15 +16,34 @@ function Home() {
...
-        // TODO include the sites from allowed_types in the count
-        setSharingPermissionsCount(sharingList.allowed_sites.length);
+        const manualSites = sharingList.allowed_sites;
+        const allowedTypes = sharingList.allowed_types;
+        const allowedTypeMap = new Map<ClientType, boolean>();
+        allowedTypes.forEach((item) => {
+          allowedTypeMap.set(item, true);
+        });
+        const siteList = await preloadAvailableSiteList();
+        const bulkSites = siteList.filter((item) => {
```
alex-yau-ttd, line 34: nit: thoughts on extracting this into a separate method?

alex-yau-ttd (approved):

src/web/screens/sharingPermissions.tsx
```diff
@@ -152,9 +151,4 @@ export const SharingPermissionsRoute: PortalRoute = {
   description: 'Sharing Permissions',
   element: <SharingPermissions />,
   path: '/dashboard/sharing',
-  loader: () => {
-    preloadAvailableSiteList();
```
alex-yau-ttd, line 156: Why exactly are we removing these?

alex-yau-ttd: Is it because we are calling it on `home.tsx` so it is just unnecessary now?

src/web/screens/home.tsx
```diff
@@ -14,15 +16,34 @@ function Home() {
...
   useEffect(() => {
     const getSharingParticipantsCount = async () => {
       setIsLoading(true);
       try {
         const sharingList = await GetSharingList();
-        // TODO include the sites from allowed_types in the count
-        setSharingPermissionsCount(sharingList.allowed_sites.length);
+        const manualSites = sharingList.allowed_sites;
+        const allowedTypes = sharingList.allowed_types;
+        const allowedTypeMap = new Map<ClientType, boolean>();
```
jingyi-gao-ttd, line 29: nit: is there a reason to use Map? If we want to O(1) I think we could Set, which Set.has() is O(1) as well

src/web/screens/home.tsx
```diff
@@ -14,15 +16,34 @@ function Home() {
...
+        allowedTypes.forEach((item) => {
+          allowedTypeMap.set(item, true);
+        });
+        const siteList = await preloadAvailableSiteList();
+        const bulkSites = siteList.filter((item) => {
+          let found = false;
+          const clientTypes = item.clientTypes || [];
+          for (let i = 0; i < clientTypes.length; i += 1) {
```
jingyi-gao-ttd, line 37: nit: how about use `clientTypes.some(clientType => allowedTypeMap.get(clientType))` instead of for loop?

src/web/screens/sharingPermissions.tsx
ssundahlTTD: They no longer seemed to be in use.  We were just sending a request for the data and never doing anything with it.

src/web/screens/home.tsx
ssundahlTTD: Good idea.  I changed it to use a Set.

src/web/screens/home.tsx
ssundahlTTD: I changed it.

src/web/screens/home.tsx
ssundahlTTD: I simplified this a bit after a suggestion from Jingyi so I don't think it's really necessary.

src/web/screens/home.tsx
```diff
@@ -26,21 +26,14 @@ function Home() {
         const sharingList = await GetSharingList();
         const manualSites = sharingList.allowed_sites;
         const allowedTypes = sharingList.allowed_types;
-        const allowedTypeMap = new Map<ClientType, boolean>();
+        const allowedTypeSet = new Set<ClientType>();
```
jingyi-gao-ttd, line 29: nit: could just be `new Set<ClientType>(allowedTypes);`

jingyi-gao-ttd (approved):

src/web/screens/home.tsx
ssundahlTTD: Good idea.  I changed it.

PR #210, opened 2023-10-08
Warn publishers against removing sharing with DSPs

alex-yau-ttd:
This includes a minimum-effort change to the usersRouter to populate the current participant with the source-of-truth types, which is used in the participant context in the bulk add control.

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/ab61725a-bca3-4505-b336-a38a65645583

### DIFF
--- a/src/api/routers/usersRouter.ts
+++ b/src/api/routers/usersRouter.ts
@@ -3,9 +3,12 @@ import { v4 as uuid } from 'uuid';
 import { z } from 'zod';
 
 import { Participant, ParticipantStatus } from '../entities/Participant';
+import { ParticipantType } from '../entities/ParticipantType';
 import { UserRole } from '../entities/User';
 import { getLoggers } from '../helpers/loggingHelpers';
+import { mapClientTypeToParticipantType } from '../helpers/siteConvertingHelpers';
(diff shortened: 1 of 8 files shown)


src/api/routers/usersRouter.ts
```diff
@@ -54,8 +57,15 @@ export function createUsersRouter() {
   });
 
   usersRouter.get('/current/participant', async (req: UserRequest, res) => {
-    const participant = await req.user!.$relatedQuery('participant').withGraphFetched('types');
-    return res.status(200).json(participant);
+    const currentParticipant = await Participant.query().findOne({ id: req.user!.participantId });
+    const sites = await getSiteList();
+    const currentSite = sites.find((x) => x.id === currentParticipant?.siteId);
```
jingyi-gao-ttd, line 62: nit: maybe we need a get site by id endpoint in uid2-admin later

jingyi-gao-ttd (approved):

src/api/routers/usersRouter.ts
alex-yau-ttd: yep I agree!

PR #211, opened 2023-10-09
ajy-UID2-1929-Call-v2-keypairs-endpoint

alex-yau-ttd:
For manual testing with the keypair `name` populated, please check out `ajy-UID2-1929-Add-name-to-client-side-keypair-response` on `uid2-admin`. Otherwise, everything else should work but `name` will be blank.

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/99138dde-e4ac-410b-83e5-d139576e2159

### DIFF
--- a/src/api/services/adminServiceClient.ts
+++ b/src/api/services/adminServiceClient.ts
@@ -75,8 +75,9 @@ export const getSiteList = async (): Promise<SiteDTO[]> => {
 };
 
 export const getKeyPairsList = async (siteId: number): Promise<KeyPairDTO[]> => {
-  // convert this to use site-specific endpoint after UID2-1847
-  const response = await adminServiceClient.get<KeyPairDTO[]>('/api/client_side_keypairs/list');
+  const response = await adminServiceClient.get<KeyPairDTO[]>(
+    `api/v2/sites/${siteId}/client-side-keypairs`
+  );
   return response.data;
 };
 
--- a/src/api/services/adminServiceClient.ts
+++ b/src/api/services/adminServiceClient.ts
@@ -76,7 +76,7 @@ export const getSiteList = async (): Promise<SiteDTO[]> => {
 
 export const getKeyPairsList = async (siteId: number): Promise<KeyPairDTO[]> => {
   const response = await adminServiceClient.get<KeyPairDTO[]>(
-    `api/v2/sites/${siteId}/client-side-keypairs`
+    `/api/v2/sites/${siteId}/client-side-keypairs`
   );
   return response.data;
 };
--- a/src/web/components/KeyPairs/KeyPairModel.tsx
+++ b/src/web/components/KeyPairs/KeyPairModel.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable camelcase */
 import { KeyPairDTO } from '../../../api/services/adminServiceHelpers';
 
 export type KeyPairModel = {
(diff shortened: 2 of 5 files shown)


src/web/screens/keyPairsScreen.tsx
```diff
@@ -19,7 +19,8 @@ function KeyPairsScreen() {
     const data = await GetKeyPairs(participant!.id!);
     const sortedKeyPairs = data?.sort((a, b) => a.created.getTime() - b.created.getTime());
     setKeyPairData(sortedKeyPairs);
-  }, [participant]);
```
alex-yau-ttd, line 22: This was causing an extra re-run of this method, but I don't think it's necessary. Let me know if you think otherwise

src/api/services/adminServiceClient.ts
```diff
@@ -75,8 +75,9 @@ export const getSiteList = async (): Promise<SiteDTO[]> => {
 };
 
 export const getKeyPairsList = async (siteId: number): Promise<KeyPairDTO[]> => {
-  // convert this to use site-specific endpoint after UID2-1847
-  const response = await adminServiceClient.get<KeyPairDTO[]>('/api/client_side_keypairs/list');
+  const response = await adminServiceClient.get<KeyPairDTO[]>(
+    `api/v2/sites/${siteId}/client-side-keypairs`
```
jingyi-gao-ttd, line 79: nit: i think default local SSP_ADMIN_SERVICE_BASE_URL does not have '/' at the end. Maybe we should add slash in front of`/api/v2/sites/`

alex-yau-ttd: true, done!

src/web/screens/keyPairsScreen.tsx
jingyi-gao-ttd: I think there are two ways to solve it
1. make this call in loader, so this will only be called when it gets to the page
2. remove `participant!.id!` from GetKeyPairs, and use `participants/${participantId ?? 'current'}/keyPairs`. I am pretty sure this endpoint will go pass the enrichCurrentParticipant middleware

jingyi-gao-ttd: and we could remove   
`const { participant } = useContext(ParticipantContext);`

alex-yau-ttd: thanks! I went ahead and used your second suggestion to allow `GetKeyPairs` to be param-less. How's that look?

jingyi-gao-ttd: LGTM

alex-yau-ttd: the participant is used for `AddKeyPair`. Are you suggesting I do a similar treatment to that method and just make it not accept a `participantId`? So it will just be 
```
    const result = await axios.post(
      `/participants/current/keyPair/add`,
      props
    );
```

alex-yau-ttd: or less drastically, just make `participantId` on `AddKeyPairFormProps` optional

alex-yau-ttd: and have
```
    const result = await axios.post(
      `/participants/${participantId ?? 'current'}/keyPair/add`,
      props
    );
```

src/web/components/KeyPairs/KeyPairModel.tsx
```diff
@@ -15,12 +14,12 @@ export type KeyPairModel = {
 export const mapKeyPairDTOToModel = (dto: KeyPairDTO): KeyPairModel => {
   const model: KeyPairModel = {
     contact: dto.contact as string | undefined,
-    createdString: new Date(dto.created * 1000).toUTCString(),
-    created: new Date(dto.created * 1000),
+    createdString: new Date(dto.created).toISOString().split('T')[0],
```
jingyi-gao-ttd, line 17: nit: I wonder if we could use .toLocaleDateString() instead of `.toISOString().split('T')[0]`?

src/web/screens/keyPairsScreen.tsx
alex-yau-ttd: I've done the latter

jingyi-gao-ttd: Oh sorry, I didn't see participantId is using handleAddKeyPair. 
 I've noticed this `participant` changes after each rendering, I guess if it does not cause any rerender, we are fine with leave `const { participant } = useContext(ParticipantContext);`

jingyi-gao-ttd (approved):

src/web/components/KeyPairs/KeyPairModel.tsx
alex-yau-ttd: done!

PR #218, opened 2023-10-24
Fix participant types get deleted issue

jingyi-gao-ttd:
- Remove on delete CASCADE from participantToTypes and approvers
- Change upsert to use a transaction to update participant  types

### DIFF
--- a/src/api/services/participantsService.ts
+++ b/src/api/services/participantsService.ts
@@ -3,6 +3,7 @@ import { z } from 'zod';
 
 import {
   Participant,
+  ParticipantApprovalPartial,
   ParticipantCreationPartial,
   ParticipantDTO,
   ParticipantStatus,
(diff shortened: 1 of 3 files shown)


src/api/services/participantsService.ts
```diff
@@ -104,6 +105,27 @@ export const deleteSharingParticipants = async (
   );
 };
 
+export const updateParticipantWithRequestTypes = async (
```
alex-yau-ttd, line 108: this method is updating (deleting and/or inserting) the participant (request) types and also updating the name, siteId and status - to me, the name of the method indicates the request types are being updated, but not the other things. Maybe either split into multiple methods or use a different method name (e.g. `updateParticipantAndTypes`, though typically `and` means it should be multiple methods)?

Related question - does it have to be one transaction for any particular reason?

jingyi-gao-ttd: I think both options make sense, I updated the name and also the logic to update request types into a separate function. 

> does it have to be one transaction for any particular reason?

I want to make sure that updates to the account's name, status, and type all happen simultaneously. If, for any reason, something goes wrong during this process, I want the system to automatically revert to the original state, allowing the approver to try again. If we were to update these elements separately, there's a risk that the participant could end up in an inconsistent state if one update succeeds while another fails. For example, the participant status is marked as approved, but the participant types don't get updated, and this participant will disappear from the awaiting approval list, which approver has no way to fix it in the portal.

alex-yau-ttd (approved):

src/api/services/participantsService.ts
alex-yau-ttd: makes sense - thanks!

PR #221, opened 2023-10-26
Fix babel transpiling issue on models

jingyi-gao-ttd:
Manual testing 

https://github.com/IABTechLab/uid2-self-serve-portal/assets/116539741/fc9f0627-a187-405c-af77-0edaba29f504


.storybook/main.ts
```diff
@@ -0,0 +1,34 @@
+module.exports = {
```
alex-yau-ttd, line 1: Why do we need this change? Is it because of the later `@babel/core` version or something else?

alex-yau-ttd (approved):

package.json
```diff
@@ -93,7 +93,8 @@
     "start": "craco start",
     "build": "craco build",
     "test": "craco test --testPathPattern=src/web/",
-    "api": "nodemon -r dotenv/config ./src/api/api.ts",
+    "api": "nodemon -e \".ts\" --exec npm run babel-node-api",
+    "babel-node-api": "babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
```
alex-yau-ttd, line 97: What does this command actually do? Should we update the README?

alex-yau-ttd: oh right, `api` calls this, so I guess no README changes needed.

package.json
```diff
@@ -93,7 +93,8 @@
     "start": "craco start",
     "build": "craco build",
     "test": "craco test --testPathPattern=src/web/",
-    "api": "nodemon -r dotenv/config ./src/api/api.ts",
+    "api": "nodemon -e \".ts\" --exec npm run babel-node-api",
```
alex-yau-ttd, line 96: can you explain the actual difference this makes? Does it just ensure that it pulls in the babel config when running?

jingyi-gao-ttd: Yes, babel-node-api will use babel-node to transpile the file under api

lionell-pack-ttd (approved):

PR #225, opened 2023-11-22
Route-level error boundaries

alex-yau-ttd:
Use react router's [useRouteError](https://reactrouter.com/en/main/hooks/use-route-error) and [errorElement](https://reactrouter.com/en/main/route/error-element) to impose error boundaries at the router level.

Also factor out `ErrorView` into its own file, given it is now used by the `PortalErrorBoundary` and the `RouteErrorBoundary`. 

**Manual testing**
I've made local changes so that a front end error occurs on the Sharing Permissions page and an API error occurs on the Manage Participants page. Note that the front end error results in an `errorId`, while the backend shows an `errorHash`. Also note that other pages continue to work
 
https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/307c0f37-efff-4348-bd82-3324fb243363


src/web/utils/ErrorView.tsx
```diff
@@ -0,0 +1,28 @@
+const analyticsIdentifier = (errorId?: string, errorHash?: string) => {
```
alex-yau-ttd, line 1: Changed this to arrow notation, but otherwise it's a copy+paste

src/web/utils/ErrorView.tsx
```diff
@@ -0,0 +1,28 @@
...
+  errorId?: string;
+  errorHash?: string;
+}) {
+  return (
+    <div className='error-content'>
+      <img alt='Error icon' src='/uid2-logo.png' />
+      <div>Error</div>
+      <div>
+        {message ??
+          'Unexpected error encountered, please contact support if the problem persists and provide the information below'}
```
alex-yau-ttd, line 23: I added a default message - it is all we ever used anyway. I'm happy to remove the `message` prop entirely if we don't see the need for it.

src/web/utils/errorHandler.tsx
```diff
@@ -22,31 +22,6 @@ export interface RenderedErrorProps {
   componentStack?: string;
 }
 
-export function analyticsIdentifier(errorId?: string, errorHash?: string) {
```
alex-yau-ttd, line 25: Extracted to `src/web/utils/ErrorView.tsx`

lionell-pack-ttd (approved):

PR #227, opened 2023-11-26
Ensure local dev has same roles as prod

alex-yau-ttd:
Use a key that only has `SHARING_PORTAL`
Depends on https://github.com/IABTechLab/uid2-admin/pull/179


ssundahlTTD (approved):
Approved, but this should be merged after the admin changes.

PR #228, opened 2023-11-27
Include GUID in existing logging

alex-yau-ttd:
- Pass traceId from the request through to the service, so that it can be logged
- Refactor the loggers to reduce chance of using the wrong logger (to prevent have cases of `logger.error` instead of `errorLogger.logger`)
- We might have to revisit this design in the future to avoid having to prevent `traceId`s around everywhere, but that will involve a bit more work. Alternative options include using dependency injection or the factory pattern.

## Manual testing 
**Fail to add a sharing permission**
image

**Info when sending an email invite**
image

### DIFF
--- a/src/api/configureApi.ts
+++ b/src/api/configureApi.ts
@@ -23,6 +23,7 @@ import {
   getErrorLoggingMiddleware,
   getLoggers,
   getLoggingMiddleware,
+  getTraceId,
 } from './helpers/loggingHelpers';
 import makeMetricsApiMiddleware from './middleware/metrics';
 import { createParticipantsRouter } from './routers/participantsRouter';
@@ -167,7 +168,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
 
   app.use(getErrorLoggingMiddleware());
   const errorHandler: ErrorRequestHandler = (err, req, res, _next) => {
-    const traceId = req?.headers?.traceId?.toString() ?? '';
+    const traceId = getTraceId(req);
     errorLogger.error(`Fallback error handler invoked: ${err.message}`, traceId);
     if (err.statusCode === 401) {
       res.status(401).json({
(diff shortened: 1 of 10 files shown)


src/api/configureApi.ts
```diff
@@ -167,7 +167,8 @@ export function configureAndStartApi(useMetrics: boolean = true) {
 
   app.use(getErrorLoggingMiddleware());
   const errorHandler: ErrorRequestHandler = (err, req, res, _next) => {
-    logger.error(`Fallback error handler invoked: ${err.message}`);
+    const traceId = req?.headers?.traceId?.toString() ?? '';
```
alex-yau-ttd, line 170: `req.headers.traceId` has the type `string | string[] | undefined` for some reason, hence the need for `.toString()`

src/api/helpers/loggingHelpers.ts
```diff
@@ -50,8 +50,21 @@ const errorLogger = winston.createLogger({
...
+};
+
+const errorLoggerWrapper = {
+  error: (message: string, traceId: string) =>
+    errorLogger.error(`${message}, [traceId=${traceId}]`),
+};
+
 export const getLoggers = () => {
-  return [logger, errorLogger];
+  return {
```
alex-yau-ttd, line 63: Changed this to an object so that we didn't need to specify `logger`, etc. if we only needed `errorLogger`. Also so the order doesn't matter.

src/api/routers/participantsRouter.ts
```diff
@@ -90,6 +90,7 @@ export function createParticipantsRouter() {
 
   participantsRouter.post('/', async (req, res) => {
     try {
+      const traceId = req?.headers?.traceId?.toString() ?? '';
```
lionell-pack-ttd, line 93: This is repeated quite a few times. It might be worth creating a helper function, maybe in `loggingHelpers.ts`?

alex-yau-ttd: done! created `getTraceId` in `loggingHelpers.ts`

lionell-pack-ttd (approved):

PR #229, opened 2023-11-28
Filter out sites from sharing permissions features by using `visible` flag

alex-yau-ttd:
- Create a new method `getVisibleSiteList` in `adminServiceClient.ts` to call `getSiteList()` then filter out the appropriate sites before returning to the front end. 
- The method of filtering allows for existing sites that have `visible==null` to still be included, as that is the default for sites that were created before the `visible` flag was introduced. The default for new sites is `visible==true`
- Use `getVisibleSiteList` in the base (`/`) router and `/available` in the sites router. These routes are only used for non-admin functionality and does not break core functionality of the website when the current participant has a site that has visible set to false
- Ultimately this hides invisible sites from being included on the sharing permissions count on the home screen, as well as being hidden from the sharing permissions screen (i.e. in the bulk add control, the search and add control, and the sharing permissions table)

# Manual testing
Given the following sites exist:
```
  {
    "name": "a visible site",
    "clientTypes": [
      "DSP"
    ],
    "visible": true, //or null
    ...
  },
  {
    "name": "an invisible site",
    "clientTypes": [
      "DSP"
    ],
    "visible": false,
    ...
  }
```
We see that `a visible site` shows up in areas of the UI relating to sharing, but `an invisible site` does not. 
image

image

image

The `an invisible site` should still show up in admin areas such as approving participant requests.

image

Other manual testing verified that the functionality of the portal did not break when the user was associated with a site that had `visible==false`

### DIFF
--- a/src/web/components/ParticipantRequests/ParticipantApprovalForm.spec.tsx
+++ b/src/web/components/ParticipantRequests/ParticipantApprovalForm.spec.tsx
@@ -6,8 +6,16 @@ import { HighlightedResult } from './ParticipantApprovalForm';
 
 const createResult = (text: string, indices: [number, number][]) => {
   const result: Fuse.FuseResult<SiteDTO> = {
-    // eslint-disable-next-line camelcase
-    item: { name: text, id: 1, enabled: true, roles: [], clientTypes: [], client_count: 1 },
+    item: {
+      name: text,
+      id: 1,
+      enabled: true,
+      roles: [],
+      clientTypes: [],
+      // eslint-disable-next-line camelcase
+      client_count: 1,
+      visible: false,
+    },
     matches: [
       {
         indices,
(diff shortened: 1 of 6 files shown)


src/web/components/ParticipantRequests/ParticipantApprovalForm.spec.tsx
```diff
@@ -6,8 +6,16 @@ import { HighlightedResult } from './ParticipantApprovalForm';
...
-    item: { name: text, id: 1, enabled: true, roles: [], clientTypes: [], client_count: 1 },
+    item: {
+      name: text,
+      id: 1,
+      enabled: true,
+      roles: [],
+      clientTypes: [],
+      // eslint-disable-next-line camelcase
+      client_count: 1,
+      visible: false,
```
alex-yau-ttd, line 17: Sites with `visible==false` should still show in the participant approval form

src/web/components/ParticipantRequests/ParticipantRequestForm.stories.tsx
```diff
@@ -20,6 +20,7 @@ const response: SiteDTO[] = [
     clientTypes: ['PUBLISHER'],
     // eslint-disable-next-line camelcase
     client_count: 1,
+    visible: true,
```
alex-yau-ttd, line 23: It doesn't matter what value `visible` is in the participant request form, so I just did one with `true` and one with `false`

lionell-pack-ttd (approved):

PR #233, opened 2023-12-06
List participant on management page

joshua-smee-ttd:
image
Added list of Approved Participants to the management screen.

### DIFF
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -11,14 +11,17 @@ import {
   ParticipantSchema,
   ParticipantStatus,
 } from '../entities/Participant';
+import { ParticipantType } from '../entities/ParticipantType';
 import { UserDTO, UserRole } from '../entities/User';
 import { getTraceId } from '../helpers/loggingHelpers';
+import { mapClientTypeToParticipantType } from '../helpers/siteConvertingHelpers';
 import { getKcAdminClient } from '../keycloakAdminClient';
 import { isApproverCheck } from '../middleware/approversMiddleware';
 import {
   addKeyPair,
   getKeyPairsList,
   getSharingList,
+  getSiteList,
   setSiteClientTypes,
 } from '../services/adminServiceClient';
 import {
@@ -33,7 +36,7 @@ import {
   addSharingParticipants,
   checkParticipantId,
   deleteSharingParticipants,
-  getApprovedParticipant,
+  getApprovedParticipant as getParticipantsApproved,
   getParticipantsAwaitingApproval,
   ParticipantRequest,
   sendNewParticipantEmail,
@@ -95,9 +98,17 @@ export function createParticipantsRouter() {
     }
   );
 
-  participantsRouter.get('/list', isApproverCheck, async (req: ParticipantRequest, res) => {
-    const participants = await getApprovedParticipant();
-    const result = participants.map(p => )
+  participantsRouter.get('/approved', isApproverCheck, async (req, res) => {
+    const participants = await getParticipantsApproved();
+    const sites = await getSiteList();
+    const allParticipantTypes = await ParticipantType.query();
+    const result = participants.map((p) => {
+      const currentSite = sites.find((x) => x.id === p?.siteId);
+      return {
+        ...p,
+        types: mapClientTypeToParticipantType(currentSite?.clientTypes || [], allParticipantTypes),
+      };
+    });
     return res.status(200).json(result);
   });
 
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -106,13 +106,18 @@ export function createParticipantsRouter() {
     const siteMap = new Map<number, SiteDTO>(sitesList.map((s) => [s.id, s]));
 
     const allParticipantTypes = await ParticipantType.query();
-    const result = participants.map((p) => {
-      const currentSite = p?.siteId === undefined ? undefined : siteMap.get(p.siteId);
-      return {
-        ...p,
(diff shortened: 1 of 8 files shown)


src/api/routers/participantsRouter.ts
```diff
@@ -94,6 +98,20 @@ export function createParticipantsRouter() {
     }
   );
 
+  participantsRouter.get('/approved', isApproverCheck, async (req, res) => {
+    const participants = await getParticipantsApproved();
+    const sites = await getSiteList();
+    const allParticipantTypes = await ParticipantType.query();
```
alex-yau-ttd, line 104: We are basically doing the same thing here as we are in https://github.com/IABTechLab/uid2-self-serve-portal/blob/0db97a0ae8be4bcaf987ef257838fecdc8169af0/src/api/routers/usersRouter.ts#L59. Just that this is for all approved participants and the `usersRouter` is just for the current participant. Perhaps this warrants extracting to a separate method?

src/api/services/participantsService.ts
```diff
@@ -73,6 +73,10 @@ export const getAttachedSiteIDs = async (): Promise<SiteIdType[]> => {
   return sites.map((s) => s.siteId);
 };
 
+export const getApprovedParticipant = async (): Promise<Participant[]> => {
```
alex-yau-ttd, line 76: maybe just name this `getParticipantsApproved` so we don't rename the import in `participantsRouter`?

alex-yau-ttd: Alternatively we could just call it `getParticipants`? We already differentiate it from the ones awaiting approval. What do you think?

alex-yau-ttd:
Could you please include a screenshot of the new UI :D

alex-yau-ttd:
> image Sorry, here's a screen shot of the new table.

can you please put it in the MR description instead so it's easier to find

src/web/components/ParticipantRequests/ApprovedParticipantsTable.tsx
```diff
@@ -22,14 +22,15 @@ function NoParticipants() {
...
         <thead>
           <tr>
             <th>Participant Name</th>
             <th>Participant Type</th>
-            <th>Allowed API Roles</th>
+            {/* <th>Allowed API Roles</th>
             <th>Name of Approver</th>
-            <th>Date of Approval</th>
+            <th>Date of Approval</th> */}
             <th className='action'>Actions</th>
```
alex-yau-ttd, line 34: Per Scott's point, maybe we can even just delete `Actions` as well?

src/web/components/ParticipantRequests/ApprovedParticipantsTable.stories.tsx
```diff
@@ -18,59 +18,104 @@ export const AllParticipants: Story = {
...
+            role: UserRole.Marketing,
+            acceptedTerms: false,
+          },
+        ],
       },
       {
         id: 2,
         name: 'Participant 2',
+        types: [{ id: 1, typeName: 'Type 1' }],
+        status: ParticipantStatus.AwaitingApproval,
```
alex-yau-ttd, line 40: This component just displays all participants it is given, though we do just plan for it to be for participants that have `ParticipantStatus.Approved`. What do you think about changing the dummy data to reflect that?

src/api/services/participantsService.ts
joshua-smee-ttd: I think keeping it as approved is probably better. There's some chance someone might want a function that just returns all participants.

src/api/routers/participantsRouter.ts
```diff
@@ -33,6 +36,7 @@ import {
   addSharingParticipants,
   checkParticipantId,
   deleteSharingParticipants,
+  getApprovedParticipant as getParticipantsApproved,
```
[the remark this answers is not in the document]
joshua-smee-ttd: Honestly not sure sorry. Have fixed it.

src/web/components/ParticipantRequests/ApprovedParticipantsTable.stories.tsx
joshua-smee-ttd: Yeah I've done that. I thought I already had but must've gotten it mixed up.

src/api/routers/participantsRouter.ts
alex-yau-ttd: true, I guess it's sufficiently different now.

src/web/components/ParticipantRequests/ParticipantRequestsTable.tsx
```diff
@@ -30,6 +30,7 @@ export function ParticipantRequestsTable({
 }: ParticipantRequestsTableProps) {
   return (
     <div className='participant-requests-container'>
+      <h2>{participantRequests.length} Participants Pending Approval</h2>
```
alex-yau-ttd, line 33: Let's conditionally render the string to be singular when there is only one `participantRequest` to avoid 
image

src/web/screens/manageParticipants.tsx
```diff
@@ -42,15 +45,19 @@ function ManageParticipants() {
...
             ParticipantRequestDTO[],
+            ParticipantDTO[],
             ParticipantTypeDTO[]
           ]) => (
-            <ParticipantRequestsTable
-              participantRequests={participantRequests}
-              participantTypes={participantTypes}
-              onApprove={handleApproveParticipantRequest}
-            />
+            <div>
```
alex-yau-ttd, line 53: nit: I'd recommend just using `<>` here instead of a `<div>`. I think it ends up being a `<div>` in the DOM anyway but it reads a bit cleaner. You'd want to use `<div>` if it was being used for styling.

alex-yau-ttd (approved):

PR #235, opened 2023-12-08
Introduce dependency injection

alex-yau-ttd:
- Add [inversify](https://github.com/inversify/InversifyJS) & [inversify-express-utils](https://www.npmjs.com/package/inversify-express-utils)
- Introduce new pattern so that `usersRouter` calls new `UserController`, which is injected with a `UserService`. I've currently only implemented the `UserService` methods for a few endpoints - I've left a lot of logic in the controller for now for simplicity.

**Future work**
Use DI so that the UserService gets a logger


src/api/tests/utils.ts
```diff
@@ -31,9 +31,9 @@ function useTestServer() {
   });
   beforeAll(async () => {
     const serverDetails = configureAndStartApi(false);
-    api = serverDetails.server;
```
alex-yau-ttd, line 34: This line results in:
```
Type 'InversifyExpressServer' is missing the following properties from type 'Server<typeof IncomingMessage, typeof ServerResponse>': setTimeout, maxHeadersCount, maxRequestsPerSocket, timeout, and 27 more.ts(2740)
```
Still trying to figure out how to resolve.

src/api/tests/utils.ts
```diff
@@ -31,9 +31,9 @@ function useTestServer() {
...
+    // api = serverDetails.server;
     routers = serverDetails.routers;
-    terminator = createHttpTerminator({ server: api });
```
alex-yau-ttd, line 36: This line results in:
```
Type 'Server<typeof IncomingMessage, typeof ServerResponse> | null' is not assignable to type 'Server<...> | Http2SecureServer | Server<...>'.
  Type 'null' is not assignable to type 'Server<...> | Http2SecureServer | Server<...>'.ts(2322)
types.d.ts(13, 14): The expected type comes from property 'server' which is declared here on type 'HttpTerminatorConfigurationInput'
```
Still trying to figure out how to resolve.

src/api/constant/types.ts
lionell-pack-ttd, line 1: Will this file end up needing all of the types involved in DI? It's not too bad to maintain, but it'd be nice if we didn't need it.

alex-yau-ttd: yes it will - it has to be in a separate file, per https://github.com/inversify/InversifyJS?tab=readme-ov-file#step-1-declare-your-interfaces-and-types. Agree it won't be too bad to maintain.

src/api/routers/usersRouter.ts
```diff
@@ -1,153 +1,30 @@
...
-    const allParticipantTypes = await ParticipantType.query();
-    const result = {
-      ...currentParticipant,
-      types: mapClientTypeToParticipantType(currentSite?.clientTypes || [], allParticipantTypes),
-    };
-    return res.status(200).json(result);
-  });
+  const userController = new UserController(new UserService());
 
+  usersRouter.use('/current', enrichCurrentUser);
```
lionell-pack-ttd, line 11: We'll have to be careful about the orderings here (because some of them are performing auth checks), and ... if we don't have to use different middleware for different routes, can we just rely on the annotations to register the routers with the right HTTP verbs?

One option would be to use a different controller class depending on which middleware (e.g. enrichCurrentUser, enrichWithUserFromParams, etc) that route needs. I.e. there could be a CurrentUserController and a ParamsUserController.

src/api/tests/utils.ts
alex-yau-ttd: fixed!

src/api/tests/utils.ts
alex-yau-ttd: fixed!

PR #238, opened 2023-12-12
Add participant api role table

joshua-smee-ttd:
No visual changes. Added table and linked to participants using join table. Updated seed data so they have API roles linked to each participant.

image
apiRoles table

image
participantsToApiRoles when loaded from the seed.

image
ER diagram of DB


src/api/entities/ApiRole.ts
```diff
@@ -0,0 +1,34 @@
...
+      },
+    },
+  };
+  declare id: number;
+  declare roleName: string;
+}
+
+export type ApiRoleDTO = ModelObjectOpt<ApiRole>;
+
+export const ApiRoleTypeSchema = z.object({
```
lionell-pack-ttd, line 31: These schema are just for parsing - I know we have a few in models, but we've seen some bugs where people assume they're shared and should model the entity itself, and change them. I think they should really be closer to the thing that's parsing something (probably the route handlers).

src/database/migrations/20231211041333_CreateApiTable.ts
```diff
@@ -0,0 +1,32 @@
...
+  const defaultRoles = ['MAPPER', 'GENERATOR', 'ID_READER', 'SHARER', 'OPTOUT'];
+  await knex('apiRoles').insert(
+    defaultRoles.map((role) => {
+      return { roleName: role };
+    })
+  );
+
+  await knex.schema.createTable('participantsToRoles', (table) => {
+    table.integer('participantId').references('participants.id').onDelete('CASCADE');
+    table.integer('apiRoleId').references('id').inTable('apiRoles').onDelete('CASCADE');
```
lionell-pack-ttd, line 24: I don't think we want the cascade here: if someone tries to delete a role that's in use, I'd prefer to have the delete fail.

The delete on participants is OK - I don't imagine we'll delete participants often (or ever), but if we do, removing the linking table entry makes sense.

src/database/migrations/20231211041333_CreateApiTable.ts
```diff
@@ -0,0 +1,32 @@
+import { Knex } from 'knex';
+
+import { ApiRole } from '../../api/entities/ApiRole';
```
alex-yau-ttd, line 3: These imports aren't used

src/database/migrations/20231211041333_CreateApiTable.ts
```diff
@@ -0,0 +1,32 @@
+import { Knex } from 'knex';
```
alex-yau-ttd, line 1: nit: filename could be `...CreateApiRolesTables...`? It's a bit more accurate

src/database/migrations/20231211041333_CreateApiTable.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  // Add default roles
+  const defaultRoles = ['MAPPER', 'GENERATOR', 'ID_READER', 'SHARER', 'OPTOUT'];
+  await knex('apiRoles').insert(
+    defaultRoles.map((role) => {
+      return { roleName: role };
+    })
+  );
+
+  await knex.schema.createTable('participantsToRoles', (table) => {
```
alex-yau-ttd, line 22: Perhaps `participantsToApiRoles` instead of `participantsToRoles`? Since we may implement roles (by user) in the future - this would avoid future confusion.

src/database/migrations/20231211041333_CreateApiTable.ts
lionell-pack-ttd, line 3: I think a lot of these imports are unnecessary - better to remove them if they're not needed.

src/api/entities/Participant.ts
```diff
@@ -29,6 +30,18 @@ export class Participant extends BaseModel {
         to: 'participantTypes.id',
       },
     },
+    roles: {
```
alex-yau-ttd, line 33: thoughts on `apiRoles` instead? Given we may introduce roles per user in the future.

joshua-smee-ttd: Ah yeah. Sounds good. I'll go shift those names around everywhere.

src/database/migrations/20231211041333_CreateApiTable.ts
joshua-smee-ttd: Cool, yeah that makes sense.

src/api/entities/ApiRole.ts
joshua-smee-ttd: I think for now it may still be needed as it's used by the Participant Schema?

src/database/seeds/Participants.ts
```diff
@@ -78,6 +80,18 @@ export async function CreateParticipant(
     participantId: participant[0].id as number,
     participantTypeId: participantType[0].id as number,
   });
+
+  await AllowedSiteRoles[type].forEach(async (role) => {
```
alex-yau-ttd, line 84: It smells a bit odd that a method in `api/services` is being called in a seed file. I think it might be better to just update the `sampleData` above to include the `apiRoles` manually. Or at least move the method to a more generic folder, maybe `api/helpers`. Or perhaps @lionell-pack-ttd can correct me if I'm off-base here?

alex-yau-ttd (approved):

src/database/seeds/Participants.ts
joshua-smee-ttd: Yeah I think that's fair. I'm happy to basically just copy over the strings and use them in the loop instead?

lionell-pack-ttd: Seeds doesn't matter as much - I think it's OK to point to `api/services`, but it doesn't hurt to avoid the dependency as well.

It is important to avoid importing things to schema migrations. The migrations need to keep doing the same thing even as the code outside changes - if we need to change a table, it's a new migration. If we use code from somewhere like `api/services` in a migration, when that code changes the migration might change behaviour, and we wouldn't want that.

PR #239, opened 2023-12-13
Show participant API Roles in Management Page

joshua-smee-ttd:
image
Shows the API roles each participant has on the Participant Management page.

image
Modified the apiRoles table to have a second column storing names to show on site for roles. 'test' was role used to check behavior when role not specifically mapped to externalName.

image
Storyboard of the table.


src/database/migrations/20231213033935_addDescriptiveNameToApiRoles.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('apiRoles', (table) => {
+    table.string('externalName', 128);
+  });
+
+  await knex('apiRoles').where({ roleName: 'GENERATOR' }).update({ externalName: 'Token Creator' });
+  await knex('apiRoles').where({ roleName: 'MAPPER' }).update({ externalName: 'UID2 Creator' });
+  await knex('apiRoles').where({ roleName: 'ID_READER' }).update({ externalName: 'Decryptor' });
+  await knex('apiRoles').where({ roleName: 'SHARER' }).update({ externalName: 'Securer' });
+  await knex('apiRoles').where({ roleName: 'OPTOUT' }).update({ externalName: 'OPTOUT' });
```
alex-yau-ttd, line 12: nit: perhaps the externalName should be "Opt-out"

joshua-smee-ttd: Sounds good! Have changed.

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
```diff
@@ -8,6 +8,13 @@
     margin-right: 8px;
   }
 
+  .participant-request-api-role-label {
```
alex-yau-ttd, line 11: It feels a bit odd that the ApprovedParticipantItem is using styling from ParticipantRequestItem.scss. It does make sense that they share some styling, but maybe there is a more clear way to can do this. Few options:
-  Have a common scss file, something like `ManageParticipantItem.scss` which is shared by approved participants and participant requests.
- Completely separate the scss files so that ParticipantRequestItem is the only user of ParticipantRequestItem.scss - and ApprovedParticipantItem is the only user of ApprovedParticipantItem.
- Something else?

I could be wrong here though - what do you think?

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
```diff
@@ -8,6 +8,13 @@
...
+    background: var(--theme-label);
```
alex-yau-ttd, line 12: An alternative to the more severe refactoring above would to extract some of the styling within this file, e.g. something like:
```
.participant-request-label {
  border-radius: 3px;
  padding: 4px 8px;
  margin-right: 8px;
  background: var(--theme-label-background);
}

.participant-request-type-label {
  @extend .participant-request-label;
}

.participant-request-api-role-label {
  @extend .participant-request-label;
  background: var(--theme-label);
}
```

src/web/components/ParticipantRequests/ParticipantRequestItem.tsx
```diff
@@ -48,7 +48,7 @@ export function ParticipantRequestItem({
         <div className='participant-name'>{participant.name}</div>
       </td>
       <td>
-        <div className='participant-types'>{getParticipantTypes(participant.types)}</div>
+        <div className='participant-request-types'>{getParticipantTypes(participant.types)}</div>
```
alex-yau-ttd, line 51: There might be some transition period between the existing names and external names - perhaps it would be handy to have a tooltip that shows the roleName? It may not strictly be necessary, so worth checking with the team.

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
joshua-smee-ttd: I think that's a fair observation. This set of components has a slightly strange naming all around since it contains the approved participants table, the requested participants and the form for that. My opinion would be to rename the whole folder they're in to be more along the lines of ```ManagementParticipant``` and then rename the two scss files to be along the lines of ```ManagementParticipantTable.scss``` and  ```ManagementParticipantItem.scss``` then slightly rename the classes inside to reflect this.

src/web/components/ParticipantRequests/ParticipantRequestItem.tsx
joshua-smee-ttd: I really like that idea actually, since I wasn't really sure the best way to go about it anyway. I'll put that in there.

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
```diff
@@ -20,6 +27,13 @@
     line-height: 0.75rem;
   }
 
+  .participant-request-api-roles {
```
alex-yau-ttd, line 30: Similar comment to above - perhaps worth refactoring this block

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
joshua-smee-ttd: That name is also what the stories are actually under on storybook.
image

alex-yau-ttd: Something like that sounds good to me, as long as we can easily tell the difference between the participant requests and approved participants (as well as stuff that is shared!).

joshua-smee-ttd: Have done a bunch of refactoring to it. Sorry for the mass of changes.

See screenshots below of unchanged storyboard.
image
image

src/web/components/ParticipantRequests/ParticipantRequestItem.tsx
joshua-smee-ttd: Have added this in:
chrome_rbJo6o15IE

src/web/components/ParticipantRequests/ParticipantRequestItem.scss
alex-yau-ttd: The font weight of the API Roles looks a bit too strong now (i.e. it looks bold) otherwise looks good visually! Still reading the actual code

alex-yau-ttd: I think it is inheriting a font-weight of 600 from the tooltip, if that helps.

joshua-smee-ttd: Yeah it was. I think that this is the only spot the trigger as text was given so just adjusted its formatting.

src/database/migrations/20231213033935_addDescriptiveNameToApiRoles.ts
```diff
@@ -0,0 +1,27 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('apiRoles', (table) => {
+    table.string('externalName', 128);
+  });
+
+  await knex('apiRoles').where({ roleName: 'GENERATOR' }).update({ externalName: 'Token Creator' });
+  await knex('apiRoles').where({ roleName: 'MAPPER' }).update({ externalName: 'UID2 Creator' });
+  await knex('apiRoles').where({ roleName: 'ID_READER' }).update({ externalName: 'Decryptor' });
```
alex-yau-ttd, line 10: This has been updated to "Bidder"

joshua-smee-ttd: What has been changed sorry?

alex-yau-ttd: Decryptor

alex-yau-ttd (approved):

PR #240, opened 2023-12-18
Extract handleErrorPopup method

alex-yau-ttd:
- Extract `handleErrorPopup` method so that it is no longer defined locally multiple times
- Use `handleErrorPopup` instead of manually setting a message in the frontend - if the messages aren't friendly enough we can just change them on our backend.
- Ensure the `errorHash` is extracted from different error types
- Stronger type safety

### DIFF
--- a/src/web/utils/apiError.tsx
+++ b/src/web/utils/apiError.tsx
@@ -28,17 +28,32 @@ export function backendError(e: unknown, overrideMessage: string) {
   return Error(overrideMessage);
 }
 
-export function handleErrorPopup(
-  e: Error,
+const isError = (obj: unknown): obj is Error => obj instanceof Error;
+
+const getHash = (e: Error) => {
+  let result = '';
+  if (e instanceof ApiError) {
+    result = e.errorHash ?? '';
+  }
+  if (e instanceof AxiosError) {
+    result = (e.response?.data?.errorHash as string) ?? '';
+  }
+  return result;
+};
+
+export const handleErrorPopup = (
+  e: unknown,
   setStatusPopup: React.Dispatch<React.SetStateAction<StatusNotificationType | undefined>>,
   setShowStatusPopup: React.Dispatch<React.SetStateAction<boolean>>
-) {
-  const hasHash = Object.hasOwn(e, 'errorHash') && (e as ApiError).errorHash;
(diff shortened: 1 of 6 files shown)


ssundahlTTD (approved):

src/web/utils/apiError.tsx
```diff
@@ -25,3 +27,18 @@ export function backendError(e: unknown, overrideMessage: string) {
...
+export function handleErrorPopup(
+  e: Error,
+  setStatusPopup: React.Dispatch<React.SetStateAction<StatusNotificationType | undefined>>,
+  setShowStatusPopup: React.Dispatch<React.SetStateAction<boolean>>
+) {
+  const hasHash = Object.hasOwn(e, 'errorHash') && (e as ApiError).errorHash;
+  const hash = hasHash ? `: (${(e as ApiError).errorHash})` : '';
+  setStatusPopup({
+    type: 'Error',
+    message: `${e.message}${hash}`,
```
lionell-pack-ttd, line 40: We don't know that `e` has a `.message` - it just looks like we do because of the `as` in the catch blocks - so in some cases `.message` might be undefined.

src/web/utils/apiError.tsx
```diff
@@ -25,3 +27,18 @@ export function backendError(e: unknown, overrideMessage: string) {
...
+  });
+  setShowStatusPopup(true);
+  throw new Error(e.message);
```
lionell-pack-ttd, line 43: This loses the call stack of the original exception and re-creates it here - if it's a real exception, you should either re-throw the original exception (to preserve the stack trace) or provide the original exception to the Error constructor - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#rethrowing_an_error_with_a_cause

src/web/utils/apiError.tsx
alex-yau-ttd: good point - I've added a type check at the top of the method.

src/web/utils/apiError.tsx
alex-yau-ttd: good point - I'll just throw the original error: `throw e`.

lionell-pack-ttd (approved):

PR #241, opened 2023-12-19
Add null check when getting participant site

joshua-smee-ttd:


src/api/services/userService.ts
```diff
@@ -34,7 +34,7 @@ export class UserService {
       id: req.user!.participantId,
     });
     const currentSite =
-      currentParticipant?.siteId === undefined
+      currentParticipant?.siteId === undefined || currentParticipant.siteId === null
```
alex-yau-ttd, line 37: nice! I think we can just simplify this to `!currentParticipant?.siteId` - that checks for [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

joshua-smee-ttd: Ah yeah cool we could. Have made that change!

alex-yau-ttd (approved):

PR #242, opened 2023-12-19
Display errorHash when resending team member invite

alex-yau-ttd:
**Before**
image

**After**
image

Also added `return` statements so that we don't get the following api error by setting the status multiple times:
```
Error: Cannot set headers after they are sent to the client
```

### DIFF
--- a/src/api/controllers/userController.ts
+++ b/src/api/controllers/userController.ts
@@ -90,14 +90,24 @@ export class UserController {
 
     const resultLength = user?.length ?? 0;
     if (resultLength < 1) {
-      res.sendStatus(404);
+      errorLogger.error(
+        `No results received when loading user entry for ${req.user?.email}`,
+        traceId
+      );
+      res.status(404).json({
+        errorHash: traceId,
+      });
+      return;
     }
     if (resultLength > 1) {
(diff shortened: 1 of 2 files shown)


src/api/controllers/userController.ts
```diff
@@ -90,14 +90,24 @@ export class UserController {
 
     const resultLength = user?.length ?? 0;
     if (resultLength < 1) {
-      res.sendStatus(404);
+      errorLogger.error(
+        `No results received when loading user entry for ${req.user?.email}`,
+        traceId
+      );
+      res.status(404).json({
+        errorHash: traceId,
```
lionell-pack-ttd, line 98: It would be nice to have a pattern where we wouldn't have to explicitly include the traceId in every route handler response that needs it. No need to do that in this PR, but it's worth thinking about whether there's a good way.

lionell-pack-ttd (approved):

src/api/controllers/userController.ts
alex-yau-ttd: yeah I'm hoping the DI work will help with this, but haven't fully thought it through yet.

PR #243, opened 2023-12-19
API Key page

joshua-smee-ttd:
image
Added a page to show API keys

chrome_DpQbUBRsVR
Tooltip to explain keyId and its importance.

chrome_yxJveG5Fj8
Turned cell to show roles into a component

Relies on Admin endpoint added here: https://github.com/IABTechLab/uid2-admin/pull/191
Fixed spelling mistake in seed data

image
Storybook for api key table


alex-yau-ttd:
Let's get rid of the padding here so that the start of the label lines up with the start of the column heading: 
image

src/web/components/ApiKeyManagement/KeyItem.scss
```diff
@@ -0,0 +1,3 @@
+.ApiKeyItem {
```
alex-yau-ttd, line 1: In our usual naming pattern this would be `api-key-item` - do you mind changing this please?

alex-yau-ttd: This applies to some other class names in this PR as well

joshua-smee-ttd: Have fixed these

joshua-smee-ttd:
image
New api roles cell.

src/web/components/ApiKeyManagement/ApiRolesCell.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  ],
+};
+export default meta;
+
+type Story = StoryObj<typeof ApiRolesCell>;
+
+export const ManyRoleShowingRoleName: Story = {
+  args: {
+    apiRoles: [
+      { id: 1, roleName: 'Role1', externalName: 'Role 1' },
```
lionell-pack-ttd, line 36: We've seen a lot of confusion in sprint showcases and other feedback sessions before when we had fake names - I think it's worthwhile to make these match the actual values we use.

lionell-pack-ttd: I think the worst was when we had some stories showing participant types as "Type 1", "Type 2", etc. - people kept losing track of what they actually were, and it got much easier once we switched to using "Advertiser", "DSP", etc.

src/web/components/ApiKeyManagement/ApiRolesCell.scss
```diff
@@ -0,0 +1,12 @@
+.api-roles-cell {
+  display: flex;
+  font-size: 0.75rem;
+  line-height: 0.75rem;
```
alex-yau-ttd, line 4: We prefer to use `line-height` in a unitless manner to act as a multiplier. It avoids situations where the text is too big for the line-height. You may need to change the number from 0.75 though

src/web/components/ApiKeyManagement/KeyItem.scss
lionell-pack-ttd, line 1: When you have sub-components that are tightly coupled to parent components and you won't be testing them separately in something like StoryBook, I don't think it's necessary to split out the styling into separate scss files - you can collect styles together into a single file for the parent component.

No need to change, and if you prefer to split out the styles it's fine.

src/web/components/ApiKeyManagement/KeyTable.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        roles: [],
+        service_id: 0,
+      },
+      {
+        contact: 'ApiKey2',
+        name: 'ApiKey2',
+        created: 1701210253,
+        key_id: 'FDSL,089',
+        disabled: false,
+        roles: [{ id: 1, roleName: 'Role1', externalName: 'Role 1' }],
```
lionell-pack-ttd, line 32: As above - it'll be a bit easier to think about these and get helpful feedback if the roles match the actual roles we use.

src/api/services/adminServiceHelpers.ts
```diff
@@ -80,3 +82,47 @@ export const mapClientTypesToAdminEnums = (
...
+  contact: string;
+  roles: ApiRoleDTO[];
+  created: number;
+  disabled: boolean;
+  service_id: number;
+};
+
+export type ApiKeyAdmin = Omit<ApiKeyDTO, 'roles'> & { roles: string };
+
+export async function mapApiKeyDTO(adminApiKeys: ApiKeyAdmin[]): Promise<ApiKeyDTO[]> {
```
alex-yau-ttd, line 98: my tired brain is finding this method a bit hard to understand. A few asks:
* Method name - are we mapping from ApiKeyDTO or to ApiKeyDTO? Perhaps we could rename this to something like `mapAdminApiKeysToApiKeyDTOs`?
* Could we possibly extract some of the logic in this to a well-named function to more clearly convey the functionality?

src/web/components/ApiKeyManagement/KeyItem.scss
```diff
@@ -0,0 +1,3 @@
+.api-key-tem {
```
alex-yau-ttd, line 1: typo here ;)

src/web/services/participant.ts
```diff
@@ -99,6 +99,18 @@ export async function GetApprovedParticipants() {
...
 
+export async function GetParticipantApiKeys(participantId?: number) {
+  try {
+    const result = await axios.get<ApiKeyDTO[]>(
+      `/participants/${participantId ?? 'current'}/apiKeys`
+    );
+
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not get the apiKeys');
```
alex-yau-ttd, line 110: We should assume that this message will end up getting shown on the UI. In that case, let's change "apiKeys" -> "API keys"

src/web/components/ApiKeyManagement/ApiRolesCell.stories.tsx
joshua-smee-ttd: Yeah that makes sense

src/web/components/ApiKeyManagement/KeyItem.scss
joshua-smee-ttd: Well I found a better solution which was one of them wasn't doing anything anyway... Have removed.

src/web/components/ApiKeyManagement/ApiRolesCell.scss
joshua-smee-ttd: That makes sense. Should I change the same thing in `ParticipantManagementItem.scss` since that has `.participant-item-types` which uses the same css.

src/api/services/adminServiceHelpers.ts
joshua-smee-ttd: Have reffactored it a bit, moved the main extra mapping function out and added a few comments.

src/web/components/ApiKeyManagement/ApiRolesCell.scss
joshua-smee-ttd: image
Have done this for both. Used 1.15 to match the line height for buttons which is being used by the apiRolesCell when showing the tooltip

src/web/screens/apiManagement.tsx
```diff
@@ -0,0 +1,40 @@
+import { Suspense } from 'react';
```
alex-yau-ttd, line 1: nit: perhaps this file would be better named `apiKeyManagement.tsx` or just `keyManagement.tsx`?

src/web/screens/apiManagement.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Await, defer, useLoaderData } from 'react-router-dom';
+
+import { ApiKeyDTO } from '../../api/services/adminServiceHelpers';
+import KeyTable from '../components/ApiKeyManagement/KeyTable';
+import { Loading } from '../components/Core/Loading';
+import { GetParticipantApiKeys } from '../services/participant';
+import { RouteErrorBoundary } from '../utils/RouteErrorBoundary';
+import { PortalRoute } from './routeUtils';
+
+function ManageParticipants() {
```
alex-yau-ttd, line 11: Let's rename this function to whatever we end up calling this file.

src/web/screens/apiManagement.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const data = useLoaderData() as {
+    result: ApiKeyDTO[];
+  };
+
+  return (
+    <div>
+      <h1>Manage API Keys</h1>
+      <p className='heading-details'>View and manage you API keys.</p>
```
alex-yau-ttd, line 19: typo: you -> your

src/web/screens/apiManagement.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <Await resolve={data.result}>
+          {(apiKeys: ApiKeyDTO[]) => <KeyTable apiKeys={apiKeys.filter((key) => !key.disabled)} />}
+        </Await>
+      </Suspense>
+    </div>
+  );
+}
+
+export const ApiManagementRoute: PortalRoute = {
+  path: '/dashboard/api',
```
alex-yau-ttd, line 30: I think this should probably match the file name too. `api` is a bit confusing because it's not an API per -e.

src/web/services/participant.ts
```diff
@@ -99,6 +99,18 @@ export async function GetApprovedParticipants() {
...
 
+export async function GetParticipantApiKeys(participantId?: number) {
+  try {
+    const result = await axios.get<ApiKeyDTO[]>(
+      `/participants/${participantId ?? 'current'}/apiKeys`
+    );
+
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not get participants API Keys');
```
alex-yau-ttd, line 110: nit: I'd probably drop the plural of "participants" -> "participant"

src/web/components/ApiKeyManagement/KeyTable.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { TableNoDataPlaceholder } from '../Core/TableNoDataPlaceholder';
+import { Tooltip } from '../Core/Tooltip';
+import KeyItem from './KeyItem';
+
+import './KeyTable.scss';
+
+function NoKeys() {
+  return (
+    <TableNoDataPlaceholder
+      icon={<img src='/group-icon.svg' alt='group-icon' />}
```
alex-yau-ttd, line 11: I think it'll be a rare case where we display the empty table, but we may eventually want to change this icon. No action needed for now, though.

src/api/services/adminServiceHelpers.ts
```diff
@@ -80,3 +82,58 @@ export const mapClientTypesToAdminEnums = (
...
+export type ApiKeyAdmin = Omit<ApiKeyDTO, 'roles'> & { roles: string };
+
+function mapAdminRolesToApiRoleDTOs(adminRoles: string[], apiRoleMap: Map<String, ApiRoleDTO>) {
+  return adminRoles.map((adminRole) => {
+    const adminRoleTrimmed = adminRole.trim();
+
+    let apiRole = apiRoleMap.get(adminRoleTrimmed);
+
+    if (apiRole === undefined) {
+      // Add new role to map if not found. Stocks duplicate new objects
```
alex-yau-ttd, line 105: to be a purist, while comments have good intentions, it may be an indication that the code is not very easy to understand and perhaps could be treated in some way (e.g. extract method).

Think about if this logic could be simplified, or if not, maybe we don't need the comments at all.

Worth checking out https://refactoring.guru/smells/comments for this and other code smells!

alex-yau-ttd: that being said, it's not like all your code is littered with comments - I'm being a bit picky here

alex-yau-ttd: if you decide to leave it in - what does "Stocks duplicate new objects" mean?

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -0,0 +1,25 @@
+import { ApiKeyDTO } from '../../../api/services/adminServiceHelpers';
+import ApiRolesCell from './ApiRolesCell';
+
+function formatDate(timeValue: number) {
```
alex-yau-ttd, line 4: Two things: 
1. Could this method possibly be exported from a more generic utils folder, or is this specific to KeyItems?
2. Could we possibly avoid magic numbers and extract `1000` to a variable which explains what it is?

src/web/components/ApiKeyManagement/KeyTable.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        key_id: 'FDSL,089',
+        disabled: false,
+        roles: [{ id: 1, roleName: 'MAPPER', externalName: 'Mapper' }],
+        service_id: 0,
+      },
+      {
+        contact: 'ApiKey3',
+        name: 'ApiKey3',
+        created: 1702830516,
+        key_id: 'FDSL,089',
```
alex-yau-ttd, line 39: I'm guessing this shouldn't have the same `key_id` as ApiKey2?

src/web/components/ApiKeyManagement/ApiRolesCell.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import ApiRolesCell from './ApiRolesCell';
+
+const meta: Meta<typeof ApiRolesCell> = {
+  component: ApiRolesCell,
+  title: 'ApiManagement/Api Roles cell',
```
alex-yau-ttd, line 7: nit: `API Management/API Roles cell`

src/web/components/ApiKeyManagement/ApiRolesCell.tsx
```diff
@@ -0,0 +1,26 @@
+import { ApiRoleDTO } from '../../../api/entities/ApiRole';
+import { Tooltip } from '../Core/Tooltip';
+
+import './ApiRolesCell.scss';
+
+type ApiRolesProps = {
+  apiRoles: ApiRoleDTO[];
+  showRole?: Boolean;
```
alex-yau-ttd, line 8: I find this prop a bit unclear.
1. Can we perhaps give it a clearer name, perhaps `showRoleNameToolTip`
2. We can use `boolean` instead of `Boolean` here - the primitive version is preferred here

src/web/screens/apiManagement.tsx
joshua-smee-ttd: Ah yeah that sounds better. I'll make sure it gets renamed everywhere.

src/web/components/ApiKeyManagement/KeyTable.tsx
joshua-smee-ttd: Yeah that was my thinking. It seemed better to have something here then nothing.

src/api/services/adminServiceHelpers.ts
joshua-smee-ttd: Yeah I don't think they're necessary, going to remove them. Was probably just being overly careful. Ahaha that's my attempt at spelling stop... sorry.

src/web/components/ApiKeyManagement/KeyItem.tsx
joshua-smee-ttd: Moved the value to a constant. I'm not sure it's worth moving to a generic. It's somewhat specific since it's receiving a Unix timestamp in seconds which is always a given.

src/web/components/ApiKeyManagement/ApiRolesCell.tsx
joshua-smee-ttd: Have renamed

src/web/components/ApiKeyManagement/KeyItem.tsx
alex-yau-ttd: Unix timestamps are used all over admin - perhaps it is indeed worth extracting out and renaming to `formatUnixDate`

joshua-smee-ttd: Sounds good. Have moved to the function to ```textHelpers.ts```.

src/web/components/ApiKeyManagement/ApiRolesCell.tsx
```diff
@@ -0,0 +1,26 @@
+import { ApiRoleDTO } from '../../../api/entities/ApiRole';
+import { Tooltip } from '../Core/Tooltip';
+
+import './ApiRolesCell.scss';
+
+type ApiRolesProps = {
+  apiRoles: ApiRoleDTO[];
+  showRoleTooltip?: boolean;
+};
+function ApiRolesCell({ apiRoles, showRoleTooltip: showRole = false }: ApiRolesProps) {
```
alex-yau-ttd, line 10: can probably get rid of this renaming (guessing it was done by VS code refactoring?).
So just `showRoleTooltip = false`

alex-yau-ttd (approved):

src/web/components/ApiKeyManagement/ApiRolesCell.tsx
joshua-smee-ttd: Yeah it was. Sorry I missed that.

PR #244, opened 2023-12-20
Add Api Roles Selection to Participant Approval form 

joshua-smee-ttd:
Added a checkbox selection to choose API roles for user.
image

chrome_go7wr2j9cY
chrome_9Fss3Y1Fx1
Added tooltip option to checkbox
image
Audit log

### DIFF
--- a/src/api/services/participantsService.ts
+++ b/src/api/services/participantsService.ts
@@ -127,7 +127,16 @@ const updateParticipantAssociatedRequestTypes = async (
   await participant.$relatedQuery('types', trx).relate(participantApprovalPartial.types);
 };
 
-export const updateParticipantAndTypes = async (
+const updateParticipantAssociatedRequestApiRoles = async (
+  participant: Participant,
+  participantApprovalPartial: z.infer<typeof ParticipantApprovalPartial>,
+  trx: TransactionOrKnex
+) => {
+  await participant.$relatedQuery('apiRoles', trx).unrelate();
+  await participant.$relatedQuery('apiRoles', trx).relate(participantApprovalPartial.apiRoles);
+};
+
+export const updateParticipantAndTypesAndRoles = async (
   participant: Participant,
   participantApprovalPartial: z.infer<typeof ParticipantApprovalPartial> & {
     status: ParticipantStatus;
@@ -140,6 +149,7 @@ export const updateParticipantAndTypes = async (
       status: participantApprovalPartial.status,
     });
     await updateParticipantAssociatedRequestTypes(participant, participantApprovalPartial, trx);
+    await updateParticipantAssociatedRequestApiRoles(participant, participantApprovalPartial, trx);
   });
 };
 
--- a/src/web/components/Input/Input.scss
+++ b/src/web/components/Input/Input.scss
@@ -10,10 +10,10 @@
     border: 1px solid var(--theme-input-border);
     border-radius: 5px;
     padding: 0.5rem;
+    margin-top: 0.5rem;
   }
 
   .input-label {
-    margin-bottom: 0.5rem;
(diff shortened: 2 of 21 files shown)


src/api/services/participantsService.ts
```diff
@@ -127,7 +127,16 @@ const updateParticipantAssociatedRequestTypes = async (
   await participant.$relatedQuery('types', trx).relate(participantApprovalPartial.types);
 };
 
-export const updateParticipantAndTypes = async (
+const updateParticipantAssociatedRequestApiRoles = async (
+  participant: Participant,
+  participantApprovalPartial: z.infer<typeof ParticipantApprovalPartial>,
+  trx: TransactionOrKnex
+) => {
+  await participant.$relatedQuery('apiRoles', trx).unrelate();
```
lionell-pack-ttd, line 135: I'm still a little cautious of Objection.js and knex - have you checked the SQL this generates?

src/web/components/Input/Input.scss
```diff
@@ -10,10 +10,10 @@
     border: 1px solid var(--theme-input-border);
     border-radius: 5px;
     padding: 0.5rem;
+    margin-top: 0.5rem;
```
lionell-pack-ttd, line 13: This (and the below change) could affect a lot of spots that we've tweaked to get just right - have you had a look at a few different places to make sure the change looks right?

src/api/services/participantsService.ts
joshua-smee-ttd: image
It removes any link between the two tables for that objects.
["Doesn't delete the items. Only removes the connection. For ManyToMany relations this deletes the join row from the join table. For other relation types this sets the join columns to null."](https://vincit.github.io/objection.js/api/query-builder/mutate-methods.html#unrelate)

src/web/components/Input/Input.scss
joshua-smee-ttd: Yeah I've gone through all the components that are built on the underlining Input component and checked formatting and added it where necessary.

src/web/components/ParticipantManagement/ParticipantRequestForm.stories.tsx
```diff
@@ -63,6 +63,12 @@ ParticipantApproval.args = {
     { id: 2, typeName: 'Type 2' },
```
alex-yau-ttd, line 63: While we are here, can we please change the name of the participant so that the site dropdown starts off with a match to begin with? That way the developer does not need to check what sites are being provided in the story.
image

Perhaps `Test Participant`?
image

alex-yau-ttd: Perhaps this should be an additional story, rather than changing the existing one.

src/web/services/apiRoles.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { backendError } from '../utils/apiError';
+
+export async function GetAllApiRoles() {
+  try {
+    const result = await axios.get<ApiRoleDTO[]>(`/apiRoles`, {
+      validateStatus: (status) => status === 200,
+    });
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not get apiRoles');
```
alex-yau-ttd, line 13: nit: `Could not get API Roles`

alex-yau-ttd (approved):

src/web/components/ParticipantManagement/ParticipantRequestForm.stories.tsx
joshua-smee-ttd: Yeah I added a second so the search can still be tested

PR #245, opened 2023-12-21
Make seed data rerunnable more relliably

joshua-smee-ttd:
image
New audit table sample.

Removing participantId field from auditTrail and adding it as an key/value in eventData to remove link.

Other changes made to make seed rerunnable more reliably:
- Made the business contact cascade delete
- Added a check to remove all users of the sample participants
- Added a better where for the sample participants


src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
```diff
@@ -0,0 +1,24 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(
```
alex-yau-ttd, line 4: I think `knex.raw` is unavoidable here, but perhaps we could leverage a bit more `knex` syntax here. Perhaps something like:
```
await knex('auditTrails')
  .whereNotNull('participantId')
  .update({
    eventData: knex.raw("JSON_MODIFY(eventData, '$.participantId', ?)", [participantId])
  });
```
Not a big deal if it's not an easy change though.

src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
```diff
@@ -0,0 +1,24 @@
+import { Knex } from 'knex';
```
alex-yau-ttd, line 1: Sorry - didn't notice this on your last migration, but the file nameconvention appears to be PascalCase for migrations.

joshua-smee-ttd: Is it worth renaming both of them?

alex-yau-ttd:
Typo in the PR title :). Also worth mentioning in the PR description (perhaps the title too, actually) that the participantId is being moved to the json column, rather than just being removed altogether. In future we might dig up PRs to try figure out why we changed something, so having a good title and a meaningful and succinct description is pretty important.

src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
alex-yau-ttd: I'm a bit hesitant to rename the old one because it may recognise it as a new migration and try to run again? Perhaps not, but probably not worth the risk?

src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
joshua-smee-ttd: Managed to change one. The other one I couldn't work out how to get the where statement to work.

src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
joshua-smee-ttd: Yeah that makes sense to me. Was kind of thinking the same thing.

src/database/migrations/20231220235718_removeParticipantIdColumnFromAudit.ts
alex-yau-ttd: I think you'd need to use `knex.raw `in the `where` clause itself. Maybe the following?
```
await knex('auditTrails')
  .whereNotNull(knex.raw("JSON_VALUE(eventData, '$.participantId')"))
  .update({
    participantId: knex.raw("JSON_VALUE(eventData, '$.participantId')")
  });
```

joshua-smee-ttd: VKextfOyrO
Yeah I tried that, it throws this error. It seems to work with just where but not whereNotNull.

alex-yau-ttd: Ah right you are. What about:
```
await knex('auditTrails')
  .where(knex.raw("JSON_VALUE(eventData, '$.participantId') IS NOT NULL"))
  .update({
    participantId: knex.raw("JSON_VALUE(eventData, '$.participantId')"),
  });
```

src/database/migrations/20231220235718_RemoveParticipantIdColumnFromAuditTrail.ts
```diff
@@ -0,0 +1,26 @@
...
+      ]),
+    });
+
+  await knex.schema.alterTable('auditTrails', (table) => {
+    table.dropForeign('participantId');
+    table.dropColumn('participantId');
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
```
alex-yau-ttd, line 18: Technically this `down` method is missing the fact that `participantId` should be removed from `eventData`. Should we include that?

joshua-smee-ttd: Yeah good point. Have added a call for this.

src/database/migrations/20231220235718_RemoveParticipantIdColumnFromAuditTrail.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  await knex('auditTrails')
+    .where(knex.raw("JSON_VALUE(eventData, '$.participantId') IS NOT NULL"))
+    .update({
+      participantId: knex.raw("JSON_VALUE(eventData, '$.participantId')"),
+      eventData: knex.raw("JSON_MODIFY(eventData, '$.participantId', null)"),
+    });
+
+  await knex.schema.alterTable('businessContacts', (table) => {
```
alex-yau-ttd, line 35: Shouldn't `down` be re-instating the FK?

joshua-smee-ttd: I'm not totally sure what you mean sorry. I think it is?

alex-yau-ttd: oh sorry got confused - I see that the cascade deletion is the thing that is changing.

src/database/seeds/Participants.ts
```diff
@@ -101,8 +101,22 @@ export async function CreateParticipant(
 }
 
 export async function seed(knex: Knex): Promise<void> {
-  // Deletes ALL existing entries
-  await knex('participants').whereILike('name', '%example').del();
+  // Deletes existing entries for sampleData
+
+  await knex('users')
```
alex-yau-ttd, line 106: Maybe extract `sampleData.map((d) => d.name)` to a variable `sampleParticipantNames`?

alex-yau-ttd (approved):

PR #248, opened 2023-12-28
Allow pointing environment to integ

alex-yau-ttd:
- `.gitignore` the `.env` file so that it can be used to store integ secrets
- Add an endpoint so that the frontend can know some of the environment variables
- Use this endpoint to display a warnining when local development is pointing to integ
- Clean up `envars`
- Check for existence of `.env` file when running API
- Fix README link

**Demo**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/69cce0bd-0796-4255-a927-012f6044368f

**Running api when `.env` file does not exist**

```
> npm run api
...
Error: .env file not found. See 1Password under "ssportal-env" or see https://github.com/IABTechLab/uid2-self-serve-portal?tab=readme-ov-file#environment-variables
ERROR: "check-env-file" exited with 1.
```


src/api/envars.ts
```diff
@@ -2,8 +2,8 @@ const errorMessage = 'Unable to get envar value';
 
 // General Config
 export const SSP_APP_NAME = process.env.SSP_APP_NAME ?? 'uid2-ssportal';
-export const SSP_IS_DEVELOPMENT = process.env.NODE_ENV === 'development';
 export const SSP_IS_PRODUCTION = process.env.NODE_ENV === 'production';
+export const SSP_IS_DEVELOPMENT = !SSP_IS_PRODUCTION;
```
alex-yau-ttd, line 6: `process.env.NODE_ENV ` was always just returning `undefined` for me locally

package.json
```diff
@@ -97,7 +97,9 @@
     "start": "craco start",
     "build": "craco build",
     "test": "craco test --testPathPattern=src/web/",
-    "api": "nodemon -e \".ts\" --exec npm run babel-node-api",
+    "api": "run-p check-env-file api-watch",
```
alex-yau-ttd, line 100: Splitting these out gives a better error message: `ERROR: "check-env-file" exited with 1.` rather than `ERROR: "api" exited with 1.`

src/checkEnvFile.ts
```diff
@@ -0,0 +1,20 @@
...
+const checkEnvFile = (): void => {
+  const envFilePath = path.join(process.cwd(), '.env');
+
+  try {
+    // Check if the .env file exists
+    fs.accessSync(envFilePath, fs.constants.R_OK);
+    console.log('.env file found!');
+  } catch (err) {
+    console.error(
+      'Error: .env file not found. See 1Password under "ssportal-env" or see https://github.com/IABTechLab/uid2-self-serve-portal?tab=readme-ov-file#environment-variables'
```
ssundahlTTD, line 13: I know I personally said to mention 1password in this check, but thinking about it now I don't love putting this in something public.  Maybe just reference the readme?

src/web/services/environmentVariables.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export async function GetEnvironmentVariables() {
+  try {
+    const result = await axios.get<EnvironmentVariable>(`/envars`, {
+      validateStatus: (status) => status === 200,
+    });
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not get environment variables');
```
ssundahlTTD, line 17: Is it worth calling this out more explicitly in the UI rather than just a backendError?

src/checkEnvFile.ts
alex-yau-ttd: yeah agreed - done!

src/web/services/environmentVariables.tsx
alex-yau-ttd: I don't think it is really worth it:
* I think the endpoint is very safe, it just returns some strings which have fallback values in `envars.ts`. 
* The `.env` file would have to already be created to reach the UI, so I'm not sure what the error would actually say?

src/web/components/Core/EnvironmentBanner.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Banner } from './Banner';
+
+function EnvironmentBanner() {
+  const [environment, setEnvironment] = useState<EnvironmentVariable>({
+    baseUrl: '',
+    isDevelopment: false,
+  });
+  const throwError = useAsyncError();
+
+  useEffect(() => {
```
lionell-pack-ttd, line 15: I think this is OK, this part of the component tree should be rendered rarely so I don't think we'll end up fetching this often. Normally I'd suggest something like this should be provided via context rather than fetched, but in this case it looks fine 🙂

alex-yau-ttd: yeah I considered using context but came to the same conclusion. thanks!

lionell-pack-ttd (approved):

PR #250, opened 2024-01-04
Add ability for participants to create API Keys

joshua-smee-ttd:
Added ability for users to create API keys from the portal. See below for flow.
chrome_61WaxPiX8p

Relies on admin changes made here: https://github.com/IABTechLab/uid2-admin/pull/199

image
Audit Log from creating a Key

image
Button does not appear if participant has no api roles approved for them.


src/api/routers/participantsRouter.ts
```diff
@@ -272,6 +280,55 @@ export function createParticipantsRouter() {
...
+        participant,
+        currentUser!.id,
+        currentUser!.email,
+        AuditAction.Add,
+        keyName,
+        apiRoles,
+        traceId
+      );
+
+      if (!checkApiRoles(apiRoles, participant)) {
```
joshua-smee-ttd, line 321: Should the audit log be start before or after this check?

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
...
+    .findById(participant.id)
+    .withGraphFetched('apiRoles');
+
+  if (!participantCurr?.apiRoles) {
+    return [];
+  }
+  return participantCurr.apiRoles;
+};
+
+export const checkApiRoles = async (
```
alex-yau-ttd, line 30: this function name is a bit unclear to me: what are we checking? I know it's similar, but maybe `validateApiRoles` is clearer

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
...
+  keyRoles: string[],
+  participant: Participant
+): Promise<boolean> => {
+  const userRoles = (await getApiRoles(participant)).map((role) => role.roleName);
```
alex-yau-ttd, line 34: maybe `participantRoles` is more clear here - we may introduce user roles in the future so we want to avoid possible confusion.

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
...
+import { CreatedApiKeyDTO } from './adminServiceHelpers';
+
+export type ApiKeySecretsDTO = {
+  plaintextKey: string;
+  secret: string;
+  name: string;
+};
+
+export const createdApiKeyToApiKeySecrets = (apiKey: CreatedApiKeyDTO): ApiKeySecretsDTO => {
+  // eslint-disable-next-line camelcase
```
alex-yau-ttd, line 11: Is it just me or is this warning not needed?
image

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
+import { Participant } from '../entities/Participant';
+import { CreatedApiKeyDTO } from './adminServiceHelpers';
+
+export type ApiKeySecretsDTO = {
```
alex-yau-ttd, line 4: I think the can drop the plural and just call it `ApiKeySecretDTO`

alex-yau-ttd: hmm actually nvm

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
...
+  // eslint-disable-next-line camelcase
+  return {
+    plaintextKey: apiKey.plaintext_key,
+    secret: apiKey.authorizable.secret,
+    name: apiKey.authorizable.name,
+  };
+};
+
+export const getApiRoles = async (participant: Participant) => {
+  const participantCurr = await Participant.query()
```
alex-yau-ttd, line 20: perhaps rename to `participantWithApiRoles`? the `Curr` is probably unnecessary here, thouhgh I'm also okay with `currentParticipantWithApiRoles`

src/api/services/adminServiceClient.ts
```diff
@@ -7,6 +7,7 @@ import { SSP_ADMIN_SERVICE_BASE_URL, SSP_ADMIN_SERVICE_CLIENT_KEY } from '../env
 import { getLoggers } from '../helpers/loggingHelpers';
 import {
   ApiKeyAdmin,
+  CreatedApiKeyDTO,
```
alex-yau-ttd, line 10: I think these imports need autofix run on them
image
image

src/api/routers/participantsRouter.ts
```diff
@@ -272,6 +280,55 @@ export function createParticipantsRouter() {
...
+        return res.status(400).send('Invalid api Roles');
```
alex-yau-ttd, line 322: nit: `Invalid API Roles`

src/api/services/adminServiceHelpers.ts
```diff
@@ -93,6 +93,11 @@ export type ApiKeyDTO = {
   service_id: number;
 };
 
+export type CreatedApiKeyDTO = {
+  authorizable: ApiKeyDTO & { key_hash: string; key_salt: string; secret: string };
```
alex-yau-ttd, line 97: Is there a nice way we can do this which avoid the `& { ... }`. I know the object we get from admin is constructed weirdly, but perhaps we could just manipulate it ourselves so that it is something like:
```
export type CreatedApiKeyDTO = {
  apiKey: ApiKeyDTO
  secrets: ApiKeySecretsDTO;
};
```
Happy to be convinced otherwise.

src/api/services/apiKeyService.ts
```diff
@@ -0,0 +1,41 @@
...
+    .findById(participant.id)
+    .withGraphFetched('apiRoles');
+
+  if (!participantCurr?.apiRoles) {
```
alex-yau-ttd, line 24: nit: I think `return participantCurr?.apiRoles ?? [];` reads a bit nicer.

alex-yau-ttd: or whatever we end up calling that variable

src/api/routers/participantsRouter.ts
```diff
@@ -272,6 +280,55 @@ export function createParticipantsRouter() {
     }
   );
 
+  participantsRouter.get(
+    '/:participantId/apiRoles',
+    async (req: ParticipantRequest, res: Response) => {
+      const { participant } = req;
+      if (!participant?.siteId) {
```
alex-yau-ttd, line 287: Is this check strictly necessary?

src/api/routers/participantsRouter.ts
alex-yau-ttd: hmm what do you think? maybe we only want to audit stuff that worked? perhaps worth raising with the team

src/api/services/auditTrailService.ts
```diff
@@ -54,6 +54,38 @@ export const insertSharingAuditTrails = async (
...
+        apiRoles,
+        keyName,
+        participantId: participant.id,
+      },
+      succeeded: false,
+    };
+
+    return await AuditTrail.query().insert(manageApiKeyTrail);
+  } catch (error) {
+    const { errorLogger } = getLoggers();
```
alex-yau-ttd, line 83: The contents of this `catch` block are repeated many times in this file. Maybe now is a good time to extract it out? What do you think?

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+              </button>
+            </div>
+          </>
+        ) : (
+          <div>
+            <h1>New Key: {secrets.name}</h1>
+            <p>
+              Please copy the key and secret as they will not be saved after this window is closed.
+              Keep these secrets in a secure location and do not share them with anyone. If the
+              secrets are lost a new key will have be generated.
```
alex-yau-ttd, line 95: I assume they'll be some feedback on the copy here, but just a minor one from me: 
> If the secrets are lost a new key will have be generated.
-> 
> If the secrets are lost a new key will have to be generated.

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+              <TextInput inputName='name' label='Name' required />
+              <CheckboxInput
+                label='API Roles'
+                inputName='roles'
+                options={availableRoles.map((role) => ({
+                  optionLabel: role.externalName,
+                  value: role.roleName,
+                }))}
+                rules={{
+                  required: 'Your new API Key must have at least one role.',
```
alex-yau-ttd, line 74: Could we perhaps change the copy and add some padding? `Please select at least one API Role.`
image

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    setSecrets(undefined);
+    setOpen(false);
+  }
+
+  return (
+    <div className='key-creation-dialog'>
+      <Dialog triggerButton={triggerButton} open={open} onOpenChange={setOpen} hideClose>
+        {!secrets ? (
```
alex-yau-ttd, line 61: Would it make sense to extract some of this out rather than having a big ternary? Something like:
```
{!secrets ? <CreateApiKeyForm /> : <ApiSecrets />}
```
Just an idea to improve readability

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  setOpen(false);
+                }}
+              >
+                Cancel
+              </button>
+            </div>
+          </>
+        ) : (
+          <div>
+            <h1>New Key: {secrets.name}</h1>
```
alex-yau-ttd, line 91: maybe `<h1>{secrets.name} Secrets</h1>` would be more informative? Again, just an idea that might get overridden :)

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+                options={availableRoles.map((role) => ({
+                  optionLabel: role.externalName,
+                  value: role.roleName,
+                }))}
+                rules={{
+                  required: 'Your new API Key must have at least one role.',
+                }}
+              />
+            </Form>
+            <div className='cancel-container'>
```
alex-yau-ttd, line 78: We can get rid of this div and just use the `closeButton` prop on the `<Dialog>`:
```
      <Dialog
        triggerButton={triggerButton}
        open={open}
        onOpenChange={setOpen}
        hideClose
        closeButton='Cancel'
      >
```
image

src/web/components/Core/KeySecretReveal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      {display}
+      {value.length > MAX_SHOWN_VALUE_LENGTH && (
+        <>
+          <Popover
+            triggerButton={
+              <button
+                className='icon-button show-button'
+                aria-label={title}
+                type='button'
+                title={title}
```
alex-yau-ttd, line 38: Maybe this should be: `title={`View ${title}`}`

src/web/components/Core/KeySecretReveal.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const MAX_SHOWN_VALUE_LENGTH = 20;
+
+function ViewKeyButton({
+  value,
+  title,
+  confirmCopiedSecret,
+}: {
+  value: string;
+  title: string;
```
alex-yau-ttd, line 16: maybe a different name like `valueName` would better convey the meaning of this prop better than `title`?
e.g. for
```
<h2>Secret</h2>
<KeySecretReveal
  title='Secret'
  value={secrets.secret}
  setCopiedSecrets={setCopiedSecrets}
/>
```
It would seem like `<h2>Secret</h2>` is the title.

src/api/routers/participantsRouter.ts
joshua-smee-ttd: I would verge on having it before. That way we have a record if the api is being called incorrectly or maliciously.

src/api/services/apiKeyService.ts
joshua-smee-ttd: Yeah, that makes sense.

src/api/services/apiKeyService.ts
joshua-smee-ttd: Yeah it's not. mb, sorry.

src/api/services/apiKeyService.ts
joshua-smee-ttd: Renamed it too `participantWithApiRoles`

src/api/services/adminServiceHelpers.ts
joshua-smee-ttd: I'm not sure how that would be done. That's the type of return we get from admin. The issue is that it's adding fields inside authorizable and then adding plaintext_key as well. We could alter it once we recieve but we'd still need this object I think.
```
{
  "authorizable": {
    "key_hash": "Sk7OaQC7AUyZTeB1ubp8iE1FgL8QqCaN+pLN3lYgaRaU012qlsc4CXpXUE0CJxep8f8xtvFzjiBlad69saKyRA==",
    "key_salt": "aUFdfV+WYov+CG2Y/xdztBIbqlT+n3pfyEBfHdhwIUc=",
    "secret": "6Le7vmRaZMlPzRldVuFbCU9Nf44nL6uPnX69VglWVpU=",
    "name": "TESTING",
    "contact": "Special_UID2-C-L-999-ZY967",
    "created": 1704671901,
    "roles": [
      "MAPPER"
    ],
    "site_id": 999,
    "disabled": false,
    "service_id": 0,
    "key_id": "UID2-C-L-999-ZY967"
  },
  "plaintext_key": "UID2-C-L-999-ZY967t.aZ0rPhS347aULe2qraDOpvKYWKj6pdvUfw1jQ="
}
```

src/api/services/apiKeyService.ts
joshua-smee-ttd: Ah yeah that's nicer

src/api/routers/participantsRouter.ts
joshua-smee-ttd: Participants is not guaranteed and it having a site is not a guaranteed so I believe it is good to have.

src/api/services/auditTrailService.ts
joshua-smee-ttd: I think that would be good to do, though this PR is already pretty large so perhaps another PR can be created to do that later? 
Once note though,`insertApproveAccountAuditTrail` doesn't seem to have it?

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
joshua-smee-ttd: Sounds good!

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
joshua-smee-ttd: Have done this since saves having to define all the values. Happy to use actual components if you think that looks better.

```
{!secrets
          ? CreateApiKeyForm(onFormSubmit, availableRoles, setOpen)
          : ShowApiKeySecrets(
              secrets,
              setCopiedSecrets,
              confirmClose,
              showStatusPopup,
              setShowStatusPopup,
              statusPopupMessage
            )}
```

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
joshua-smee-ttd: I'm not sure this would work. The reason I have cancel and close separate is so I can control there behavior. This is how I'm forcing the copy/view when showing the secrets.

src/web/components/Core/KeySecretReveal.tsx
joshua-smee-ttd: Yeah that makes sense. That was a leftover from when I had the title `<h2>Secret</h2>` in this component but now that it's not this make far more sense.

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,130 @@
...
+      'Your key has been created. Please copy the credentials before closing the window.'
+    );
+    setShowStatusPopup(true);
+  };
+
+  function confirmClose(): void {
+    const values = [...copiedSecrets.values()];
+
+    if (values.filter((value) => value === false).length > 0) {
+      setStatusPopupMessage('Please copy all secrets shown before closing the page');
```
alex-yau-ttd, line 131: This message should be associated warning or info, rather than success.

src/web/components/Core/KeySecretReveal.tsx
```diff
@@ -0,0 +1,113 @@
+import { useEffect, useState } from 'react';
+
+import Popover from './Popover';
+import { StatusPopup } from './StatusPopup';
+
+import './KeySecretReveal.scss';
+
+const MAX_SHOWN_VALUE_LENGTH = 20;
+
+function ViewKeyButton({
```
alex-yau-ttd, line 10: maybe `ViewSecretButton` would be a better name here? I think references to `Key` in this file are a bit confusing

src/web/components/Core/KeySecretReveal.tsx
```diff
@@ -0,0 +1,113 @@
...
+                title={title}
+                onClick={confirmCopiedSecret}
+              >
+                View
+              </button>
+            }
+          >
+            <p>{value}</p>
+          </Popover>
+          |
```
alex-yau-ttd, line 47: I think we should remove the separator from this component so there are more loosely-coupled.

src/api/routers/participantsRouter.ts
alex-yau-ttd: yes, while a participant should have a `siteId`, I don't think we need to check it here. I think this endpoint should still succeed even if the participant does not have a `siteId`. Other endpoints in this file rely on a `siteId` because they are making calls to the admin service, but this one is just calling our db. Does that make sense?

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
alex-yau-ttd: I see, at least we could make the styling consistent with other dialogs though? e.g.
image

src/web/services/apiKeyService.tsx
```diff
@@ -0,0 +1,36 @@
...
+  participantId?: number
+): Promise<ApiKeySecretsDTO> {
+  try {
+    const result = await axios.post<ApiKeySecretsDTO>(
+      `/participants/${participantId ?? 'current'}/apiKeys/create`,
+      form
+    );
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, "Couldn't create API Key");
```
alex-yau-ttd, line 34: nit: perhaps "Could not..." for consistency?

src/web/screens/apiKeyManagement.scss
```diff
@@ -0,0 +1,5 @@
+.api-key-management-page {
+  .create-key-button {
```
alex-yau-ttd, line 2: In other examples in our code we do this styling at the div around the Dialog, rather than on the button itself. e.g.
```
<div className='add-new-item'>
  <BusinessContactDialog
    onFormSubmit={onAddEmailContact}
    triggerButton={
      <button className='small-button' type='button'>
        Add Email Contact
      </button>
    }
  />
</div>
```

where `add-new-item'` has the margin-top specified. Perhaps it is worth changing this one to make these all consistent?

src/api/routers/participantsRouter.ts
alex-yau-ttd: seems reasonable

src/api/services/auditTrailService.ts
alex-yau-ttd: yeah let's raise a separate ticket for this

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
alex-yau-ttd: I personally prefer actual components but perhaps not a big deal either way. Fine to leave as-is

src/api/routers/participantsRouter.ts
joshua-smee-ttd: Ah yeah that makes sense.

joshua-smee-ttd: Wait sorry I just had another look and it does need it. Since it's calling admin to create the key. See line 322.

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
joshua-smee-ttd: Makes sense.

src/web/services/apiKeyService.tsx
joshua-smee-ttd: Ah yeah. Makes sense.

src/web/screens/apiKeyManagement.scss
joshua-smee-ttd: Ah ok, that makes sense then.

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+            showPopupMessage={showPopupMessage}
+          />
+        )}
+
+        {showStatusPopup && (
+          <StatusPopup
+            status='Info'
+            show={showStatusPopup}
+            setShow={setShowStatusPopup}
+            message={statusPopupMessage}
```
alex-yau-ttd, line 173: this might deserve a longer display duration e.g. `displayDuration={10000}` since it's possibly not super clear to users what they need to do initially.

src/web/components/Core/Dialog.tsx
```diff
@@ -14,6 +14,7 @@ export type DialogProps = {
   onOpenChange?: (open: boolean) => void;
   fullScreen?: boolean;
   className?: string;
+  hideClose?: boolean;
```
alex-yau-ttd, line 17: maybe `hideCloseContainer`? Or something that clearly conveys which thing we are hiding (rather than `hideCloseButton` since the `closeButton` is different). Maybe some other renaming required if you think of something reasonable.

src/web/components/Core/DisplaySecret.scss
```diff
@@ -0,0 +1,11 @@
+.key-secret-reveal {
+  .show-button,
```
alex-yau-ttd, line 2: can't remember if I commented on this before, but perhaps `view-button` is a better name?

src/web/components/Core/DisplaySecret.scss
```diff
@@ -0,0 +1,11 @@
+.key-secret-reveal {
```
alex-yau-ttd, line 1: this name should probably change to `.display-secret` since we have changed the name of the component + file

src/web/components/Core/DisplaySecret.stories.tsx
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-console */
+import type { Meta, StoryObj } from '@storybook/react';
+
+import DisplaySecret from './DisplaySecret';
+
+const meta: Meta<typeof DisplaySecret> = {
+  component: DisplaySecret,
+  title: 'Shared Components/Key Secret Reveal',
```
alex-yau-ttd, line 8: Maybe change the title of this to `Shared Components/Display Secret`

src/web/components/Core/Dialog.tsx
alex-yau-ttd: or `hideCloseContainerButton`?

src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
joshua-smee-ttd: Yeah that makes sense

src/web/components/Core/Dialog.tsx
joshua-smee-ttd: `hideCloseContainerButton` I like this more since it is actually button.

alex-yau-ttd: actually just noticed that `closeButton` is a `string`, rather than a button. All uses of `closeButton` are just `'Cancel'`. Maybe we could refactor this so that your prop is indeed called `hideCloseButton` while `closeButton` is refactored to a boolean `showCancelButton` with a default text of `Cancel`?

joshua-smee-ttd: I think the operation should remain as it is currently. There are some uses where `closeButton` is empty but we still want the top right cross which is being hidden by `hideCloseContainerButton`. See `CreateAccountForm`, while it's quite weird implementation it's not using the default button. I wonder if the better solution would be to rename `closeButton` to `closeButtonText` and my new one `hideCloseButtons`, making it hide both the cross and the close button.

joshua-smee-ttd: This is what I was envisaging.

```
export type DialogProps = {
  triggerButton?: JSX.Element;
  children: ReactNode;
  title?: string;
  closeButtonText?: string;
  open?: boolean;
  onOpenChange?: (open: boolean) => void;
  fullScreen?: boolean;
  className?: string;
  hideCloseButtons?: boolean;
};
export function Dialog({
  triggerButton,
  children,
  title,
  closeButtonText,
  open,
  onOpenChange,
  fullScreen,
  className,
  hideCloseButtons = false,
}: DialogProps) {
  return (
    <RadixDialog.Root open={open} onOpenChange={onOpenChange}>
      {triggerButton && <RadixDialog.Trigger asChild>{triggerButton}</RadixDialog.Trigger>}
      <RadixDialog.Overlay className='dialog-overlay' />
      <RadixDialog.Content
        className='dialog-container'
        // To prevent dialog closed from user clicking outside of the dialog
        onInteractOutside={(fireEvent) => fireEvent.preventDefault()}
      >
        <div className={clsx({ fullScreen }, 'dialog-content', className)}>
          {title && <RadixDialog.Title className='dialog-title'>{title}</RadixDialog.Title>}
          {children}
          {closeButtonText && !hideCloseButtons && (
            <div className='dialog-close-button'>
              <RadixDialog.Close asChild>
                <button className='transparent-button' type='button'>
                  {closeButtonText}
                </button>
              </RadixDialog.Close>
            </div>
          )}
          {!hideCloseButtons && (
            <div className='dialog-close-container'>
              <RadixDialog.Close asChild>
                <button className='dialog-close-icon icon-button' aria-label='Close' type='button'>
                  <FontAwesomeIcon icon='xmark' />
                </button>
              </RadixDialog.Close>
            </div>
          )}
        </div>
      </RadixDialog.Content>
    </RadixDialog.Root>
  );
}
```
Sorry it's kinda a mess

src/web/components/Core/Dialog.spec.tsx
```diff
@@ -18,13 +18,20 @@ describe('Dialog', () => {
 
   it('does not render text close button if closeButton is undefined', () => {
     render(<WithoutCloseText />);
+    const openButton = screen.getByText('Open Dialog');
+    fireEvent.click(openButton);
+
     expect(screen.queryByText('Close')).not.toBeInTheDocument();
-    expect(screen.queryByLabelText('Close')).not.toBeInTheDocument();
+    expect(screen.getByLabelText('Close')).toBeInTheDocument();
```
alex-yau-ttd, line 25: I think we can replace this with the preferred: `expect(screen.getByRole('button', { name: 'Close' })).toBeInTheDocument();`

joshua-smee-ttd: Ah yeah sounds good

alex-yau-ttd (approved):

PR #251, opened 2024-01-04
Hide disabled API roles when selecting roles for participant

joshua-smee-ttd:
Added migration which adds disabled field to apiRoles in DB. Disabled all keys except for known good keys by default.

image
DB table
image
Resulting approval screen

### DIFF
--- a/src/api/configureApi.ts
+++ b/src/api/configureApi.ts
@@ -159,7 +159,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
   });
 
   router.get('/apiRoles', async (_req, res) => {
-    const apiRoles = await ApiRole.query();
+    const apiRoles = await ApiRole.query().where({ disabled: false });
     return res.status(200).json(apiRoles);
   });
 
--- a/src/api/configureApi.ts
+++ b/src/api/configureApi.ts
@@ -159,7 +159,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
   });
 
   router.get('/apiRoles', async (_req, res) => {
-    const apiRoles = await ApiRole.query().where({ disabled: false });
+    const apiRoles = await ApiRole.query().where('disabled', false);
     return res.status(200).json(apiRoles);
   });
 
--- a/src/api/entities/ApiRole.ts
+++ b/src/api/entities/ApiRole.ts
@@ -34,5 +34,4 @@ export const ApiRoleSchema = z.object({
   id: z.number(),
   roleName: z.string(),
   externalName: z.string(),
-  disabled: z.boolean(),
 });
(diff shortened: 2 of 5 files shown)


src/database/migrations/20240104050829_DisableOptOutRole.ts
```diff
@@ -0,0 +1,17 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('apiRoles', (table) => {
+    table.boolean('disabled').defaultTo(true).notNullable();
+  });
+
+  const allowedRoles = ['MAPPER', 'GENERATOR', 'ID_READER', 'SHARER'];
```
alex-yau-ttd, line 8: perhaps `enabledRoles` makes more sense here?

src/api/entities/ApiRole.ts
```diff
@@ -25,12 +25,14 @@ export class ApiRole extends BaseModel {
...
 }
 
-export type ApiRoleDTO = ModelObjectOpt<ApiRole>;
+export type ApiRoleDTO = Omit<ModelObjectOpt<ApiRole>, 'disabled'>;
 
 export const ApiRoleSchema = z.object({
   id: z.number(),
   roleName: z.string(),
   externalName: z.string(),
+  disabled: z.boolean(),
```
alex-yau-ttd, line 37: I don't think we need this line?

src/api/configureApi.ts
```diff
@@ -159,7 +159,7 @@ export function configureAndStartApi(useMetrics: boolean = true) {
   });
 
   router.get('/apiRoles', async (_req, res) => {
-    const apiRoles = await ApiRole.query();
+    const apiRoles = await ApiRole.query().where({ disabled: false });
```
alex-yau-ttd, line 162: can this just be `.where('disabled', false })`?

src/api/entities/ApiRole.ts
joshua-smee-ttd: Yeah we don't atm. I would think though it's better to have for later? Happy to remove if not really a point.

src/database/migrations/20240104050829_DisableOptOutRole.ts
joshua-smee-ttd: Ah yeah that does work better.

src/api/entities/ApiRole.ts
alex-yau-ttd: I think remove it if we aren't using it now. We may never use it?

joshua-smee-ttd: Sounds good!

alex-yau-ttd (approved):

PR #252, opened 2024-01-07
Edit participant API Roles

joshua-smee-ttd:
Added a edit participant dialog with the ability to update the API Roles of a participant.
image


src/web/components/ParticipantManagement/ApprovedParticipantItem.tsx
```diff
@@ -33,6 +42,18 @@ export function ApprovedParticipantItem({ participant }: ApprovedParticipantProp
...
         />
       </td>
+      <td>
+        <UpdateParticipantDialog
+          apiRoles={apiRoles}
+          onUpdateParticipant={onUpdateParticipant}
+          participant={participant}
+          triggerButton={
+            <button type='button' className='transparent-button'>
+              Edit
```
alex-yau-ttd, line 52: Let's use 
```
<FontAwesomeIcon icon='pencil' />
```
instead of `Edit`

src/api/routers/participantsRouter.ts
```diff
@@ -180,6 +180,34 @@ export function createParticipantsRouter() {
...
+  const updateParticipantParser = z.object({ apiRoles: z.array(z.number()) });
+
+  participantsRouter.put(
+    '/:participantId',
+    isApproverCheck,
+    async (req: ParticipantRequest, res: Response) => {
+      const { participant } = req;
+
+      if (!participant) {
+        return res.status(400).send('Unable to find participant');
```
alex-yau-ttd, line 192: should this be 404? On another note, a bunch of our other endpoints in here don't do this check, they just use `participant!`. Do you think we should do this check more often, or not at all?

src/api/services/participantsService.ts
```diff
@@ -127,13 +127,15 @@ const updateParticipantAssociatedRequestTypes = async (
   await participant.$relatedQuery('types', trx).relate(participantApprovalPartial.types);
 };
 
-const updateParticipantAssociatedRequestApiRoles = async (
+export const updateParticipantAssociatedRequestApiRoles = async (
```
alex-yau-ttd, line 130: This name is confusing me now, can it just be `updateParticipantApiRoles`?

src/web/components/ParticipantManagement/ApprovedParticipantItem.tsx
joshua-smee-ttd: Sounds good!
image

src/api/routers/participantsRouter.ts
joshua-smee-ttd: I don't think it's needed for most as the checkParticipantId usually guarantees a participant I believe. I think though for routes not using it, and instead using `isApprovedCheck` they should probably has the check as I don't think there's a guarantee of participant existing.

src/api/services/participantsService.ts
joshua-smee-ttd: I assume it was named this in reference to the fact that it's doing the update as part of a large db transaction. I'm not sure renaming it would be for the better as it woudl ignore the fact that a transaction has to be passed in.

src/api/routers/participantsRouter.ts
```diff
@@ -180,6 +180,34 @@ export function createParticipantsRouter() {
...
+      }
+
+      const { apiRoles } = updateParticipantParser.parse(req.body);
+
+      await Participant.transaction(async (trx) => {
```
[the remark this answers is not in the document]
joshua-smee-ttd: Yeah that makes sense, though unsure quite how to proceed from this. My best idea would be to rename `updateParticipantAndTypesAndRoles` (participantsService @ line 141) to `updateParticipantOnApproval` and then create a new function called `updateParticipant` which contains this code and can be expanded as other parts like name are added to the form.

src/api/routers/participantsRouter.ts
alex-yau-ttd: fair enough. Shouldn't this still be a 404 though?

joshua-smee-ttd: Yeah it would

src/api/services/participantsService.ts
joshua-smee-ttd: Renamed the function (`updateParticipantApiRolesUsingTransaction`) and moved the logic that was in the router to a new function in service (`updateParticipantApiRoles`).

src/web/screens/manageParticipants.tsx
```diff
@@ -33,7 +35,7 @@ function ManageParticipants() {
   };
 
   const reloader = useRevalidator();
-  const onParticipantRequestsUpdate = useCallback(() => {
+  const onParticipantUpdate = useCallback(() => {
```
alex-yau-ttd, line 38: Maybe we could give this a better name, having both `onUpdateParticipant` and `onParticipantUpdate` will be confusing. We could call this `onParticipantUpdated` or `handleParticipantUpdated`, per the other uses of `reloader.revalidate()` in our codebase.

src/web/components/ParticipantManagement/UpdateParticipantDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { CheckboxInput } from '../Input/CheckboxInput';
+
+type UpdateParticipantDialogProps = {
+  triggerButton: JSX.Element;
+  participant: ParticipantDTO;
+  onUpdateParticipant: (form: UpdateParticipantForm, participant: ParticipantDTO) => Promise<void>;
+  apiRoles: ApiRoleDTO[];
+};
+
+function UpdateParticipantDialog({
```
alex-yau-ttd, line 18: nit: I feel like it's good practice to have the props listed here in the same order as the above type?

src/web/components/ParticipantManagement/UpdateParticipantDialog.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  title: 'Manage Participants/Update Participant Dialog',
+};
+export default meta;
+
+type Story = StoryObj<typeof UpdateParticipantDialog>;
+
+export const ParticipantWithExistingRoles: Story = {
+  args: {
+    apiRoles: [
+      { id: 1, roleName: 'Role1', externalName: 'Role 1' },
```
alex-yau-ttd, line 19: Perhaps we should use actual role names rather than placeholders? It has caused some confusion in the past.

src/web/components/ParticipantManagement/UpdateParticipantDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <Dialog
+      triggerButton={triggerButton}
+      title={`Edit ${participant.name}`}
+      closeButton='Cancel'
+      open={open}
+      onOpenChange={setOpen}
+    >
+      <Form<UpdateParticipantForm>
+        onSubmit={onSubmit}
+        submitButtonText='Update Participant'
```
alex-yau-ttd, line 45: I think this should probably be 'Save Participant' for consistency with other edit dialogs e.g. TeamMember and BusinessContact

alex-yau-ttd (approved):

src/web/components/ParticipantManagement/UpdateParticipantDialog.tsx
joshua-smee-ttd: Ah yeah that makes sense.

PR #253, opened 2024-01-08
Bypass handler for envar endpoint

alex-yau-ttd:
The `envars` endpoint was being called before a user had been approved (and therefore received the keycloak role to call our API), so it was failing with `unauthorised`, which would crash the page. 

This PR adds the `envars` endpoint to the list of paths that bypass the claim check, so now it returns with `200` and will not crash the page after a user has verified their email and they are on the `account/create` screen.


joshua-smee-ttd (approved):
Looks good!

PR #258, opened 2024-01-12
Edit API Key function

joshua-smee-ttd:
Added action button for API Keys.
image

Opens dialog which allows API Key name and roles to be edited.
image

Flow
chrome_p3Ofi6r8lS

### DIFF
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -300,6 +300,8 @@ export function createParticipantsRouter() {
         return res.status(400).send('Site id is not set');
       }
 
+      // TODO Add audit log
+
       const { keyId, newName, newApiRoles } = apiKeyEditInputParser.parse(req.body);
 
       // Get key in current form and check that is valid
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -354,7 +354,8 @@ export function createParticipantsRouter() {
         return res.status(400).send('Name is invalid');
       }
 
-      if (newName !== editedKey.name) {
+      const apiKeyNameChanged = newName !== editedKey.name;
+      if (apiKeyNameChanged) {
         await renameApiKey(editedKey.contact, newName);
       }
 
@@ -363,7 +364,6 @@ export function createParticipantsRouter() {
           .map((role) => role.roleName)
           .sort()
           .join(',') !== newApiRoles.sort().join(',');
-
       if (apiKeyRolesChanged) {
         await updateApiKeyRoles(editedKey.contact, newApiRoles);
       }
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -25,7 +25,9 @@ import {
   getKeyPairsList,
   getSharingList,
   getSiteList,
+  renameApiKey,
   setSiteClientTypes,
+  updateApiKeyRoles,
 } from '../services/adminServiceClient';
 import { mapAdminApiKeysToApiKeyDTOs, SiteDTO } from '../services/adminServiceHelpers';
 import {
@@ -284,6 +286,47 @@ export function createParticipantsRouter() {
     }
   );
 
+  const apiKeyEditInputParser = z.object({
+    keyId: z.string(),
+    newName: z.string(),
+    newApiRoles: z.array(z.string()),
+  });
+
+  participantsRouter.put(
+    '/:participantId/apiKeys',
+    async (req: ParticipantRequest, res: Response) => {
+      const { participant } = req;
+      if (!participant?.siteId) {
+        return res.status(400).send('Site id is not set');
+      }
+
+      const { keyId, newName, newApiRoles } = apiKeyEditInputParser.parse(req.body);
+
+      // Get key in current form and check that is valid
+      const siteApiKeys = await getApiKeys(participant!.siteId);
+      const editedKeyAdmin = siteApiKeys.find((key) => key.key_id === keyId);
+      if (!editedKeyAdmin) {
+        return res.status(404).send('KeyId was invalid');
+      }
+      const editedKey = (await mapAdminApiKeysToApiKeyDTOs([editedKeyAdmin]))[0];
+
+      const participantRoles = await getApiRoles(participant);
+      const validRoles = editedKey.roles.concat(participantRoles);
+      if (!validateApiRoles(newApiRoles, validRoles)) {
+        return res.status(401).send('API Roles are invalid');
+      }
+
+      if (!newName) {
+        return res.status(400).send('Name is invalid');
+      }
+
+      await renameApiKey(editedKey.contact, newName);
+      await updateApiKeyRoles(editedKey.contact, newApiRoles);
+
+      return res.sendStatus(200);
+    }
+  );
+
   participantsRouter.get(
     '/:participantId/apiRoles',
     async (req: ParticipantRequest, res: Response) => {
@@ -319,7 +362,7 @@ export function createParticipantsRouter() {
         traceId
       );
 
-      if (!validateApiRoles(apiRoles, participant!)) {
+      if (!validateApiRoles(apiRoles, await getApiRoles(participant!))) {
         return res.status(400).send('Invalid API Roles');
       }
 
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -304,7 +304,7 @@ export function createParticipantsRouter() {
 
       const { keyId, newName, newApiRoles } = apiKeyEditInputParser.parse(req.body);
 
-      const editedKey = await getApiKey(participant!.siteId, keyId);
+      const editedKey = await getApiKey(participant.siteId, keyId);
       if (!editedKey) {
         return res.status(404).send('KeyId was invalid');
       }
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -301,17 +301,29 @@ export function createParticipantsRouter() {
         return res.status(400).send('Site id is not set');
       }
 
-      // TODO Add audit log
-
       const { keyId, newName, newApiRoles } = apiKeyEditInputParser.parse(req.body);
 
-      // Get key in current form and check that is valid
       const siteApiKeys = await getApiKeys(participant!.siteId);
       const editedKeyAdmin = siteApiKeys.find((key) => key.key_id === keyId);
       if (!editedKeyAdmin) {
         return res.status(404).send('KeyId was invalid');
       }
+
       const editedKey = (await mapAdminApiKeysToApiKeyDTOs([editedKeyAdmin]))[0];
+
+      const traceId = getTraceId(req);
+      const currentUser = await findUserByEmail(req.auth?.payload?.email as string);
+      const auditTrail = await insertManageApiKeyAuditTrail(
+        participant!,
+        currentUser!.id,
+        currentUser!.email,
+        AuditAction.Update,
+        editedKey.name,
+        editedKey.roles.map((role) => role.roleName),
+        traceId,
+        newName,
+        newApiRoles
+      );
 
       const participantRoles = await getApiRoles(participant);
       const validRoles = editedKey.roles.concat(participantRoles);
@@ -325,6 +337,8 @@ export function createParticipantsRouter() {
 
       await renameApiKey(editedKey.contact, newName);
       await updateApiKeyRoles(editedKey.contact, newApiRoles);
+
+      await updateAuditTrailToProceed(auditTrail.id);
 
       return res.sendStatus(200);
     }
--- a/src/api/routers/participantsRouter.ts
+++ b/src/api/routers/participantsRouter.ts
@@ -305,7 +305,6 @@ export function createParticipantsRouter() {
       const { keyId, newName, newApiRoles } = apiKeyEditInputParser.parse(req.body);
 
       const editedKey = await getApiKey(participant!.siteId, keyId);
-
       if (!editedKey) {
         return res.status(404).send('KeyId was invalid');
       }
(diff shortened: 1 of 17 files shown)


src/api/services/adminServiceClient.ts
```diff
@@ -81,11 +81,38 @@ export const getSite = async (siteId: number): Promise<SiteDTO> => {
...
   const response = await adminServiceClient.get<ApiKeyAdmin[]>(`/api/client/list/${siteId}`);
   return response.data;
 };
 
+export const getApiKeyFromAdmin = async (keyId: String): Promise<ApiKeyAdmin | undefined> => {
+  const response = await adminServiceClient.get<ApiKeyAdmin>(`/api/client/keyId`, {
+    params: { keyId },
+  });
+
+  if (response.status === 404) {
```
joshua-smee-ttd, line 94: Can't really see a way we currently nicely deal with errors from admin. My idea here was a 404 means it can't find the key and that's the same message we'd want to return. Other errors are more likely something actually broken.

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -1,11 +1,20 @@
...
 import ApiRolesCell from './ApiRolesCell';
+import KeyEditDialog, { OnApiKeyEdit } from './KeyEditDialog';
 
 type KeyItemProps = {
   apiKey: ApiKeyDTO;
+  onEdit: OnApiKeyEdit;
+  availableRoles: ApiRoleDTO[];
 };
-function KeyItem({ apiKey }: KeyItemProps) {
+function KeyItem({ apiKey: apiKeyInitial, onEdit, availableRoles }: KeyItemProps) {
```
joshua-smee-ttd, line 15: Unsure what the best practice on this naming is. My thinking here is above this component it knows it just as the api key being passed in but once it's here we want it to be initial since useState is being used.

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+      closeButtonText='Cancel'
+      open={open}
+      onOpenChange={setOpen}
+      triggerButton={triggerButton}
+      title={`Edit ${apiKey.name}`}
+    >
+      <Form<EditApiKeyFormDTO>
+        onSubmit={onSubmit}
+        defaultValues={defaultFormData}
+        submitButtonText='Edit'
```
alex-yau-ttd, line 61: I'm thinking this should be `'Save Key'` or similar, for consistency

src/api/routers/participantsRouter.ts
```diff
@@ -266,25 +269,107 @@ export function createParticipantsRouter() {
...
 
       return res.status(200).json(apiKeys);
     }
   );
 
+  const apiKeyIdParser = z.object({
+    keyId: z.string(),
+  });
+  participantsRouter.get(
+    '/:participantId/apiKeys',
```
alex-yau-ttd, line 291: would it make sense to change this to the singular `/:participantId/apiKey`?

src/api/routers/participantsRouter.ts
```diff
@@ -266,25 +269,107 @@ export function createParticipantsRouter() {
...
+    }
+  );
+
+  const apiKeyEditInputParser = z.object({
+    keyId: z.string(),
+    newName: z.string(),
+    newApiRoles: z.array(z.string()),
+  });
+  participantsRouter.put(
+    '/:participantId/apiKeys',
```
alex-yau-ttd, line 318: would it make sense to change this to the singular /:participantId/apiKey?

src/api/routers/participantsRouter.ts
```diff
@@ -299,7 +384,7 @@ export function createParticipantsRouter() {
   const apiKeyCreateInputParser = z.object({ name: z.string(), roles: z.array(z.string()) });
 
   participantsRouter.post(
-    '/:participantId/apiKeys/create',
+    '/:participantId/apiKeys',
```
alex-yau-ttd, line 387: would it make sense to change this to the singular /:participantId/apiKey?

src/api/services/adminServiceClient.ts
```diff
@@ -81,11 +81,38 @@ export const getSite = async (siteId: number): Promise<SiteDTO> => {
   return response.data;
 };
 
-export const getApiKeys = async (siteId: number): Promise<ApiKeyAdmin[]> => {
+export const getApiKeysFromAdmin = async (siteId: number): Promise<ApiKeyAdmin[]> => {
```
alex-yau-ttd, line 84: I think the `FromAdmin` might be a bit redundant. What about `getApiKeysBySite`?

src/api/services/adminServiceClient.ts
```diff
@@ -81,11 +81,38 @@ export const getSite = async (siteId: number): Promise<SiteDTO> => {
...
   const response = await adminServiceClient.get<ApiKeyAdmin[]>(`/api/client/list/${siteId}`);
   return response.data;
 };
 
+export const getApiKeyFromAdmin = async (keyId: String): Promise<ApiKeyAdmin | undefined> => {
```
alex-yau-ttd, line 89: what about `getApiKeyById`?

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+  setApiKey,
+}: KeyEditDialogProps) {
+  const [open, setOpen] = useState(false);
+
+  const onSubmit: SubmitHandler<EditApiKeyFormDTO> = async (formData) => {
+    await onEdit(formData, setApiKey);
+    setOpen(false);
+  };
+
+  const possibleRolesMap = new Map<number, ApiRoleDTO>();
```
alex-yau-ttd, line 40: I wonder if we could extract some of this away, so that it just becomes 
```
  const allowedRoles = getAllowedRoles(availableRoles, apiKey.roles);
```
which is an array so that you don't need the 
```
Array.from(possibleRolesMap...
```
in the `options` of the `CheckboxInput` below.

src/web/screens/apiKeyManagement.tsx
```diff
@@ -35,6 +40,17 @@ function ApiKeyManagement() {
     }
   };
 
+  const onKeyEdit: OnApiKeyEdit = async (form, setApiKey, participantId?) => {
+    try {
+      await EditApiKey(form, participantId);
+      setApiKey(await GetParticipantApiKey(form.keyId, participantId));
+      setStatusPopup({ message: 'Key succesfully edited', type: 'Success' });
```
alex-yau-ttd, line 47: typo ;)

alex-yau-ttd: "successfully"

src/web/screens/apiKeyManagement.tsx
```diff
@@ -35,6 +40,17 @@ function ApiKeyManagement() {
     }
   };
 
+  const onKeyEdit: OnApiKeyEdit = async (form, setApiKey, participantId?) => {
```
alex-yau-ttd, line 43: is `participantId?` necessary here? I thought it would just be `participantId`

src/web/screens/apiKeyManagement.tsx
alex-yau-ttd, line 47: Maybe something like `'Your key has been updated'` would be a nicer message here? The popup already says 'Success'

src/web/components/ApiKeyManagement/KeyEditDialog.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+    ],
+    onEdit: (form) => {
+      console.log(form);
+      return Promise.resolve();
+    },
+    triggerButton: <button type='button'>Open</button>,
+  },
+};
+
+export const KeyWithNotAllowedRoles: Story = {
```
alex-yau-ttd, line 44: The name of this story is confusing me - what is it actually testing? I think it is testing the fact that it is showing Mapper, Generator and Bidder, despite Bidder not being one of the `apiKey.roles`? Maybe we could think of a better name for this story?

src/api/services/apiKeyService.ts
```diff
@@ -25,13 +27,21 @@ export const getApiRoles = async (participant: Participant) => {
...
     if (!participantRoles.includes(role)) return false;
   }
 
   return true;
 };
+
+export const getApiKey = async (siteId: number, keyId: String): Promise<ApiKeyDTO | undefined> => {
+  const apiKeyAdmin = await getApiKeyFromAdmin(keyId);
+
+  if (apiKeyAdmin?.site_id !== siteId) return undefined;
```
alex-yau-ttd, line 44: Do you see any point in extracting this conditional to a variable, say:
```
const isApiKeyAssociatedWithSite = apiKeyAdmin?.site_id === siteId;
```
I think it would make it clearer

src/api/routers/participantsRouter.ts
joshua-smee-ttd: Yeah this makes sense to me. Will adjust the routes to match. I would think then the listing one could just be `/:participantId/apiKeys`

src/api/services/adminServiceClient.ts
joshua-smee-ttd: Yeah that makes sense to me.

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
joshua-smee-ttd: Yeah sounds good

src/web/screens/apiKeyManagement.tsx
joshua-smee-ttd: Practically at this time this has very little relevance. I could just remove the participant id completely. 
`await onEdit(formData, setApiKey);`
I don't actually use it when I call the onEdit function. My reasoning for leaving it in is if in the future we wish to use the portal to manage keys as admin there's the ability to pass through the participant id.

src/web/components/ApiKeyManagement/KeyEditDialog.stories.tsx
joshua-smee-ttd: Have renamed it to this: `KeyWithRolesParticipantIsntAllowed`. The tests aim is to show the behaviour when a key contains roles which the participant isn't actually allowed.

src/api/services/apiKeyService.ts
joshua-smee-ttd: Ah yeah that makes sense!

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
```diff
@@ -23,6 +23,15 @@ type KeyEditDialogProps = {
   setApiKey: React.Dispatch<React.SetStateAction<ApiKeyDTO>>;
 };
 
+function getAllowedRoles(apiRoleAllowedLists: ApiRoleDTO[][]): ApiRoleDTO[] {
```
alex-yau-ttd, line 26: Is it worth adding some basic unit tests for this?

src/api/routers/participantsRouter.ts
```diff
@@ -278,13 +277,99 @@ export function createParticipantsRouter() {
...
+        return res.status(400).send('Name is invalid');
+      }
+
+      if (newName !== editedKey.name) await renameApiKey(editedKey.contact, newName);
+
+      if (
+        editedKey.roles
+          .map((role) => role.roleName)
+          .sort()
+          .join(',') !== newApiRoles.sort().join(',')
```
alex-yau-ttd, line 363: It's probably worth extracting this condition to a variable

src/api/routers/participantsRouter.ts
```diff
@@ -278,13 +277,99 @@ export function createParticipantsRouter() {
...
+      const validRoles = editedKey.roles.concat(participantRoles);
+      if (!validateApiRoles(newApiRoles, validRoles)) {
+        return res.status(401).send('API Roles are invalid');
+      }
+
+      if (!newName) {
+        return res.status(400).send('Name is invalid');
+      }
+
+      if (newName !== editedKey.name) await renameApiKey(editedKey.contact, newName);
```
alex-yau-ttd, line 357: Perhaps our formatting should be consistent with the above and use 
```
if (newName !== editedKey.name) {
  await renameApiKey(editedKey.contact, newName);
}
```

src/web/screens/apiKeyManagement.tsx
alex-yau-ttd: I'm thinking we should just get rid of `participantId` and change `EditApiKey` to not accept it, and instead just rely on the `'current'` participant. What do you think?

joshua-smee-ttd: I think getting rid of it from `OnApiKeyEdit` is good. Unsure if removing it from `EditApiKey` is good. It maintains the same pattern as all our other endpoints atm and it would help if in the future we wish to have admin edit any participants api keys for example.

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
joshua-smee-ttd: Sounds good. I'll put one in.

src/web/components/ApiKeyManagement/KeyEditDialog.spec.tsx
```diff
@@ -0,0 +1,88 @@
+import { getAllowedRoles } from './KeyEditDialog';
+
+describe('test GetAllowedRoles function', () => {
```
alex-yau-ttd, line 3: I think it's a bit misleading to have this file name and extension because it implies it is testing the dialog itself but it's really just testing one helper method.

Perhaps it would be better to extract `getAllowedRoles` to a `KeyEditDialogHelpers.ts` and then have this file be `KeyEditDialogHelpers.spec.ts` with the describe line being:
```
describe('#getAllowedRoles', () => {
```
This uses the convention we have in other files e.g. `siteConvertingHelpers.spec.ts` and `bulkAddPermissionsHelpers.spec.ts`

src/web/screens/apiKeyManagement.tsx
alex-yau-ttd: Let's get rid of it from `OnApiKeyEdit`. I don't feel too strongly either way about removing it from `EditApiKey` - there's a case that if it isn't used today or in the immediate future then we should remove it, but also you are right in that matches the pattern we usually use for web service endpoints.

src/web/components/ApiKeyManagement/KeyEditDialog.spec.tsx
joshua-smee-ttd: Ah yeah that makes sense.

alex-yau-ttd (approved):

PR #259, opened 2024-01-16
Added knex down command

joshua-smee-ttd:
Added the knex down command, which only rolls back the most recent migration instead of the last batch. Should be used in most circumstances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ We use knex for migrations and seed data. Once your Docker environment is runnin
 - `npm run knex:migrate:latest`
   - Update your schema to the latest.
 - `npm run knex:migrate:down`
-  - Rollback the most single recent migration to your schema.
+  - Undo the last migration that was run.
 - `npm run knex:migrate:rollback`
   - Rollback the most recent batch of migrations to your schema. This command can be dangerous as it may rollback multiple migrations, in most cases `down` should be used instead.
 - `npm run knex:seed:run`
--- a/README.md
+++ b/README.md
@@ -47,8 +47,10 @@ We use knex for migrations and seed data. Once your Docker environment is runnin
 
 - `npm run knex:migrate:latest`
   - Update your schema to the latest.
+- `npm run knex:migrate:down`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -159,7 +161,8 @@ This action should be triggered after adding or updating any templates and befor
 > **Note:** If you only need to update an existing template, you can simply modify the corresponding `.hbs` file and trigger the **Sync Handlebars Template with SendGrid** action. There is no need to redeploy or rebuild Docker images in this case.
 
 ## Environment Variables
-We do not have a committed `.env` file as it goes against best practice, per https://www.npmjs.com/package/dotenv. However, we do provide a sample `.env.sample` file that can be copied into a `.env` file to get development up and running.
```
joshua-smee-ttd, line 162: This is autoformatting adding this line, not relevant to this but is good to have.

README.md
```diff
@@ -47,8 +47,10 @@ We use knex for migrations and seed data. Once your Docker environment is runnin
 
 - `npm run knex:migrate:latest`
   - Update your schema to the latest.
+- `npm run knex:migrate:down`
+  - Rollback the most single recent migration to your schema.
```
alex-yau-ttd, line 51: I think this should be `Rollback the most recent single migration to your schema.`

joshua-smee-ttd: I'm unsure if it should be. I think that more says that your rolling back the most recent single migration you've run. I'm unsure quite how to explain it but it just sounds not quite right to me.

alex-yau-ttd: hmm I sorta see what you are saying but I still think "Rollback the most single recent migration to your schema." could be improved - because it sounds like it is the "most single" which doesn't make sense. 

We should probably just use what the [knex docs](https://knexjs.org/guide/migrations.html) says, so I'd suggest "Undo the last migration that was run".

joshua-smee-ttd: Ah yeah that makes sense!

alex-yau-ttd (approved):

PR #264, opened 2024-01-18
Fix and add tests to validateApiRoles function

joshua-smee-ttd:


src/api/tests/apiKeyService.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { ApiRoleDTO } from '../entities/ApiRole';
+import { validateApiRoles } from '../services/apiKeyService';
+
+describe('#validateApiRoles', () => {
+  test('1 Valid Roles', () => {
```
alex-yau-ttd, line 5: I wonder if we can give this test a more informative name. What do you think about "should return true when the role is valid". And then similar for the other tests.

alex-yau-ttd (approved):

src/api/tests/apiKeyService.spec.ts
joshua-smee-ttd: Yeah that makes sense. Have renamed them.

src/api/tests/apiKeyService.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { ApiRoleDTO } from '../entities/ApiRole';
+import { validateApiRoles } from '../services/apiKeyService';
+
+describe('Returns true if role(s) are valid', () => {
```
alex-yau-ttd, line 4: This was better before: `#validateApiRoles`

alex-yau-ttd (changes requested):

alex-yau-ttd (approved):

PR #267, opened 2024-01-21
Key Management UI tests

joshua-smee-ttd:
Added testing for Edit and Create dialogs for API Key Management

### DIFF
--- a/src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
+++ b/src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
@@ -1,8 +1,7 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
-import { ApiRole, ApiRoleDTO } from '../../../api/entities/ApiRole';
-import { ApiKeySecretsDTO } from '../../../api/services/apiKeyService';
+import { ApiRoleDTO } from '../../../api/entities/ApiRole';
 import KeyCreationDialog from './KeyCreationDialog';
 
 const writeText = jest.fn();
@@ -142,7 +141,7 @@ describe('Key creation dialog', () => {
   });
 
   test.each(apiRoleTests)('Should display key after form submitted', async (apiRoles) => {
-    const onSubmitMock = await loadComponent(apiRoles);
+    await loadComponent(apiRoles);
 
     await enterApiName('key_name');
     await clickApiRole(apiRoles[0]);
--- a/src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
+++ b/src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
@@ -1,12 +1,9 @@
-import { composeStories } from '@storybook/testing-react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { close } from 'fs';
+import userEvent from '@testing-library/user-event';
 
+import { ApiRole, ApiRoleDTO } from '../../../api/entities/ApiRole';
 import { ApiKeySecretsDTO } from '../../../api/services/apiKeyService';
 import KeyCreationDialog from './KeyCreationDialog';
-import * as stories from './KeyCreationDialog.stories';
-
-const { MultipleRoles, OneRole } = composeStories(stories);
 
 const writeText = jest.fn();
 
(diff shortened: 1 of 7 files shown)


src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
```diff
@@ -0,0 +1,242 @@
...
+    });
+
+    await waitFor(() => {
+      expect(screen.getByText('ABCD')).toBeInTheDocument();
+    });
+
+    expect(screen.getByText('1234')).toBeInTheDocument();
+  });
+
+  it('should work with multiple allowed api role', async () => {
```
lionell-pack-ttd, line 98: Some of these tests have a lot of duplication. Jest has a few ways to write data-driven tests - check out https://jestjs.io/docs/api#testeachtablename-fn-timeout first and see if that improves things.

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { composeStories } from '@storybook/testing-react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+
+import KeyEditDialog, {  } from './KeyEditDialog';
+import * as stories from './KeyEditDialog.stories';
+
+const { KeyWithRolesParticipantIsntAllowed, MultipleRoles } = composeStories(stories);
+
+describe('Key edit dialog', () => {
+  it('has the correct title', () => {
```
alex-yau-ttd, line 11: I do wonder how much value this test adds - it seems more like a basic "did it successfully render" test which might be better asserted by checking for the presence of a checkbox or a save/cancel button? Then it won't be tightly coupled with the title which doesn't really influence the behaviour and may be more subject to change? What do you think?

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+  it('should prefill name input', () => {
+    render(<MultipleRoles />);
+
+    const openButton = screen.getByText('Open');
+    fireEvent.click(openButton);
+
+    expect(screen.getByRole('textbox', { name: 'newName' })).toHaveValue('ApiKey');
+  });
+
+  it('should select the current rows', () => {
```
alex-yau-ttd, line 29: "should select the current rows" -> "should pre-select the current API roles"?

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+    for (const role of ['ID_READER', 'SHARER']) {
+      expect(screen.getByDisplayValue(role)).not.toBeChecked();
+    }
+
+    for (const role of ['MAPPER', 'GENERATOR']) {
+      expect(screen.getByDisplayValue(role)).toBeChecked();
+    }
+  });
+
+  it('should show both the keys and allowed roles', () => {
```
alex-yau-ttd, line 44: "should show both the keys and allowed roles" -> "should show both the key's roles and the participant's allowed API roles"?

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+    for (const role of ['MAPPER', 'GENERATOR']) {
+      expect(screen.getByDisplayValue(role)).toBeChecked();
+    }
+
+    for (const role of ['ID_READER']) {
+      expect(screen.getByDisplayValue(role)).not.toBeChecked();
+    }
+  });
+
+  it('should return the correct values with multiple roles', async () => {
```
alex-yau-ttd, line 59: this test only asserts that the endpoint is called with the expected data, perhaps the test name should reflect this? Since we cannot guarantee that the endpoint will return the correct data. Perhaps "should call API endpoint with expected data"?

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+      {
+        keyId: 'F4lfa.fdas',
+        newName: 'ApiKey',
+        newApiRoles: ['MAPPER', 'ID_READER'],
+      },
+      setApiKeyMock
+    );
+  });
+
+  it('should return the correct values with one role', async () => {
```
alex-yau-ttd, line 116: similar comment to above regarding test name

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+      {
+        keyId: 'F4lfa.fdas',
+        newName: 'ApiKey',
+        newApiRoles: ['MAPPER'],
+      },
+      setApiKeyMock
+    );
+  });
+
+  it('should return the key be renamed', async () => {
```
alex-yau-ttd, line 166: similar comment to above

src/web/components/ApiKeyManagement/keyEditDialog.spec.tsx
```diff
@@ -0,0 +1,267 @@
...
+        keyId: 'F4lfa.fdas',
+        newName: 'ApiKey Rename',
+        newApiRoles: ['MAPPER'],
+      },
+      setApiKeyMock
+    );
+  });
+});
+
+it('should show an error if no roles given', async () => {
```
alex-yau-ttd, line 220: Is there a reason this test is outside the big `describe` block?

alex-yau-ttd, line 220: Maybe best to clarify that this behaviour is only on-submit in the test name? e.g. `should display error on-submission when no roles selected`

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
 
+import { Form } from '../Core/Form';
+import { CheckboxInput } from './CheckboxInput';
 import * as stories from './CheckboxInput.stories';
 
-const { WithValidation } = composeStories(stories);
+const { WithValidation, OneOption } = composeStories(stories);
 
 describe('CheckboxInput', () => {
+  it('should work with one option', async () => {
```
alex-yau-ttd, line 12: Similar to above, can you please expand on what "work" means? Same applies to all other tests

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+      </Form>
+    );
+
+    userEvent.click(screen.getByRole('checkbox', { name: 'Option 1' }));
+
+    const submitButton = screen.getByRole('button', { name: 'Submit' });
+    userEvent.click(submitButton);
+
+    await waitFor(() => {
+      expect(onSubmitMock.mock.calls).toEqual([[{ default: ['option1'] }]]);
```
alex-yau-ttd, line 31: Can this use `.toHaveBeenCalledWith` instead?

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+    );
+
+    userEvent.click(screen.getByRole('checkbox', { name: 'Option 1' }));
+    userEvent.click(screen.getByRole('checkbox', { name: 'Option 2' }));
+
+    const submitButton = screen.getByRole('button', { name: 'Submit' });
+    userEvent.click(submitButton);
+
+    await waitFor(() => {
+      expect(onSubmitMock.mock.calls).toEqual([[{ default: ['option1', 'option2'] }]]);
```
alex-yau-ttd, line 59: Can this use `.toHaveBeenCalledWith` instead?

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+          ]}
+        />
+      </Form>
+    );
+
+    const submitButton = screen.getByRole('button', { name: 'Submit' });
+    userEvent.click(submitButton);
+
+    await waitFor(() => {
+      expect(onSubmitMock.mock.calls).toEqual([[{ default: [] }]]);
```
alex-yau-ttd, line 84: Can this use `.toHaveBeenCalledWith` instead?

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+    await waitFor(() => {
+      const option2 = screen.getByRole('checkbox', { name: 'Option 2' });
+      expect(option2).toBeChecked();
+    });
+
+    const submitButton = screen.getByRole('button', { name: 'Submit' });
+    userEvent.click(submitButton);
+
+    await waitFor(() => {
+      expect(onSubmitMock.mock.calls).toEqual([[{ default: ['option1', 'option2'] }]]);
```
alex-yau-ttd, line 119: Can this use `.toHaveBeenCalledWith` instead?

src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
```diff
@@ -0,0 +1,242 @@
...
+    writeText,
+  },
+});
+
+describe('Key creation dialog', () => {
+  it('Should show one apiRole', () => {
+    render(<OneRole />);
+
+    const openButton = screen.getByText('Open');
+    fireEvent.click(openButton);
```
alex-yau-ttd, line 24: I think we should pretty much be exclusively using `userEvent` instead of `fireEvent`, see https://testing-library.com/docs/user-event/intro/#:~:text=fireEvent%20dispatches%20DOM%20events%2C%20whereas,additional%20checks%20along%20the%20way.

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+
+    const submitButton = screen.getByRole('button', { name: 'Submit' });
+    userEvent.click(submitButton);
+
+    await waitFor(() => {
+      expect(onSubmitMock.mock.calls).toEqual([[{ default: [] }]]);
+    });
+  });
+
+  it('should show default values as selected', async () => {
```
alex-yau-ttd, line 88: perhaps "should automatically select default values"?

src/web/components/Input/CheckboxInput.spec.tsx
```diff
@@ -2,11 +2,124 @@ import { composeStories } from '@storybook/testing-react';
...
+            { optionLabel: 'Option 2', optionToolTip: 'Option2', value: 'option2' },
+            { optionLabel: 'Option 3', optionToolTip: 'Option3', value: 'option3' },
+          ]}
+        />
+      </Form>
+    );
+
+    await waitFor(() => {
+      const option1 = screen.getByRole('checkbox', { name: 'Option 1' });
+      expect(option1).toBeChecked();
```
alex-yau-ttd, line 107: I think these expect statements should be outside the `waitFor` block.

src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
```diff
@@ -0,0 +1,242 @@
...
+const writeText = jest.fn();
+
+Object.assign(navigator, {
+  clipboard: {
+    writeText,
+  },
+});
+
+describe('Key creation dialog', () => {
+  it('Should show one apiRole', () => {
```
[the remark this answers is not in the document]
joshua-smee-ttd: Yeah that makes sense.

src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
```diff
@@ -0,0 +1,242 @@
...
+    render(<OneRole />);
+
+    const openButton = screen.getByText('Open');
+    fireEvent.click(openButton);
+
+    expect(screen.getByDisplayValue('MAPPER')).not.toBeChecked();
```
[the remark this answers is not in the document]
joshua-smee-ttd: That's something we could do though I'm not totally sure about the logic of it. I guess we could have a list of all of them and compare which ones were selected. My thought would be that the risk of adding extra info is very low but happy to add if you think it's worth doing.

src/web/components/ApiKeyManagement/keyCreationDialog.spec.tsx
```diff
@@ -0,0 +1,242 @@
...
+    fireEvent.change(nameInput, { target: { value: `ApiKey` } });
+
+    const createButton = screen.getByRole('button', { name: 'Create API Key' });
+    fireEvent.click(createButton);
+    await waitFor(() => {
+      expect(screen.getByText('Please select at least one API Role.')).toBeInTheDocument();
+    });
+  });
+
+  it('should work with one allowed api role', async () => {
```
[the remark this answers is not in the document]
joshua-smee-ttd: Yeah that makes sense. I'm reging the naming but using this name

lionell-pack-ttd (approved):

PR #268, opened 2024-01-22
UID2-2674 Add publish workflow for ssportal

cYKatherine:
Successful run: https://github.com/IABTechLab/uid2-self-serve-portal/actions/runs/7608204622


.github/workflows/release-docker-image.yaml
```diff
@@ -0,0 +1,56 @@
+name: Release Self Serve Portal Image
+run-name: ${{ inputs.release_type == 'Snapshot' && 'Publish Pre-release' || format('Release {0}', inputs.release_type)}} Self Serve Portal Image by @${{ github.actor }}  
+on:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        type: choice
+        description: 'The type of release'
```
gmsdelmundo, line 8: No need for quotes, and do we need `required: true`?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uid2-ssportal",
-  "version": "0.10.1",
+  "version": "0.10.1-SNAPSHOT",
```
gmsdelmundo, line 3: Should `-SNAPSHOT` be here?

.github/workflows/release-docker-image.yaml
cYKatherine: Because it's a type: choice, I don't think we need to add required: true as it's a dropdown menu

PR #275, opened 2024-01-24
Adding testing using MSW and a SQLite DB

joshua-smee-ttd:
Added testing of API endpoint mocking admin calls and using a fake DB.

We added [MSW](https://mswjs.io/) which allows us to intercept calls to Admin and mock our own response for testing.

We integrated Knex with an SQLite DB stored in memory. Using this we were able to mount and run all our migrations on a blank database that we could then use during the test.

We created functions to create test Participants and Users which use [Faker](https://fakerjs.dev/) to provide meaningful test data. In the future this could be expanded into a full blown model factory to create the different items wanted for testing.

We refactored how the participant router was organized. `participantsRouter.ts` was a very large file. We aim to make this more manageable by moving the functions out of the main router file and into helper functions. We can then also test the functions in each of these files which means we don't have to start the server.

Command to run new tests: `npx jest --testTimeout=5000 --config=./src/api/jest.config.ts --runInBand --watch  --testPathPattern=src/api/routers/participants`


src/api/routers/helperFunctions.ts
```diff
@@ -0,0 +1,76 @@
+import { faker } from '@faker-js/faker';
```
alex-yau-ttd, line 1: TODO: rename this file

src/api/routers/participantsRouter.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+import { TestConfigure } from '../../database/TestSelfServeDatabase';
+import { Participant, ParticipantStatus } from '../entities/Participant';
+import { User } from '../entities/User';
+import { SSP_ADMIN_SERVICE_BASE_URL } from '../envars';
+import { KeyPairDTO } from '../services/adminServiceHelpers';
+import { ParticipantRequest } from '../services/participantsService';
+import { createParticipant, createResponseObject, createUser } from './helperFunctions';
+import { getParticipantKeyPairs, getParticipantUsers } from './participantsRouter';
+
+const oneKeyPair: KeyPairDTO[] = [
```
alex-yau-ttd, line 14: In the process of splitting the router and test into separate files, perhaps the mock data could also be extracted into a separate file. We could possibly also use faker.js here

test.db
alex-yau-ttd, line 1: I guess we shouldn't check this one in

src/database/migrations/20230308044836_UpdateUserSchemeAndAddStatusToParticipant.ts
```diff
@@ -4,7 +4,7 @@ export async function up(knex: Knex): Promise<void> {
   await knex.schema.alterTable('users', (table) => {
```
alex-yau-ttd, line 4: Let's extract the migration changes to a separate PR

package.json
```diff
@@ -102,7 +105,7 @@
     "api-watch": "nodemon -e \".ts\" --exec npm run babel-node-api",
     "babel-node-api": "babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
     "build-api": "npx tsc --build ./tsconfig-api.json",
-    "test-api": "npx jest --testTimeout=5000 --config=./src/api/jest.config.ts",
+    "test-api": "npx jest --testTimeout=5000 --config=./src/api/jest.config.ts --runInBand",
```
joshua-smee-ttd, line 108: MSW is not happy running unless using `--runInBand`. Runs tests serially instead of worker pool of child processes.

src/api/jest.polyfills.ts
```diff
@@ -0,0 +1,30 @@
+// jest.polyfills.js
+/**
+ * @note The block below contains polyfills for Node.js globals
+ * required for Jest to function when running JSDOM tests.
+ * These HAVE to be require's and HAVE to be in this exact
```
alex-yau-ttd, line 5: does this note still apply?

joshua-smee-ttd: Uhmmm good question. I somehow missed it. It seems to be working without it so unsure? I'd say it doesn't?

lionell-pack-ttd (approved):

PR #280, opened 2024-01-25
Rename SiteDTO to SiteAdmin

joshua-smee-ttd:
SiteDTO is only sent to the portal in one place. I am changing the option and as such aim to rename this to a more relevant name


lionell-pack-ttd:
`SiteAdmin` doesn't seem right as a name - reading it naturally, I'd assume it represented a role or user that is a site administrator.

I think `AdminSite` or `AdminSiteDTO` would make more sense.

PR #310, opened 2024-01-31
Restrict API Role approval endpoint

joshua-smee-ttd:
Added check to endpoint so only adds  enabled api roles when a participant is approved/edited.

Also removed restriction on no API Roles as this provides a way to stop API key management by a participant

### DIFF
--- a/src/api/services/participantsService.ts
+++ b/src/api/services/participantsService.ts
@@ -2,6 +2,7 @@ import { NextFunction, Request, Response } from 'express';
 import { TransactionOrKnex } from 'objection';
 import { z } from 'zod';
 
+import { ApiRole } from '../entities/ApiRole';
 import {
   Participant,
   ParticipantApprovalPartial,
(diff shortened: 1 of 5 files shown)


src/api/services/participantsService.ts
```diff
@@ -129,13 +130,23 @@ const updateParticipantRequestTypesWithTransaction = async (
...
+
+  const apiRoles = await ApiRole.query()
+    .whereIn(
+      'id',
+      apiRoleIds.map((role) => role.id)
+    )
+    .where('disabled', false);
+
+  if (apiRoles.length > 0) {
+    await participant.$relatedQuery('apiRoles', trx).relate(apiRoles);
```
joshua-smee-ttd, line 148: Found an error occurs if the  list is empty, hence check.

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -29,28 +29,30 @@ function KeyItem({ apiKey: apiKeyInitial, onEdit, onDisable, availableRoles }: K
...
-          apiKey={apiKey}
-          onDisable={onDisable}
-          triggerButton={
-            <button type='button' className='transparent-button'>
-              <FontAwesomeIcon icon='trash-can' />
-            </button>
-          }
-        />
-      </td>
+      {availableRoles.length > 0 && (
```
joshua-smee-ttd, line 32: Hide all management tools if no roles given.

PR #315, opened 2024-02-06
Fix test-api running

joshua-smee-ttd:
Fixed test that was failing for api tests


package.json
```diff
@@ -102,7 +102,7 @@
     "api-watch": "nodemon -e \".ts\" --exec npm run babel-node-api",
     "babel-node-api": "babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
     "build-api": "npx tsc --build ./tsconfig-api.json",
-    "test-api": "npx jest --testTimeout=5000 --config=./src/api/jest.config.ts",
+    "test-api": "npx jest --testTimeout=5000 --forceExit --runInBand --config=./src/api/jest.config.ts",
```
alex-yau-ttd, line 105: What does using [forceExit ](https://jestjs.io/docs/cli#--forceexit)actually achieve here? The docs suggest this is an escape-hatch but that resources should be torn down after each test instead so that Jest can shut down cleanly.

joshua-smee-ttd: It allows the test to finish. Lionell and I couldn't find the underlying issue, may be related to supertest, and there was already a comment in our code recommending this.

src/api/tests/queryMocks.ts
```diff
@@ -5,24 +5,33 @@ import { BusinessContact, ContactType } from '../entities/BusinessContact';
...
-                typeName: 'DSP',
-              },
-            ],
-            ...participant,
-          }
+type PartialParticipantOrNull = Partial<Participant> | null;
+export const mockParticipant = (
+  participant: PartialParticipantOrNull | PartialParticipantOrNull[] = {}
+) => {
+  const participants = Array.isArray(participant) ? participant : [participant];
```
alex-yau-ttd, line 12: This makes me wonder if this should just be two different methods, `mockParticipant` and `mockParticipants`. Does that make any sense or no?

joshua-smee-ttd: This is the pattern we use for the rest of the functions in this file.

alex-yau-ttd: fair, ignore me :D

alex-yau-ttd (approved):

PR #322, opened 2024-02-13
Changing copy to say API Permissions

joshua-smee-ttd:
As noted, we have used API permissions instead of roles in documentation. All mentions of API Roles in the copy ahve therefore been changed to permissions.


src/web/components/ApiKeyManagement/KeyEditDialog.tsx
```diff
@@ -82,8 +82,8 @@ function KeyEditDialog({
                   <li key={`Role-${role.id}`}>{role.externalName}</li>
                 ))}
               </ul>
-              If you remove any of these roles from the key, you will not be able to undo this
-              action.
+              If you remove any of these permission(s) from the key, you will not be able to undo
```
alex-yau-ttd, line 85: I think "permissions" makes more sense than "permission(s)" here.

alex-yau-ttd (approved):

src/web/components/ApiKeyManagement/KeyEditDialog.tsx
joshua-smee-ttd: Ah yeah that's fair

PR #324, opened 2024-02-14
UID2 Replace cancel buttons with closeButtonText prop in Dialog

ashleysmithTTD:
Using the `closeButtonText` prop of `Dialog` for the cancel button rather than an explicit <button> in several components


ashleysmithTTD:
Test Plan:

Test each of the components and make sure the cancel button still works.  

1. Email Contacts -> FIll out fields and add email contact -> Delete email contact -> Hit cancel
2. Team Member -> Try to delete yourself as team member -> Hit cancel
3. Sharing Permissions -> Add permissions - Individual -> Select participant -> Add permissions -> Hit cancel
4. Sharing Permissions -> Add permissions - Individual -> Select participant -> Add permissions -> Save -> Select participant under Your Sharing Permissions table -> Delete permissions -> Cancel  
5. (CstgDomain only in storybook) -> CstgDomainsTable -> Delete a top-level domain -> Hit cancel

ssundahlTTD (approved):

PR #337, opened 2024-02-25
UID2-2738 retrieve access token and use to access admin

Ian-Nara:
I have manually tested this against my admin instance running in AWS uid2-test account. To verify the refresh implementation, I reduced the access token lifetime in Okta and logged the request in the interceptor. I can confirm that it will use the same access token until it is within 10 minutes of expiration, at which point it will get a new one. I have not written any tests as of now. I think the change is simple enough to verify manually.


README.md
```diff
@@ -256,6 +256,6 @@ The following steps describe the minimal steps required to successfully log in t
 
 1. Run `uid2-admin` locally by following the README: https://github.com/IABTechLab/uid2-admin
 1. Ensure that the site ids of your participants exist in admin. That goes for the current participant you are logged in to, as well as the participants you are interacting (e.g. sharing) with. You can check the existing ids by looking at `sites.json` in `uid2-admin` or by going to http://localhost:8089/adm/site.html and hitting `List Sites`, given the service is running locally.
-1. Ensure that you have a key in `admins.json` in `uid2-admin` with the `SHARING_PORTAL` role, and that key is in your `.env` as `SSP_ADMIN_SERVICE_CLIENT_KEY`.
-   - You should be able to use an existing key in `admins.json`, rather than having to create a new one.
+1. Ensure that you have set the Okta client secret in your `.env` as `SSP_OKTA_CLIENT_SECRET`.
+   - You can find the key for local testing in 1Password under "Okta localhost deployment.
```
lionell-pack-ttd, line 260: How would this work for an open source contributor? I don't think we want to tie local dev to our main Okta instance.

src/api/services/adminServiceClient.ts
```diff
@@ -15,16 +21,53 @@ import {
...
+      'Content-Type': 'application/x-www-form-urlencoded',
+      Authorization: `Basic ${Buffer.from(
+        `${SSP_OKTA_CLIENT_ID}:${SSP_OKTA_CLIENT_SECRET}`,
+        'binary'
+      ).toString('base64')}`,
+    },
+  },
+});
+
+async function getAccessToken() {
```
lionell-pack-ttd, line 40: Is there a reason not to use https://github.com/okta/okta-auth-js for this? It provides some nice features like active renewal (it'll keep the token fresh and not introduce extra latency waiting for a new one), and if we end up letting people sign in to the portal using Okta (which seems likely at some point) we would need it in the front-end anyway.

Ian-Nara: The js SDK is for [OpenID Connect & OAuth 2.0 API](https://developer.okta.com/docs/api/resources/oidc) type workflows for signing in end users. It does not support the [Client Credentials](https://developer.okta.com/docs/guides/implement-grant-type/clientcreds/main/) workflow.

README.md
Ian-Nara: They can set is_auth_disabled = true in their admin config. Updated the readme to note this

src/api/services/adminServiceClient.ts
Ian-Nara: The SDK supports the Authorization Code flow, PKCE OAuth 2.0 flow, and Implicit OAuth 2.0 flow, but not the Client Credentials flow, as it is intended for authenticating end users, not machines.

lionell-pack-ttd (approved):

lionell-pack-ttd:
Approved - as long as the new values are configured for integ and prod environments.

PR #339, opened 2024-02-26
New Participant Creation Dialog

ssundahlTTD:
Add new participant creation dialog.  This will allow admin users to create new participants and corresponding users without the back-and-forth of requests and approvals.

Dialog is launched from the participant management tab.  Support for new sites will be added in a future PR.

image


src/api/routers/participants/participantsRouter.ts
```diff
@@ -217,6 +220,71 @@ export function createParticipantsRouter() {
     }
   );
 
+  participantsRouter.put('/', isApproverCheck, async (req, res) => {
+    const participantRequest = AdminParticipantCreationPartial.parse(req.body);
```
alex-yau-ttd, line 224: I wonder if the logic of this function could be put into a new file `participantsApproval.ts` or similar - or at least `participantsService`. I'd be nice to make the router as thin as possible, and it also makes it easier to test using the pattern established in https://github.com/IABTechLab/uid2-self-serve-portal/pull/275

src/api/services/adminServiceHelpers.ts
```diff
@@ -57,11 +57,23 @@ export const AllowedSiteRoles: Record<string, AvailableClientRole[]> = {
   [ParticipantTypeData.Publisher.typeName]: ['GENERATOR', 'SHARER'],
 };
 
+export const AllowedSiteRolesById: Record<number, number[]> = {
+  1: [3],
```
alex-yau-ttd, line 61: Maybe a comment in code to explain these? 
I guess it means
```
  1: [3], // DSP: ID_READER
  2: [1], // Advertiser: MAPPER
  3: [1], // Data Provider: MAPPER
  4: [2], // Publisher: GENERATOR
```

src/api/services/adminServiceHelpers.ts
```diff
@@ -57,11 +57,23 @@ export const AllowedSiteRoles: Record<string, AvailableClientRole[]> = {
   [ParticipantTypeData.Publisher.typeName]: ['GENERATOR', 'SHARER'],
 };
 
+export const AllowedSiteRolesById: Record<number, number[]> = {
```
alex-yau-ttd, line 60: I think this should be named `RecommendedSiteRolesById`, given it actually differs to the above `AllowedSiteRoles` (noting the difference with `SHARER`).

alex-yau-ttd: And a second thought: is it worth removing `AllowedSiteRoles` and `GetRecommendedRoles` given they are unused?

src/api/entities/Participant.ts
```diff
@@ -98,6 +98,17 @@ export const ParticipantCreationPartial = ParticipantSchema.pick({
   users: z.array(UserCreationPartial).optional(),
 });
 
+export const AdminParticipantCreationPartial = z.object({
```
alex-yau-ttd, line 101: I'm finding the naming in this file a bit hard to understand - e.g. what is the relevance of `Admin` in the name here? From the usage, it seems like this one is just to parse the request, whereas `participantCreationAndApprovalPartial` is used to construct the object that we send off to Objection.

Also, I think it might make more sense to move one or both of these elsewhere - I think having the parser live with the entity is, while a pattern we do have, not actually the desired pattern. Per my other comment, I think the goal would to be for the route here to just point to a method in a different file (e.g. https://github.com/IABTechLab/uid2-self-serve-portal/blob/main/src/api/routers/participants/participantsRouter.ts#L540) then the parser could also live in that separate file. 

Does that make sense?

src/web/components/Input/RadioInput.tsx
```diff
@@ -34,9 +34,14 @@ export function RadioInput<
         aria-invalid={error ? 'true' : 'false'}
         onValueChange={field.onChange}
       >
-        {options.map(({ optionLabel, value }) => (
+        {options.map(({ optionLabel, value, disabled }) => (
```
alex-yau-ttd, line 37: I think we should update the storybook to include options that are disabled. Also, it seems like we are missing some styling for the disabled option (e.g. greyed out and maybe `cursor: not-allowed`

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
```diff
@@ -0,0 +1,239 @@
...
+          </div>
+          <div>
+            <div className='add-participant-dialog-flex'>
+              <div className='left-column site-type'>
+                <RadioInput
+                  inputName='siteIdType'
+                  label='Site ID'
+                  options={[
+                    { optionLabel: 'Existing Site ID', value: 0, checked: true },
+                    { optionLabel: 'New Site ID', value: 1, disabled: true },
```
alex-yau-ttd, line 130: I presume we are just disabling this option during development?

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
```diff
@@ -0,0 +1,239 @@
...
+      open={open}
+      onOpenChange={setOpen}
+      className='add-participant-dialog'
+    >
+      <FormProvider {...formMethods}>
+        <form onSubmit={handleSubmit(onSubmit)}>
+          <h4>Participant Information</h4>
+          <span>Provide the following information about the company/participant</span>
+          <div className='add-participant-dialog-flex'>
+            <div>
```
alex-yau-ttd, line 101: I don't think we need this div

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
```diff
@@ -0,0 +1,239 @@
...
+                  optionLabel: p.typeName,
+                  value: p.id,
+                }))}
+                rules={{ required: 'Please specify Participant Type(s).' }}
+              />
+            </div>
+          </div>
+          <div>
+            <div className='add-participant-dialog-flex'>
+              <div className='left-column site-type'>
```
alex-yau-ttd, line 124: is the `left-column` class name just to help the developer understand what piece of UI it corresponds to? I don't see any css associated with it.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
```diff
@@ -0,0 +1,239 @@
...
+        const roles = GetRecommendedRolesById(types as number[]);
+        setValue('apiRoles', roles);
+      }
+    });
+    return () => subscription.unsubscribe();
+  }, [watch, setValue, open]);
+
+  useEffect(() => {
+    if (!open) {
+      reset();
```
alex-yau-ttd, line 69: I don't know what the fix is, but this doesn't seem to reset the SiteId search. Maybe can be fixed in a different ticket.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
```diff
@@ -0,0 +1,239 @@
...
+            </div>
+          </div>
+          <div>
+            <div className='add-participant-dialog-flex'>
+              <div className='left-column site-type'>
+                <RadioInput
+                  inputName='siteIdType'
+                  label='Site ID'
+                  options={[
+                    { optionLabel: 'Existing Site ID', value: 0, checked: true },
```
alex-yau-ttd, line 129: What does `checked: true` actually do here? I presumed it would select this option by default, but that doesn't seem to be the case.

src/web/screens/manageParticipants.tsx
```diff
@@ -20,19 +22,14 @@ import {
   UpdateParticipantForm,
 } from '../services/participant';
 import { GetAllParticipantTypes } from '../services/participantType';
-import { preloadSiteList } from '../services/site';
 import { RouteErrorBoundary } from '../utils/RouteErrorBoundary';
 import { PortalRoute } from './routeUtils';
 
+import '../components/ParticipantManagement/ParticipantManagement.scss';
```
alex-yau-ttd, line 28: Other screens just have the `scss` file living along side the screen file, which makes sense to me since it is styling _around_ the components (i.e. that live in `components/ParticipantManagement`) rather than styling the components themselves. 

What do you thinkl?

src/web/screens/manageParticipants.tsx
```diff
@@ -54,21 +51,39 @@ function ManageParticipants() {
...
-      </p>
       <Suspense fallback={<Loading />}>
         <Await resolve={data.results}>
           {([participantRequests, participantApproved, participantTypes, apiRoles]: [
             ParticipantRequestDTO[],
             ParticipantDTO[],
             ParticipantTypeDTO[],
-            ApiRoleDTO[]
+            ApiRoleDTO[],
+            string[]
```
alex-yau-ttd, line 68: what's the `string[]` for?

src/api/services/adminServiceHelpers.ts
ssundahlTTD: Good idea.  I added the comments.

src/api/services/adminServiceHelpers.ts
ssundahlTTD: I removed them.

src/web/components/Core/Form.tsx
```diff
@@ -52,10 +52,16 @@ export function Form<T extends FieldValues>({
         await onSubmit(formData);
       } catch (err) {
         if (onError) onError(err);
-        const message =
-          axios.isAxiosError(err) && (err.response?.data ?? [])[0]?.message
-            ? ((err.response?.data ?? [])[0]?.message as string)
-            : 'Something went wrong, please try again';
+        let message = 'Something went wrong, please try again';
+        if (axios.isAxiosError(err)) {
+          if (err.response?.data?.message) {
```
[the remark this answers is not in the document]
ssundahlTTD: So I'm going to leave this here for now because I know that it works.  The bigger issue is that I couldn't actually use Form.tsx here because it didn't support most of the hooks that I needed, so I had to use the vanilla FormProvider.  Unfortunately now my error messages aren't displaying properly on the form.  I'm going to write a new ticket for this.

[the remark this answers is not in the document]
ssundahlTTD: So I moved this to errorHelpers and in the process realized that my current implementation isn't even using this because I'm no longer using Form due to limitations.  I updated my dialog to show errors properly and I'm logging a ticket to update the Form component to add more react form hooks.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
ssundahlTTD: Correct.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
ssundahlTTD: removed

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
ssundahlTTD: Yeah, it was more for organizational purposes .  I had some extra styling on that at one point and removed it.  I'll remove the classnames.

src/web/screens/manageParticipants.tsx
ssundahlTTD: yeah, I agree.  I renamed and moved it.

src/web/components/Input/RadioInput.tsx
ssundahlTTD: I added styling and a test for disabled.

src/api/routers/participants/participantsRouter.ts
ssundahlTTD: I moved it to participantCreation.ts

src/api/entities/Participant.ts
ssundahlTTD: Admin* was just bad naming.  We already had a ParticipantCreationPartial so I needed something new but this ended up evolving into something that wasn't really a partial of the entity anymore.  I renamed it ParticipantCreationRequest.

I moved the parsers into participantClasses.ts.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
ssundahlTTD: That's a leftover from trying to get the form hooks to work before I realized Form.ts didn't support them.  I removed it.

src/web/components/ParticipantManagement/AddParticipantDialog.tsx
ssundahlTTD: I added this to https://atlassian.thetradedesk.com/jira/browse/UID2-2855.

src/web/components/Input/RadioInput.scss
```diff
@@ -35,4 +36,9 @@
 .radio-option {
   display: flex;
   align-items: center;
+
+  .disabled {
+    background-color: #ebebe4 !important;
```
alex-yau-ttd, line 41: We have `--theme-disabled-button` already - perhaps we could reuse that? Also the `!important` doesn't seem necessary when I tested locally with:
```
background-color: var(--theme-disabled-button);
```

alex-yau-ttd (approved):

src/web/components/Input/RadioInput.scss
ssundahlTTD: I changed it.  Not sure why but the default button background was taking priority when I tested this before.  I also removed the other !important

PR #340, opened 2024-02-29
Upgrade Keycloak to 23.0.0

alex-yau-ttd:
- Upgrade Keycloak to 23.0.0
- Update dependencies and docker yaml so that it actually works
- Style the show/hide password toggle that comes for free with version 23.
- Refactor `Keycloak.ts` to ensure tests in `PortalErrorBounday.test.tsx` do not fail with `Error: A 'Keycloak' instance can only be initialized once.`

## Bug fix
**Before**
If you login with user1, then successfully login with user2 while user1's session is still active, the login page will tell you you're already logged in and link you to the portal. But even though you just successfully authenticated with user2, your session will be for user1.

**After**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/cd52d17c-64ae-4c32-b2cc-b06c47cb643e

## Show/hide password toggle

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/f22fbdea-36c3-431c-b607-5654b09e6e50

**Updated hover styling**

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/50144c53-7956-4fea-925f-4bf181db5a29

## Manual testing
1. Check out this branch
2. Run `npm i`
3. Re-run docker command to run Keycloak on the updated version
4. Perform manual tests such as logging in, trying to login as user2 while user1's session is still active, logging out ,etc.


src/testHelpers/testContextProvider.tsx
```diff
@@ -1,12 +1,15 @@
 import { ReactKeycloakProvider } from '@react-keycloak/web';
+import Keycloak from 'keycloak-js';
 import { PropsWithChildren } from 'react';
 import { BrowserRouter } from 'react-router-dom';
 
-import keycloak from '../web/Keycloak';
+export const createTestKeycloakInstance = () => {
+  return new Keycloak('/api/keycloak-config');
```
alex-yau-ttd, line 7: TODO: figure out if I can just remove the config and have it be:
```
return new Keycloak('/api/keycloak-config');
```

src/web/Keycloak.ts
```diff
@@ -1,5 +1,20 @@
 import Keycloak from 'keycloak-js';
 
-const keycloak = new Keycloak('/api/keycloak-config');
+let keycloakInstance: Keycloak | null = null;
 
-export default keycloak;
+export const getKeycloakInstance = () => {
+  if (!keycloakInstance) {
+    throw new Error(
+      'No Keycloak instance found. Please call createKeycloakInstance() to initialize one.'
```
ssundahlTTD, line 8: Would it be bad to create the instance here if it doesn't already exist?

alex-yau-ttd: I think I prefer it as-is because we only ever want one keycloak instance in production, which I'm thinking should just be creating by calling `createKeycloakInstance()`, which should only be called once. I think that makes it easier to ensure only one is ever created. Does that some reasonable?

src/testHelpers/testContextProvider.tsx
alex-yau-ttd: Seems like yes, have made this change

keycloak/themes/uid2-theme/login/resources/css/login.css
```diff
@@ -112,11 +112,31 @@ body {
...
+.pf-c-button.pf-m-control[data-password-toggle]:hover {
+  color: var(--theme-action);
+}
+
 .pf-c-form-control {
   height: 45px !important;
   border-radius: 5px;
   margin-top: 8px;
   border: 1px solid var(--theme-input-border);
+  padding-right: 45px;
```
alex-yau-ttd, line 139: Ensures the password input does not interfere with the button

Dockerfile_keycloak
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/keycloak/keycloak:22.0.1
+FROM quay.io/keycloak/keycloak:23.0.0
```
alex-yau-ttd, line 1: This is the source of truth for the keycloak version.

docker-compose.yml
```diff
@@ -34,7 +34,16 @@ services:
     depends_on:
       - database
     command:
-      - start-dev --import-realm --spi-theme-static-max-age=-1 --spi-theme-cache-themes=false --spi-theme-cache-templates=false --spi-theme-welcome-theme=uid2-theme --health-enabled=true --metrics-enabled=true --features=declarative-user-profile --spi-connections-jpa-legacy-migration-strategy=update
+      - start-dev
```
alex-yau-ttd, line 37: Arguments need to be passed in separately rather than a single quoted value of all arguments, per: https://www.keycloak.org/docs/latest/upgrading/index.html#kc-sh-and-shell-metacharacters

ssundahlTTD (approved):

PR #363, opened 2024-03-13
Fix broken API tests

ssundahlTTD:
Fixed broken api tests

image


package.json
```diff
@@ -107,7 +107,7 @@
     "api-watch": "nodemon -e \".ts\" --exec npm run babel-node-api",
     "babel-node-api": "babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
     "build-api": "npx tsc --build ./tsconfig-api.json",
-    "test-api": "npx jest --testTimeout=5000 --forceExit --runInBand --config=./src/api/jest.config.ts --runInBand",
+    "test-api": "npx jest --testTimeout=5000 --forceExit --runInBand --config=./src/api/jest.config.ts",
```
alex-yau-ttd, line 110: Are we removing this because we already have the same flag earlier in the command?

src/api/tests/approversMiddleware.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('Approvers Service Tests', () => {
   const withToken = useTestServer();
 
   afterEach(() => {
+    jest.restoreAllMocks();
     jest.resetAllMocks();
```
alex-yau-ttd, line 13: I believe I can get rid of `resetAllMocks()` because `restoreAllMocks()` does everything `resetAllMocks()` does and more.

> mockFn.mockRestore()
> Does everything that [mockFn.mockReset()](https://jestjs.io/docs/mock-function-api#mockfnmockreset) does, and also restores the original (non-mocked) implementation.

Per: https://jestjs.io/docs/mock-function-api#mockfnmockrestore

src/api/tests/usersRoute.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('Users API tests', () => {
   describe('When a user re-invites another user', () => {
     afterEach(() => {
       jest.restoreAllMocks();
+      jest.clearAllMocks();
```
alex-yau-ttd, line 21: hm maybe take my other [comment](https://github.com/IABTechLab/uid2-self-serve-portal/pull/363/files#r1524060453) with a grain of salt, the same does not seem to apply here. I thought `restore` does everything that `reset` does and `reset` does everything that `clear` does, but indeed the tests fail without the `clear` here. Any idea why - am I missing something?

ssundahlTTD: it's in the info box in the docs:  

> mockFn.mockRestore() only works when the mock was created with jest.spyOn(). Thus you have to take care of restoration yourself when manually assigning jest.fn().

I had to add clearAllMocks() to reset the count on the function being called.  You were right on the other one, though

src/api/tests/approversMiddleware.spec.ts
ssundahlTTD: you're right.  I removed resetAllMocks()

src/api/tests/usersRoute.spec.ts
alex-yau-ttd: ahh makes sense, nice one. thanks!

alex-yau-ttd (approved):

PR #369, opened 2024-03-22
Add `<ScreenContentContainer>` component and use on all relevant screens

alex-yau-ttd:
Introduce a component that can be used on every screen to ensure consistent vertical spacing between components, rather than having bespoke classNames every time.

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/fbfde8b3-e930-4fe7-a491-bd2b0723c7b0


src/web/screens/emailContacts.scss
```diff
@@ -2,8 +2,4 @@
   display: flex;
   flex-direction: column;
   align-items: stretch;
-
-  .heading-details {
```
alex-yau-ttd, line 6: This is duplicated in `App.scss`

src/web/screens/manageParticipants.tsx
```diff
@@ -89,17 +90,19 @@ function ManageParticipants() {
...
-                participantTypes={participantTypes}
-                apiRoles={apiRoles}
-                onApprove={handleApproveParticipantRequest}
-              />
-              <ApprovedParticipantsTable
-                participants={participantApproved}
-                apiRoles={apiRoles}
-                onUpdateParticipant={onUpdateParticipant}
-              />
+              <ContentContainer>
```
alex-yau-ttd, line 93: This is the only visual change in this PR: I thought it could do with a bigger gap

src/web/screens/sharingPermissions.tsx
```diff
@@ -127,7 +128,7 @@ function SharingPermissions() {
         <br />
         Note: This only enables the sharing permission. No data is sent.
       </p>
-      <div className='bulk-add-and-search-collapsibles'>
+      <ContentContainer>
```
alex-yau-ttd, line 131: One reason I think this is better is because if we were add another component at the bottom of this div, we would need to change the above `className` to something like `x-and-y-and-z`

src/web/components/Core/ContentContainer.tsx
```diff
@@ -0,0 +1,11 @@
+import { ReactNode } from 'react';
+
+import './ContentContainer.scss';
+
+export type ContentContainerProps = {
+  children: ReactNode;
+};
+
+export function ContentContainer({ children }: ContentContainerProps) {
```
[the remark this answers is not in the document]
lionell-pack-ttd: Perhaps even naming might help here - if it's a ScreenContentContainer hopefully people will get that it's for Screen content, not just any old content that it happens to look right for.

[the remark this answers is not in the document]
alex-yau-ttd: I've just pushed up changes to rename to `ScreenContentContainer` and to use this across all the standard screens. Let me know what you think!

src/web/components/KeyPairs/KeyPairsTable.scss
```diff
@@ -4,7 +4,7 @@
   display: flex;
   flex-direction: column;
   align-items: stretch;
-  margin-bottom: 50px;
+  margin-bottom: 20px;
```
alex-yau-ttd, line 7: the row-gap in the `<ScreenContentContainer>` means we can reduce this by 30px

lionell-pack-ttd (approved):

PR #385, opened 2024-04-03
Add ability to rename and delete key pairs

ashleysmithTTD:
- Key pair name can be edited via edit button under Actions
- Key pair can be deleted ("disabled" on our end) via trash can button under Actions
- Disabled is no longer a property when a user creates a new key pair
- Disabled key pairs are filtered out in the api instead of the front end
- Removed display of disabled key in UI 
- Created storybook components for KeyPairDisableDialog and KeyPairEditDialog

Test Plan:

- Edit existing key pair
- Delete existing key pair
- Create new key pair and edit, then disable it
- Delete a key pair and then create a new key pair with the same name and make sure it can also be created
- Check that creating a key pair does not show a disabled checkbox


ssundahlTTD:
Looks like you can save a blank keypair name.  That should be required.

src/api/services/adminServiceClient.ts
```diff
@@ -200,6 +207,19 @@ export const addKeyPair = async (
...
 };
 
+export const updateKeyPair = async (
+  subscriptionId: string,
+  name: string,
+  disabled: boolean = false
+): Promise<KeyPairDTO> => {
+  const response = await adminServiceClient.post<KeyPairDTO>('/api/client_side_keypairs/update', {
+    subscription_id: subscriptionId,
+    disabled,
```
ssundahlTTD, line 217: The disabled flag in the update function here is good.  But the one above in the add function can be removed.  we no longer allow adding disabled keys.

src/web/components/KeyPairs/KeyPairEditDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  keyPair: KeyPairModel;
+}>;
+
+function KeyPairEditDialog({
+  onEdit,
+  triggerButton,
+  keyPair: keyPairInitial,
+}: KeyPairEditDialogProps) {
+  const [open, setOpen] = useState(false);
+  const [keyPair, updateKeyPair] = useState<KeyPairModel>(keyPairInitial);
```
ssundahlTTD, line 29: i think it's standard to call the state setter function set###().  So setKeyPair() here

src/web/screens/clientSideIntegration.tsx
```diff
@@ -46,6 +53,28 @@ function ClientSideIntegration() {
     }
   };
 
+  const handleUpdateKeyPair = async (formData: UpdateKeyPairFormProps) => {
+    const { name, subscriptionId, disabled = false } = formData;
+    try {
+      const response = await UpdateKeyPair({ name, subscriptionId, disabled });
+      if (response.status === 201) {
```
ssundahlTTD, line 60: You're not checking for a 200 status in handleDisableKeyPair below.  I wonder if you even need to here.  Is it possible to return a 200 from here but not a 201?  If not you can just not check for it because a non-20x response will trigger the fallback handler and catch block.

src/web/services/keyPairService.ts
```diff
@@ -28,3 +34,22 @@ export async function AddKeyPair(props: AddKeyPairFormProps) {
...
+  );
+  return result;
+}
+
+export async function DisableKeyPair(keyPair: KeyPairModel, participantId?: number) {
+  try {
+    await axios.delete(`/participants/${participantId ?? 'current'}/keyPair`, {
+      data: { keyPair },
+    });
+  } catch (e: unknown) {
```
ssundahlTTD, line 52: Do we actually want to catch this or just let the backup handler do it?

ssundahlTTD (approved):

PR #402, opened 2024-04-08
Replace Form.tsx and its references with FormProvider

ashleysmithTTD:
Components Changed (test that each of these components does what it should):
- KeyCreationDialog
- KeyDisableDialog
- KeyEditDialog
- KeyPairDialog
- KeyPairDisableDialog
- KeyPairEditDialog
- UpdateParticipantDialog
- TeamMemberDialog

Test Files Changed (run `npm run test` and make sure all tests pass)
- MultiCheckboxInput.spec.tsx 

Storybook Components Changed (test that each of these storybook components work as they should):
- MultiCheckboxInput
- RadioInput
- SelectInput
- TextInput

Other:
- Updated version number of react testing library
- Moved definition of withoutRef 
- Deleted Form.tsx, Form.spec.tsx, Form.scss, Form.stories.tsx


src/web/components/Input/RadioInput.stories.tsx
```diff
@@ -1,19 +1,30 @@
 import { ComponentMeta, ComponentStory } from '@storybook/react';
+import { FormProvider, useForm } from 'react-hook-form';
 
-import { Form } from '../Core/Form';
 import { RadioInput } from './RadioInput';
 import { Option } from './SelectInput';
 
+import '../Core/Form.scss';
```
ashleysmithTTD, line 7: Decided to leave the scss file since it is used several places but let me know if you think I should rename or move it

src/web/components/BusinessContacts/BusinessContactDialog.tsx
```diff
@@ -21,7 +20,12 @@ function BusinessContactDialog({
 }: BusinessContactDialogProps) {
   const [open, setOpen] = useState(false);
 
-  const onSubmit: SubmitHandler<BusinessContactForm> = async (formData) => {
+  const formMethods = useForm<BusinessContactForm>({
+    defaultValues: contact as BusinessContactForm,
```
lionell-pack-ttd, line 24: I don't think the `as BusinessContactForm` is needed - and I'm always wary of `as` because you lose type safety.

src/web/components/Input/RadioInput.stories.tsx
lionell-pack-ttd: I think src/web/styles would be a better place for it now.

src/web/components/Input/RadioInput.stories.tsx
```diff
@@ -1,19 +1,30 @@
...
-      <RadioInput {...args} inputName='radio' />{' '}
-    </Form>
+    <FormProvider {...formMethods}>
+      <form onSubmit={handleSubmit(() => {})}>
+        <RadioInput {...args} inputName='radio' />{' '}
+        <div className='form-footer'>
+          <button type='submit' className='primary-button'>
+            Submit
+          </button>
+        </div>
```
lionell-pack-ttd, line 25: I see a lot of repetition of these 5 lines throughout - I think it might be worth pulling this out into a SubmitButton component. You can have it take the same props as a button and pass them all to the button with the spread operator `...` to keep it just as flexible as an actual button.

src/web/components/Input/StyledCheckbox.tsx
```diff
@@ -1,9 +1,20 @@
...
+  FieldPath,
+  FieldValues,
+  useController,
+  UseControllerProps,
+  UseFormRegisterReturn,
+} from 'react-hook-form';
 
 import './MultiCheckboxInput.scss';
 
+export function withoutRef<TFieldName extends string>(props: UseFormRegisterReturn<TFieldName>) {
```
lionell-pack-ttd, line 13: I'm not sure if we even still need this - it was to solve a TypeScript problem but perhaps the problem doesn't exist any more.

lionell-pack-ttd: If we do still need it, it probably belongs somewhere shared, rather than with a specific input component.

src/web/components/KeyPairs/KeyPairDialog.tsx
```diff
@@ -1,47 +1,62 @@
...
   const [open, setOpen] = useState(false);
-  const { keyPair, onAddKeyPair } = props;
 
-  const onSubmit: SubmitHandler<AddKeyPairFormProps> = async (formData) => {
+  const formMethods = useForm<AddKeyPairFormProps>({
+    defaultValues: { name: keyPair?.name },
+  });
+  const { handleSubmit, setValue, reset } = formMethods;
+
+  useEffect(() => {
```
lionell-pack-ttd, line 25: I'm guessing this is to work around state still being around if you close the dialog and open it again? If so, I'd prefer a better solution - this risks us still having state floating around if we add more controls and forget to clear them here.

I think the real problem is that we keep the dialog around even after it's closed. It's not a great pattern and this is a good example why.

If you could switch to not rendering the dialog in whatever uses it, and controlling state from there, the problem would go away for good.

It'd be the classic {showKeyPairDialog && <KeyPairDialog ... />} instead of the current system, but you'd need to provide the button and move the showKeyPairDialog state to the parent. Give it a go but if it turns into too much of a change, just leave it like this and we can fix it another time.

src/web/components/Input/RadioInput.stories.tsx
ashleysmithTTD: Moved & renamed

src/web/components/KeyPairs/KeyPairDialog.tsx
ashleysmithTTD: Changed the structure for this ticket.  Tech debt ticket exists to change that pattern in the other dialogs

src/web/components/Input/RadioInput.stories.tsx
ashleysmithTTD: Created FormSubmitButton component

src/web/components/Input/StyledCheckbox.tsx
ashleysmithTTD: withoutRef removed everywhere

lionell-pack-ttd (approved):

PR #403, opened 2024-04-08
CSTG completion information panel

ssundahlTTD:
Add a CSTG completion checker and clean up layout

image

### DIFF
--- a/src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
+++ b/src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
@@ -15,6 +15,7 @@ function createKeypairFake(disabled: boolean) {
 }
 
 const meta: Meta<typeof ClientSideCompletion> = {
+  title: 'CSTG/ClientSideCompletion',
   component: ClientSideCompletion,
 };
 
--- a/src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
+++ b/src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
@@ -1,7 +1,18 @@
+import { faker } from '@faker-js/faker';
 import type { Meta, StoryObj } from '@storybook/react';
 
 import { ClientSideCompletion } from './ClientSideCompletion';
-import { faker } from '@faker-js/faker';
+
+function createKeypairFake(disabled: boolean) {
+  return {
+    created: faker.date.recent(),
+    createdString: faker.date.recent().toLocaleTimeString(),
+    disabled,
+    publicKey: faker.string.hexadecimal(),
+    siteId: faker.number.int(),
+    subscriptionId: faker.string.alpha(),
+  };
+}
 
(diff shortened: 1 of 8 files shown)


src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import { faker } from '@faker-js/faker';
```
ashleysmithTTD, line 1: ClientSideCompletion is kind of in a random place in Storybook - maybe move it under the CSTG section?

src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+export default meta;
+type Story = StoryObj<typeof ClientSideCompletion>;
+
+export const ReadyToCSTG: Story = {
+  args: {
+    domainNames: ['example.com'],
+    keyPairData: [createKeypairFake(false)],
+  },
+  storyName: 'Ready to CSTG (renders empty - nothing to say!)',
+  name: 'Ready to CSTG (blank)',
```
ashleysmithTTD, line 30: I know the Ready to CSTG tab is supposed to be blank, but it's a little confusing if it is broken or blank.  I like the text under storyName here if it could be displayed

src/web/components/ClientSideCompletion/ClientSideCompletion.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { KeyPairModel } from '../KeyPairs/KeyPairModel';
+
+import './ClientSideCompletion.scss';
+
+type ClientSideCompletionProps = Readonly<{
+  keyPairData: KeyPairModel[] | undefined;
+  domainNames: string[] | undefined;
+}>;
+export function ClientSideCompletion({ keyPairData, domainNames }: ClientSideCompletionProps) {
+  const hasKeyPairData = !!keyPairData?.filter((kp) => !kp.disabled).length;
```
ashleysmithTTD, line 13: Should not have to filter on disabled here, looks like `keyPairData` only contains key pairs where disabled = true

src/web/components/ClientSideCompletion/ClientSideCompletion.stories.tsx
ssundahlTTD: It's a little weird but it's working as intended because the component is rendering itself as it should.  We'd have to create a wrapper component to show a message.  Hopefully the (blank) in the navigation is semi-clear.

src/web/components/ClientSideCompletion/ClientSideCompletion.tsx
ssundahlTTD: I'm going to leave this on the off-chance that we change the API to bring back disabled keys in the future.  It's probably best that the component not need to worry about the details of how the API is returning data.

PR #405, opened 2024-04-10
Unique Key Pair Names per Site

ashleysmithTTD:
- If a key pair name exists and is enabled for a specific site id, do not allow user to create a new key pair with the same name
- When a key pair gets disabled, rename the key pair to [keyPairName]-disabled-[timestamp]

Test Plan:
- Try and create new key pair with same name as one displayed and make sure it does not create the key pair
- Create a new key pair, delete it, then create a new key pair with the same name and make sure the key pair gets created
- Edit a key pair to the same name as another existing key pair and check that it does not allow this

### DIFF
--- a/src/web/components/KeyPairs/KeyPairHelper.ts
+++ b/src/web/components/KeyPairs/KeyPairHelper.ts
@@ -1,0 +1,12 @@
+import { KeyPairModel } from './KeyPairModel';
+
+export const validateUniqueKeyPairName = (
+  value: string,
+  existingKeyPairs: KeyPairModel[] | undefined
+) => {
+  const existingKeyPairNames = existingKeyPairs?.map((keypair) => keypair.name);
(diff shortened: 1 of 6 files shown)


src/web/components/KeyPairs/KeyPairHelper.ts
```diff
@@ -0,0 +1,12 @@
+import { KeyPairModel } from './KeyPairModel';
+
+export const validateUniqueKeyPairName = (
+  value: string,
+  existingKeyPairs: KeyPairModel[] | undefined
+) => {
+  const existingKeyPairNames = existingKeyPairs?.map((keypair) => keypair.name);
```
ssundahlTTD, line 7: nit: you can probably simplify this to something like (existingKeyPairs.filter((k) => k.name === value).length > 0)

ssundahlTTD (approved):

PR #407, opened 2024-04-16
Change add top level domain to be a dialog

ashleysmithTTD:
What Changed

- Created CstgAddDomainDialog and corresponding test file and storybook file
- Used new dialog opening format
- Deleted any files associated with CstgInputRow
- Added two text helper functions (comma separated list, top-level domains)
- Wrote test file for textHelper functions
- Changed "Create" to "Add" Keyword in several places

Test Plan:
- Test that CstgAddDomainDialog works for both single domain and comma-separated list
- Test that entering invalid domain does not work for single and multiple domains
- Test that entering a valid, non top-level domain or domains gets automatically converted
- Test that submitting on a blank domain name does not work
- Test that deleting domains still works
- Test storybook component for CstgAddDomainDialog 
- Review test files and check they work


src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+};
+export default meta;
+
+type Story = StoryObj<typeof CstgAddDomainDialog>;
+
+export const Default: Story = {
+  args: {
+    onAddDomains: (newDomain) => Promise.resolve(console.log('New domain added: ', newDomain)),
+  },
+};
```
ashleysmithTTD, line 17: With the new dialog format, the dialog is automatically open in storybook now instead of pressing the on button.  Submitting still works and prints in the console the new domain, but we may need to update later if we would rather have the "Open" button displayed first and the user then clicks that.  Since nothing is broken though, I vote this is addressed in the ticket that fixes up the dialog pattern altogether.

ssundahlTTD:
Might be worth actually telling the user that they entered a duplicate here.
image

src/web/components/Core/InlineMessage.tsx
```diff
@@ -12,7 +12,7 @@ export type InlineMessageProps = {
 export function InlineMessage({ message, type }: InlineMessageProps) {
   return (
     <div className={clsx('inline-message message-container', type)}>
-      <span data-testid='banner-message' className='banner-text'>
+      <span data-testid='banner-message' className='banner-text' role='alert'>
```
ssundahlTTD, line 15: i don't know if you want to do this.  pretty sure this makes it always show up red.  we have a type prop for this.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+  onOpenChange: () => void;
+}>;
+
+type DomainProps = {
+  isIcann: boolean | null;
+  isPrivate: boolean | null;
+  domain: string | null;
+};
+
+type ValidDomainProps = Omit<DomainProps, 'domain'> & { domain: string };
```
ssundahlTTD, line 24: This seems a little awkward.  omitting the prop and re-adding a slightly different one with the same name.  Also, can the fields on DomainProps just be optional?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+function CstgAddDomainDialog({ onAddDomains, onOpenChange }: AddDomainNamesDialogProps) {
+  const [open, setOpen] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string>();
+
+  const formMethods = useForm<AddDomainNamesFormProps>();
+  const { handleSubmit, setValue, reset } = formMethods;
+
+  useEffect(() => {
+    if (!open) {
```
ssundahlTTD, line 34: I don't think you need to reset state anymore since this dialog is now de-rendering when it closes.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+      setValue('newDomainNames', '');
+      setErrorMessage('');
+      reset();
+    }
+  }, [open, setValue, reset]);
+
+  const isValidDomain = (domainProps: DomainProps): domainProps is ValidDomainProps => {
```
ssundahlTTD, line 41: This can probably be moved to textHelpers.ts.  It will probably be used in the edit dialog, too

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+      newDomainNamesFormatted.forEach((newDomainName, index) => {
+        newDomainNamesFormatted[index] = extractTopLevelDomain(newDomainName);
+      });
+      await onAddDomains(newDomainNamesFormatted);
+      setOpen(false);
+    }
+  };
+
+  return (
+    <div className='key-pair-dialog'>
```
ssundahlTTD, line 63: this should probably have its own classes

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Dialog title='Add Domains' closeButtonText='Cancel' open onOpenChange={onOpenChange}>
+        {!!errorMessage && <InlineMessage message={errorMessage} type='Error' />}
```
ssundahlTTD, line 65: check out the <RootFormErrors /> component for replacing this.  There's an example in AddParticipantsDialog

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <FormProvider {...formMethods}>
+          <form onSubmit={handleSubmit(onSubmit)}>
+            You may enter a single domain or enter domains as a comma separated list.
+            <TextInput
+              inputName='newDomainNames'
+              label='Domain Names'
+              rules={{ required: 'Please specify domain name.' }}
```
ssundahlTTD, line 72: grammar nit: specify 'a' domain name or maybe specify domain name(s)?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: Bad example of me copying an existing example :/ ValidDomainProps isn't needed, can just have DomainProps and domain is oprtional

src/web/components/Core/InlineMessage.tsx
ashleysmithTTD: Removed

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: Replaced

ashleysmithTTD: Replaced

ssundahlTTD (approved):

PR #408, opened 2024-04-18
Add ability to edit domain names

ashleysmithTTD:
What Changed:

- Created dialog to edit domain name
- Extracted code and created dialog to delete domain name
- Updated open/close pattern for edit domain dialog and delete domain dialog 
- Generalized update domain success toast to show different verbs depending on the action
- Fixed bug happening when editing a key pair where the user clicks edit, does not change anything, then clicks submit, it was showing an error but this is not an error and now just closes the dialog
- Updated styling for positioning of Add Domain and Add Key Pair buttons to be more clear 

Test Plan:
- Test edit domain dialog to work
- Test editing to non-valid or blank domain
- Test editing to domain that already is in the list
- Test opening edit domain dialog, changing nothing, clicking submit
- Test delete domain still works
- Test edit key pair dialog to open, change nothing, click submit


src/web/components/ClientSideTokenGeneration/CstgDomain.tsx
```diff
@@ -57,7 +74,41 @@ export function CstgDomainItem({ domain, onClick, onDelete, checked }: CstgDomai
...
+          <button
+            type='button'
+            className='icon-button'
+            title='Edit'
+            onClick={() => {
+              setShowEditDialog(true);
+            }}
+          >
+            <FontAwesomeIcon icon='pencil' />
+          </button>
```
ashleysmithTTD, line 86: Probably would create another ticket for this but thoughts about having EditButton component and DeleteButton component?  We are doing this in several places now with editing domains, key pairs, participant stuff and API keys.

src/web/components/ClientSideTokenGeneration/CstgDomain.tsx
```diff
@@ -3,30 +3,28 @@ import { useState } from 'react';
...
+  onOpenChange: () => void;
+}>;
 
-function DeleteConfirmationDialog({ domain, onRemoveDomain }: DeleteConfirmationDialogProps) {
-  const [openConfirmation, setOpenConfirmation] = useState(false);
-
-  const handleRemove = async () => {
-    setOpenConfirmation(false);
-    await onRemoveDomain();
+function DeleteConfirmationDialog({
```
alex-yau-ttd, line 14: There's now separate files for adding and editing domains. Perhaps now is a good time to also extract the delete dialog component to a separate file, for consistency?

alex-yau-ttd: Then this file could be renamed `CstgDomainItem.tsx`

src/web/components/ClientSideTokenGeneration/CstgDomain.tsx
alex-yau-ttd: Sounds like a great idea!

src/web/components/KeyPairs/KeyPairsTable.scss
```diff
@@ -4,6 +4,7 @@
   display: flex;
   flex-direction: column;
   align-items: stretch;
+  margin-bottom: 75px;
```
alex-yau-ttd, line 7: This gap feels a bit big for me. I'm also planning on changing how we space components generally in https://github.com/IABTechLab/uid2-self-serve-portal/pull/369. Perhaps once we are consistent with where all the "Add" buttons are positioned (i.e. above the table, rather than below), this will not cause as much confusion, and hence a smaller gap would be acceptable?

alex-yau-ttd (approved):

src/web/components/KeyPairs/KeyPairsTable.scss
ashleysmithTTD: Sounds good, reduced to 50px.  I think while you were out, we decided all add buttons should be above the table so that it does not become hard to find once someone has a lot on their table and has to scroll.  A smaller gap is good, but lets not keep it so small that it could look like Add Domain belongs to the key pair table - that was a concern Kimberly brought up I believe

src/web/components/ClientSideTokenGeneration/CstgDomain.tsx
ashleysmithTTD: Done

PR #409, opened 2024-04-22
Bulk update for domains

ashleysmithTTD:
What Changed:
- Added several types of separators for adding domains
- Added optional button to allow users to delete their existing domains before adding new ones
- Display the invalid domains to help the user correct their list
- Created MultilineTextInput component plus storybook and test files that use text area (larger and scrollable)
- Moved over DeleteDomainDialog to where the component is defined and adapted it for multiple domains
- Fixed bug when deleting several domains to make the list scrollable
- Switched up the order of domain name validation due to bug found of two different domains added turning into the same top-level domain
- Wording/grammar changes from Gen: removed use of (s), capitalized add button words 
- Updated test files as necessary

Test Plan:
- Add multiple domains with as many different combinations as possible including deleting domains before or not, and entering invalid domains. 
- Test out deleting domains especially a lot and make sure dialog still looks good/works


src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -28,49 +35,67 @@ function CstgAddDomainDialog({
   } = formMethods;
 
   const onSubmit = async (formData: AddDomainNamesFormProps) => {
-    const newDomainNames = separateStringsCommaSeparatedList(formData.newDomainNames);
-    // filter for uniqueness on what the user entered AND against existing domains
-    const dedupedDomains = deduplicateStrings(newDomainNames);
-    const uniqueDomains = dedupedDomains.filter((domain) => !existingDomains?.includes(domain));
+    const newDomains = separateStringsList(formData.newDomains);
+    // filter out domain names that already exist in the list
+    const uniqueDomains = newDomains.filter((domain) => !existingDomains?.includes(domain));
```
lionell-pack-ttd, line 40: Nit: They're not necessarily unique - the user might have entered the same domain twice.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -28,49 +35,67 @@ function CstgAddDomainDialog({
...
+            </div>
+            <MultilineTextInput
+              inputName='newDomains'
+              label='Domains'
+              rules={{ required: 'Please specify domains.' }}
             />
             <div className='form-footer'>
               <button type='submit' className='primary-button'>
-                Add domains
+                Add Domains
```
lionell-pack-ttd, line 98: You could update the text of this button if `deleteExistingList` is checked - maybe "Replace Domains"?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: Yes - I am filtering again in line 66.  I originally had both filters in this line, but then I found a bug where two different valid domains entered changed to the same top-level domain, so I moved the second filter for deduplicating the list to after all domains are top-level.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: Good idea! Changed

ssundahlTTD (approved):

PR #412, opened 2024-04-24
Added vulnerability scanning

gmsdelmundo:


.trivyignore
```diff
@@ -1,13 +1,5 @@
...
 # for more details
-
-# e.g.
-
+# e.g. 
 # CVE-2022-3996
-
-# https://atlassian.thetradedesk.com/jira/browse/UID2-3300
-
 CVE-2022-29622
```
alex-yau-ttd, line 5: Should we leave a comment for why we are ignoring or linking to a Jira ticket to address the vulnerability?

package.json
```diff
@@ -216,6 +215,7 @@
     "nodemon": "^2.0.22",
     "npm-run-all": "^4.1.5",
     "prop-types": "^15.8.1",
+    "react-scripts": "5.0.1",
```
alex-yau-ttd, line 218: For my understanding, what's the significance of moving react-scripts to a dev dependency in this PR?

alex-yau-ttd: Follow-up question: does the scan only apply to `dependencies` and not to `devDependencies`?

.trivyignore
gmsdelmundo: @ssundahlTTD do you have the ticket number for this?

package.json
gmsdelmundo: > Follow-up question: does the scan only apply to dependencies and not to devDependencies?

Yep, it skips `devDependencies` https://github.com/aquasecurity/trivy/discussions/3714

I don't think `react-scripts` needs to be in `dependencies` since it might be a dev-only dependency, and since it has the vulnerability that failed the build previously, it should be safe to move it over to `devDependencies`

@ssundahlTTD can confirm

alex-yau-ttd: excellent, thanks for explaining!

alex-yau-ttd: There's definitely [discussion](https://github.com/facebook/create-react-app/issues/2696) as to whether `react-scripts` should be installed as a `devDependecy` by default by create-react-app, but in any case it looks safe to move 👍

alex-yau-ttd (approved):

.trivyignore
ssundahlTTD: https://atlassian.thetradedesk.com/jira/browse/UID2-2633 is what I'm working on for the other vulnerbailities

PR #413, opened 2024-04-24
New pattern for opening dialog from parent part 1

ashleysmithTTD:
In the components that have been changed:

- Trigger button removed
- Added state variable in parent to conditionally render the dialog
- onOpenChange function created in parent and sent to child component
- Currently, open is always set to true for the dialog, will get rid of open variable once I edit the Dialog component itself
- Removed [open, setOpen] from the dialog components
- Button to open dialog is now defined outside the dialog with an onClick function

Test Plan:

- When creating an account, test that terms and conditions dialog opens and closes correctly
- Dialogs to create API key, edit API key, delete API key 
- Dialogs to create email contact, edit email contact, delete email contact
- Dialog to edit domain name
- Dialogs to create key pair, edit key pair, delete key pair


src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -16,6 +16,16 @@ type KeyItemProps = {
 };
 function KeyItem({ apiKey: apiKeyInitial, onEdit, onDisable, availableRoles }: KeyItemProps) {
   const [apiKey, setApiKey] = useState<ApiKeyDTO>(apiKeyInitial);
+  const [showKeyDisableDialog, setShowKeyDisableDialog] = useState<boolean>(false);
+  const [showKeyEditDialog, setShowKeyEditDialog] = useState<boolean>(false);
```
ashleysmithTTD, line 20: Is there a way to combine these or is is best to leave it as two separate lines?

src/web/components/BusinessContacts/BusinessContactsTable.tsx
```diff
@@ -10,19 +12,28 @@ function NoEmailContact({
 }: {
   onAddEmailContact: (form: BusinessContactForm) => Promise<void>;
 }) {
+  const [showBusinessContactDialog, setShowBusinessContactDialog] = useState<boolean>(false);
+  const onOpenChangeBusinessContactDialog = () => {
+    setShowBusinessContactDialog(!showBusinessContactDialog);
+  };
   return (
     <TableNoDataPlaceholder
```
ssundahlTTD, line 20: This feels like it can be combined into the component below.  Right now there's a bunch of duplicated code.

src/web/components/ApiKeyManagement/KeyItem.tsx
ssundahlTTD: you could probably be tricky and store an object in state.  like { firstDialog: true, secondDialog: false }, but that seems more complicated than helpful.

ssundahlTTD (commented):
.

src/web/components/BusinessContacts/BusinessContactsTable.tsx
ashleysmithTTD: You're right - cleaned up a lot

PR #419, opened 2024-04-29
New dialog pattern part 3 storybook files

ashleysmithTTD:
Adjusted the storybook files to use the new pattern so that you can still open and close the dialog, not just see the dialog open all the time.  Every dialog should print something in the console when submitted to verify it worked and submitted the correct information.  Cleaned up reusable code where possible.

### DIFF
--- a/src/web/components/ClientSideTokenGeneration/CstgDeleteDomainDialog.stories.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgDeleteDomainDialog.stories.tsx
@@ -1,4 +1,5 @@
-import { Meta, StoryObj } from '@storybook/react';
+import { Meta } from '@storybook/react';
+import { useState } from 'react';
 
 import CstgDeleteDomainDialog from './CstgDeleteDomainDialog';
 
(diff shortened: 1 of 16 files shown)


src/web/components/ClientSideTokenGeneration/CstgDeleteDomainDialog.stories.tsx
```diff
@@ -8,10 +9,44 @@ const meta: Meta<typeof CstgDeleteDomainDialog> = {
...
+      <button className='small-button' type='button' onClick={() => setIsOpen(!isOpen)}>
+        Open Dialog
+      </button>
+      {isOpen && (
+        <CstgDeleteDomainDialog
+          onOpenChange={() => setIsOpen(!isOpen)}
+          domains={['testdomain.com']}
+          onRemoveDomains={() => {
+            Promise.resolve(console.log(`Disabling Domain`));
+            setIsOpen(!isOpen);
```
ssundahlTTD, line 25: This isn't working.  I think it may need to be before returning the promise like in CstgAddDomainDialog.  It also doesn't work with editing, which follows this pattern.

ssundahlTTD:
Now is probably a cool time to move the Key Pairs storybooks under the CSTG section.  Easier to find.

ssundahlTTD (commented):
couple of comments

src/web/components/ClientSideTokenGeneration/CstgDeleteDomainDialog.stories.tsx
ashleysmithTTD: Hmm I'm seeing that these two are working on my end

ashleysmithTTD:
> Now is probably a cool time to move the Key Pairs storybooks under the CSTG section. Easier to find.

Moved them to CSTG -> Key Pair and then put all the domains components into CSTG -> Domains

ssundahlTTD (approved):

PR #424, opened 2024-05-02
Paging and searching domains

ashleysmithTTD:
What Changed:
- Added search bar in domains table 
- New paging tool component
- New Select Dropdown component largely based off MultiSelectDropdown
- Two new Font Awesome icons 
- Wrote sort strings helper function
- Changed 'top-level domain' to 'root-level domain' from slack discussions

Test Plan:

Test as many combinations as possible of paging the domain:
- switching pages with arrows and the dropdown
- changing the number of domains per page
- searching domains
- editing domain
- single delete domain
- bulk delete domain
- bulk delete all domains 
- adding domains
- adding & replacing domains

Test storybook components (PagingTool and Select Dropdown)


src/web/components/Core/PagingTool.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const rowsPerPageOptions =
+    numberTotalRows > 10
+      ? rowsPerPageValues
+          .filter((value) => value <= numberTotalRows)
+          .map((value) => ({
+            name: value.toString(),
+            id: value,
+          }))
+      : [{ name: rowsPerPageValues[0].toString(), id: rowsPerPageValues[0] }];
```
ashleysmithTTD, line 62: Making sure that there is at least one option for rows per page so it's not a blank dropdown even if there are only say 5 rows

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -21,10 +23,50 @@ export function CstgDomainsTable({
...
+      );
+    }
+  }, [domains, pageNumber, rowsPerPage, searchText, searchedDomains]);
+
+  useEffect(() => {
+    if (searchText !== '') {
+      setPageNumber(initialPageNumber);
+      setRowsPerPage(initialRowsPerPage);
+    }
+  }, [searchedDomains.length, searchText]);
```
ashleysmithTTD, line 68: When the search changes, I found it best to just go back to the first page and initial rows per page for both functionality and UX

src/web/utils/textHelpers.spec.ts
alex-yau-ttd, line 1: I see it wasn't adding in this PR, but I noticed the `format unix date` test is failing for me locally:
```
  ● Text helper tests › format unix date › should return correct unix date

    expect(received).toEqual(expected) // deep equality

    Expected: "4/16/2024"
    Received: "17/04/2024"

      59 |   describe('format unix date', () => {
      60 |     it('should return correct unix date', () => {
    > 61 |       expect(formatUnixDate(1713306239)).toEqual('4/16/2024');
         |                                          ^
      62 |     });
      63 |   });
      64 |
```
I can't think of a great way to address that without essentially duplicated the logic from the production code:
```
describe('format unix date', () => {
  it('should return correct unix date', () => {
    const unixTimestamp = 1713306239;
    const date = new Date(unixTimestamp * 1000);
    const dateString = date.toLocaleDateString();
    expect(formatUnixDate(unixTimestamp)).toEqual(dateString);
  });
});
```
Perhaps worth changing to that or removing the test entirely. What are your thoughts?

src/web/utils/textHelpers.ts
```diff
@@ -33,3 +33,11 @@ export const separateStringsList = (strings: string): string[] => {
 export const deduplicateStrings = (strings: string[]) => {
   return strings.filter((val, index, arr) => arr.indexOf(val) === index);
 };
+
+export const sortStrings = (strings: string[]) => {
```
alex-yau-ttd, line 37: perhaps worth calling this something like `sortStringsAlphabetically` to avoid confusion with [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order)

src/web/utils/textHelpers.spec.ts
```diff
@@ -94,4 +95,17 @@ describe('Text helper tests', () => {
       }
     });
   });
+
+  describe('sortstrings', () => {
```
alex-yau-ttd, line 99: per my comment on the `sortStrings` method - perhaps it's worth covering the case where the strings are numbers to differentiate from natural order:
```
describe('sort number strings alphabetically', () => {
  const testStringArr = [
    ['1', '2', '10'],
    ['1', '10', '2'],
  ];
  it.each(testStringArr)('should return array sorted alphabetically', () => {
    for (const t of testStringArr) {
      expect(sortStrings(t)).toEqual(['1', '10', '2']);
    }
  });
});
```

alex-yau-ttd:
Is it worth changing comments-in-code that refer to `top-level` domains to use `root-level` as well? e.g. `CstgAddDomainDialog.tsx`

src/web/components/Core/PagingTool.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export function PagingTool({
+  numberTotalRows,
+  onChangeRows,
+  rowsPerPageTitle = 'Rows Per Page',
+  initialRowsPerPage = 10,
+  initialPageNumber = 1,
+}: PagingProps) {
+  const initialPageNumberOptions = Array.from(
+    { length: Math.ceil(numberTotalRows / initialRowsPerPage) },
```
alex-yau-ttd, line 28: There's a bunch of math in this component which might be nice to be refactored out. As an example, this math is done twice in this file. Then we could potentially add tests around them.

Ideally this file should probably only change if the wanted to change how it looked, while the behind-the-scenes calculations could be extracted away into a separate `.ts` file.

Does that seem reasonable?

src/web/components/Core/PagingTool.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [pagedDomains, setPagedDomains] = useState<string[]>(totalDomains);
+  const [pageNumber, setPageNumber] = useState<number>(1);
+  const [rowsPerPage, setRowsPerPage] = useState<number>(10);
+
+  const onChangeDisplayedRows = (currentPageNumber: number, currentRowsPerPage: number) => {
+    setPageNumber(currentPageNumber);
+    setRowsPerPage(currentRowsPerPage);
+  };
+
+  useEffect(() => {
```
alex-yau-ttd, line 55: I'm not sure how it would work - but is it possible to have this story function properly without this `useEffect`?

src/web/components/Core/PagingTool.tsx
```diff
@@ -0,0 +1,160 @@
...
+        >
+          <FontAwesomeIcon icon='angle-left' />
+        </button>
+      </div>
+      <SelectDropdown
+        containerClass='page-numbers-dropdown-container'
+        className='page-numbers-dropdown'
+        initialValue={pageNumberOptions[0]}
+        updatedValue={pageNumberOptions.find((number) => number.id === pageNumber)}
+        title=''
```
alex-yau-ttd, line 143: maybe it would be better if the `title` prop was optional?

src/web/utils/textHelpers.ts
ashleysmithTTD: renamed

src/web/components/Core/PagingTool.tsx
```diff
@@ -0,0 +1,160 @@
...
+  onChangeRows: (currentPageNumber: number, currentRowsPerPage: RowsPerPageValues) => void;
+  rowsPerPageTitle?: string;
+  initialRowsPerPage?: RowsPerPageValues;
+  initialPageNumber?: number;
+}>;
+
+export function PagingTool({
+  numberTotalRows,
+  onChangeRows,
+  rowsPerPageTitle = 'Rows Per Page',
```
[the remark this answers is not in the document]
ashleysmithTTD: Yep! Done

src/web/utils/textHelpers.spec.ts
ashleysmithTTD: Done

src/web/components/Core/PagingTool.tsx
ashleysmithTTD: Yeah good call - changed

src/web/components/Core/PagingTool.tsx
ashleysmithTTD: Definitely - it cleaned up the file nicely

src/web/components/Core/PagingTool.stories.tsx
ashleysmithTTD: Yeah, fixed

src/web/utils/textHelpers.spec.ts
ashleysmithTTD: Changed!

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.spec.tsx
alex-yau-ttd, line 1: This can just be a `.ts` file, rather than `.tsx`

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.spec.tsx
```diff
@@ -19,4 +19,66 @@ describe('test domain helper functions', () => {
...
+      'test23.com',
+      'test24.com',
+      'test25.com',
+      'test26.com',
+      'test27.com',
+      'test28.com',
+      'test29.com',
+    ];
+    it('should return a filtered first page domain list', () => {
+      expect(getPagedDomains(domainsList, 1, 10)).toEqual([
```
alex-yau-ttd, line 56: nit: it'd be nice if we could extract these numbers into a variable so it's clearer what it is testing without having to look at the signature of the method, so something like:
```
const pageNumber = 1;
const rowsPerPage = 10;
expect(getPagedDomains(domainsList, pageNumber, rowsPerPage)).toEqual([
...
```

Then similar for the other tests.

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -95,6 +92,10 @@ export function CstgDomainsTable({
     );
     setShowDeleteDomainsDialog(false);
     setSelectedDomains([]);
+    setSearchedDomains([]);
+    setSearchText('');
+    if (deleteDomains.length === 1) setDomainAction('DeleteOne');
```
alex-yau-ttd, line 97: nit: perhaps we should be consistent with how we format `if-else` blocks. Just below on line 102 through 106 we use brackets. Or we could use a ternary but doesn't bother me: 
```
setDomainAction(deleteDomains.length === 1 ? 'DeleteOne' : 'DeleteMany');
```

src/web/components/Core/PagingTool.tsx
```diff
@@ -1,70 +1,46 @@
...
-            id: value,
-          }))
-      : [{ name: rowsPerPageValues[0].toString(), id: rowsPerPageValues[0] }];
-
-  const initialRowsPerPageOption = rowsPerPageOptions.find(
-    (number) => number.id === initialRowsPerPage
-  );
-
   const onIncreasePageNumber = () => {
     if (pageNumber < Math.ceil(numberTotalRows / rowsPerPage)) {
```
alex-yau-ttd, line 45: nit: could we still extract this condition please? something like:
```
export const canIncreasePageNumber = (pageNumber, numberTotalRows, rowsPerPage) => {
  return pageNumber < Math.ceil(numberTotalRows / rowsPerPage);
};
```

src/web/components/Core/PagingToolHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const getRowsPerPageOptions = (numberTotalRows: number) => {
+  const rowsPerPageOptions: RowsPerPageValues[] = [];
+  let i = 0;
+  while (i < rowsPerPageValues.length) {
+    rowsPerPageOptions.push(rowsPerPageValues[i]);
+    if (numberTotalRows <= rowsPerPageValues[i]) break;
+    i += 1;
+  }
+  return rowsPerPageOptions.map((value) => ({
```
alex-yau-ttd, line 23: could this possibly return a `RowsPerPageOptions` type that has a `string` and a `RowsPerPageValues`? I'm finding it hard to understand what `name` and `id` mean here, especially since the same pattern is used for the page number options, so a similar comment applies to that.

src/web/components/Core/PagingToolHelper.spec.tsx
```diff
@@ -0,0 +1,133 @@
+import {
+  getPageNumberOptions,
+  getRowsPerPageOptions,
+  getShowingRowsText,
+} from './PagingToolHelper';
+
+describe('test paging tool helper functions', () => {
+  describe('check that correct rows per page options are returned', () => {
+    it('should return at least one rows per page option', () => {
+      expect(getRowsPerPageOptions(3)).toEqual([{ name: '10', id: 10 }]);
```
alex-yau-ttd, line 10: I think this could benefit from some re-jigging. Could we extract out the expected result to a const and potentially parameterise the test and do some renaming - something like:
```
const singleRowsPerPageOption = [{ name: '10', id: 10 }];

test.each([[3], [1], [10]])(
  'should return a single rows per page option for %i total rows',
  (numberTotalRows) => {
    expect(getRowsPerPageOptions(numberTotalRows)).toEqual(singleOption);
  }
);
```

src/web/components/Core/PagingToolHelper.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+        { name: '10', id: 10 },
+        { name: '25', id: 25 },
+        { name: '50', id: 50 },
+        { name: '100', id: 100 },
+        { name: '250', id: 250 },
+      ]);
+    });
+  });
+
+  describe('check that correct page numbers are returned', () => {
```
alex-yau-ttd, line 39: Similar in vein to my other comments, there's quite a few magic numbers and repetition here that makes it hard to follow - perhaps we could try extracting some constants, parameterisem etc.?

src/web/utils/textHelpers.spec.ts
```diff
@@ -96,15 +99,39 @@ describe('Text helper tests', () => {
...
     it.each(testStringArr)('should return array sorted alphabetically', () => {
       for (const t of testStringArr) {
-        expect(sortStrings(t)).toEqual(['atest', 'btest', 'ctest']);
+        expect(sortStringsAlphabetically(t)).toEqual(['atest', 'btest', 'ctest']);
+      }
+    });
+
+    const testStringArrNumbers = [['1test', '7test', '10test']];
+    it.each(testStringArrNumbers)('should return array sorted alphabetically', () => {
+      for (const t of testStringArrNumbers) {
```
alex-yau-ttd, line 116: I don't think this `for` loop is necessary if you are using `it.each`, as that should already do the loop for you. Try something like:
```
const testStringArrays = [
  ['atest', 'btest', 'ctest'],
  ['btest', 'atest', 'ctest'],
  ['ctest', 'btest', 'atest'],
];
const expected = ['atest', 'btest', 'ctest'];
testStringArrays.forEach((testStringArray) => {
  it('should return an alphabetically sorted array', () => {
    expect(sortStringsAlphabetically(testStringArray)).toEqual(expected);
  });
});
```

src/web/utils/textHelpers.spec.ts
```diff
@@ -96,15 +99,39 @@ describe('Text helper tests', () => {
     });
   });
 
-  describe('sortstrings', () => {
+  describe('sort strings alphabetically', () => {
```
alex-yau-ttd, line 102: Is it worth adding tests for other cases - mixed case strings, special characters, duplicate strings, single element, empty array?

src/web/utils/textHelpers.spec.ts
```diff
@@ -96,15 +99,39 @@ describe('Text helper tests', () => {
...
     const testStringArr = [
       ['atest', 'btest', 'ctest'],
       ['btest', 'atest', 'ctest'],
       ['ctest', 'btest', 'atest'],
     ];
     it.each(testStringArr)('should return array sorted alphabetically', () => {
       for (const t of testStringArr) {
-        expect(sortStrings(t)).toEqual(['atest', 'btest', 'ctest']);
+        expect(sortStringsAlphabetically(t)).toEqual(['atest', 'btest', 'ctest']);
```
alex-yau-ttd, line 110: I don't think this `for` loop is necessary if you are using `it.each`, as that should already do the loop for you. Try something like:
```
const testStringArrays = [
  ['atest', 'btest', 'ctest'],
  ['btest', 'atest', 'ctest'],
  ['ctest', 'btest', 'atest'],
];
const expected = ['atest', 'btest', 'ctest'];
testStringArrays.forEach((testStringArray) => {
  it('should return an alphabetically sorted array', () => {
    expect(sortStringsAlphabetically(testStringArray)).toEqual(expected);
  });
});
```

src/web/utils/textHelpers.spec.ts
```diff
@@ -96,15 +99,39 @@ describe('Text helper tests', () => {
...
+      }
+    });
+
+    const testStringArrNumbers = [['1test', '7test', '10test']];
```
alex-yau-ttd, line 114: I think we can just simplify this to:
```
it('should return array sorted alphabetically', () => {
  const testStringArrNumbers = ['1test', '7test', '10test'];
    expect(sortStringsAlphabetically(testStringArrNumbers)).toEqual(['10test', '1test', '7test']);
  })
});
```

alex-yau-ttd: And this probably applies to other methods below as well

src/web/components/Core/PagingToolHelper.ts
ashleysmithTTD: This is related to the SelectOption type that the SelectDropdown and MultiselectDropdowns use- the SelectOption type has the format - SelectOption<TValue> is { id: TValue, name: string}

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -43,7 +47,34 @@
...
+    }
+
+    .domains-search-bar {
+      width: 100%;
+      border: none;
+      outline: none;
+      color: var(--theme-action);
+      font-weight: 400;
+      font-size: 0.75rem;
+      line-height: 0.9375rem;
```
alex-yau-ttd, line 77: Can we replace this with a unitless line-height? i.e. no `rem`. Otherwise it can lead to issues where the font is too big for the line-height.

src/web/components/Core/PagingToolHelper.ts
alex-yau-ttd: ah - so can it use the `SelectOption` type then?

src/web/components/Core/PagingTool.tsx
ashleysmithTTD: Done

src/web/components/Core/PagingToolHelper.ts
ashleysmithTTD: Done

src/web/utils/textHelpers.spec.ts
ashleysmithTTD: Fixed

src/web/utils/textHelpers.spec.ts
ashleysmithTTD: Done

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.spec.tsx
ashleysmithTTD: Done

src/web/utils/textHelpers.spec.ts
ashleysmithTTD: Added more tests to account for this

src/web/components/Core/PagingTool.tsx
```diff
@@ -42,14 +44,14 @@ export function PagingTool({
   }, [numberTotalRows, initialRowsPerPage, initialPageNumber]);
 
   const onIncreasePageNumber = () => {
-    if (pageNumber < Math.ceil(numberTotalRows / rowsPerPage)) {
+    if (canIncreasePageNumber(pageNumber, numberTotalRows, rowsPerPage)) {
```
alex-yau-ttd, line 47: I think we can actually move this condition to the disabled state of the button - then we won't need to check it here in the `onClick`? i.e. 
```
const onIncreasePageNumber = () => {
  setPageNumber(pageNumber + 1);
  onChangeRows(pageNumber + 1, rowsPerPage);
};
...
<button
  type='button'
  className='icon-button'
  title='Next Page'
  onClick={onIncreasePageNumber}
  disabled={!canIncreasePageNumber(pageNumber, numberTotalRows, rowsPerPage)}
>
  <FontAwesomeIcon icon='angle-right' />
</button>
```

How's that sound?

alex-yau-ttd: then ditto for decrementing the page number

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
alex-yau-ttd: I think we can probably just remove this line entirely

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -43,7 +47,34 @@
 
     &-right {
       display: flex;
-      align-items: center;
+      justify-content: right;
+    }
+
+    .domains-search-bar-container {
+      display: flex;
```
alex-yau-ttd, line 54: it might be worth adding `align-items: center` here so that the search icon lines up with the text better: 

Before
image


After
image

alex-yau-ttd (approved):

PR #432, opened 2024-05-14
Show all invalid domains in add domains dialog 

ashleysmithTTD:
What Changed:

- Domain validation all happening on the backend in admin (removed domain validation from front end)
- All types of invalid root-level domains entered will show in both Add Domain dialog and Edit Domain Dialog
- Created an UpdatedDomainResponse type that returns the domains and if they are valid or invalid
- Added removal for '[' and ']' in the trim function

Test Plan:

- Test adding invalid domain with typo (example: test....com)
- Test adding invalid domain that is not valid root-level (examples: blogspot.com, github.io, wmflabs.org)
- Test adding multiple invalid domains 
- Test editing a valid domain to an invalid domain
- Test AddDomainDialog and EditDomainDialog storybook components

### DIFF
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -13,7 +13,7 @@ describe('CstgDomainAddDomainDialog', () => {
     const user = userEvent.setup();
 
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve();
+      return Promise.resolve([]);
     });
 
     render(
@@ -21,7 +21,6 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={[]}
       />
     );
 
@@ -42,7 +41,7 @@ describe('CstgDomainAddDomainDialog', () => {
     const user = userEvent.setup();
 
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve();
+      return Promise.resolve([]);
     });
 
     render(
@@ -50,7 +49,6 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={[]}
       />
     );
 
@@ -68,7 +66,7 @@ describe('CstgDomainAddDomainDialog', () => {
 
   it('should render error when user types single incorrect domain', async () => {
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve();
+      return Promise.resolve(['newDomains']);
     });
 
     render(
@@ -76,7 +74,6 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={['newDomains']}
       />
     );
 
@@ -87,7 +84,7 @@ describe('CstgDomainAddDomainDialog', () => {
 
   it('should render error if user types in at least one incorrect domain in a list', async () => {
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve();
+      return Promise.resolve(['test', 'test2']);
     });
 
     render(
@@ -95,7 +92,6 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={['test', 'test2']}
       />
     );
 
@@ -106,7 +102,7 @@ describe('CstgDomainAddDomainDialog', () => {
 
   it('should render error if user submits empty text box for domain names', async () => {
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve();
+      return Promise.resolve([]);
     });
 
     render(
@@ -114,7 +110,6 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={[]}
       />
     );
     await submitDialog();
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -61,6 +61,11 @@ describe('CstgDomainAddDomainDialog', () => {
 
     await submitDialog();
 
+    expect(onAddDomainsMock).toHaveBeenCalledWith(
+      ['test.com', 'test2.com', 'test3.com', 'test4.com'],
+      expect.anything()
+    );
+
     expect(screen.queryByRole('alert')).not.toBeInTheDocument();
   });
 
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -65,6 +65,8 @@ describe('CstgDomainAddDomainDialog', () => {
   });
 
   it('should render error when user types single incorrect domain', async () => {
+    const user = userEvent.setup();
+
     const onAddDomainsMock = jest.fn(() => {
       return Promise.resolve(['newDomains']);
     });
@@ -77,14 +79,20 @@ describe('CstgDomainAddDomainDialog', () => {
       />
     );
 
+    await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'newDomains');
+
     await submitDialog();
 
-    expect(screen.getByRole('alert')).toBeInTheDocument();
+    expect(
+      screen.getByText('The domains entered are invalid root-level domains', { exact: false })
+    ).toBeInTheDocument();
   });
 
   it('should render error if user types in at least one incorrect domain in a list', async () => {
+    const user = userEvent.setup();
+
     const onAddDomainsMock = jest.fn(() => {
-      return Promise.resolve(['test', 'test2']);
+      return Promise.resolve(['test1', 'test2']);
     });
 
     render(
@@ -95,9 +103,13 @@ describe('CstgDomainAddDomainDialog', () => {
       />
     );
 
+    await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'newDomains');
+
     await submitDialog();
 
-    expect(screen.getByRole('alert')).toBeInTheDocument();
+    expect(
+      screen.getByText('The domains entered are invalid root-level domains', { exact: false })
+    ).toBeInTheDocument();
   });
 
   it('should render error if user submits empty text box for domain names', async () => {
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -21,6 +21,7 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
+        invalidDomains={[]}
       />
     );
 
@@ -49,6 +50,7 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
+        invalidDomains={[]}
       />
     );
 
@@ -83,6 +85,7 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
+        invalidDomains={[]}
       />
     );
 
@@ -104,6 +107,7 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
+        invalidDomains={[]}
       />
     );
 
@@ -124,6 +128,7 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
+        invalidDomains={[]}
       />
     );
     await submitDialog();
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -4,8 +4,8 @@ import userEvent from '@testing-library/user-event';
 import CstgAddDomainDialog from './CstgAddDomainDialog';
 
 const submitDialog = async () => {
-  const createButton = screen.getByRole('button', { name: 'Add Domains' });
-  await userEvent.click(createButton);
+  const submitButton = screen.getByRole('button', { name: 'Add Domains' });
+  await userEvent.click(submitButton);
 };
 
 describe('CstgDomainAddDomainDialog', () => {
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
@@ -63,19 +63,10 @@ describe('CstgDomainAddDomainDialog', () => {
 
     await submitDialog();
 
-    await waitFor(() => {
-      expect(onAddDomainsMock).toHaveBeenCalledWith(
-        ['test.com', 'test2.com', 'test3.com', 'test4.com'],
-        expect.anything()
-      );
-    });
-
     expect(screen.queryByRole('alert')).not.toBeInTheDocument();
   });
 
   it('should render error when user types single incorrect domain', async () => {
-    const user = userEvent.setup();
-
     const onAddDomainsMock = jest.fn(() => {
       return Promise.resolve();
     });
@@ -85,11 +76,9 @@ describe('CstgDomainAddDomainDialog', () => {
         onAddDomains={onAddDomainsMock}
         onOpenChange={() => {}}
         existingDomains={[]}
-        invalidDomains={[]}
+        invalidDomains={['newDomains']}
       />
     );
-
-    await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'test');
 
     await submitDialog();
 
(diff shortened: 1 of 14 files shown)


src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -46,28 +59,9 @@ function CstgAddDomainDialog({
...
-          uniqueDomains[index] = extractTopLevelDomain(newDomain);
-        });
-        // filter for uniqueness (e.g. 2 different domains entered could have the same root-level domain)
-        const dedupedDomains = deduplicateStrings(uniqueDomains);
-        await onAddDomains(dedupedDomains, deleteExistingList);
-      }
+      // filter for uniqueness (e.g. 2 different domains entered could have the same root-level domain)
+      const dedupedDomains = deduplicateStrings(uniqueDomains);
+      await onAddDomains(dedupedDomains, deleteExistingList);
     }
```
ashleysmithTTD, line 65: All the validation can be done on the back-end now

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.spec.ts
```diff
@@ -39,17 +39,6 @@ describe('test domain helper functions', () => {
...
-    it.each(['test.c', 'https://test', '', 't    st.com', 'test'])(
-      'should check the domain is invalid',
-      (stringsList) => {
-        expect(isValidDomain(stringsList)).toEqual(false);
-      }
-    );
-    it.each(validDomainsList)('should return the domain is valid', (stringList) => {
-      expect(isValidDomain(stringList)).toEqual(true);
-    });
-  });
```
ashleysmithTTD, line 52: Because this functionality is done on the backend now, this helper function is no longer needed

src/web/services/domainNamesService.ts
```diff
@@ -26,16 +27,29 @@ export async function GetDomainNames(participantId?: number) {
...
+      e?.response?.status === 400 &&
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      e?.response?.data.message.includes('Invalid Domain Names')
+    ) {
+      return {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
+        domains: [e?.response?.data.message.replace('Invalid Domain Names:', '')],
+        isValidDomains: false,
+      };
+    }
```
ashleysmithTTD, line 51: It was requested by admin people that the error still says "invalid domain names" so still need to filter it out but also it is a good check that the error is actually from invalid domains

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -46,27 +45,16 @@ function CstgAddDomainDialog({
...
-      const invalidDomains: string[] = [];
-      let allValid = true;
-      uniqueDomains.forEach((newDomain) => {
-        if (!isValidDomain(newDomain)) {
-          invalidDomains.push(newDomain);
-          allValid = false;
-        }
-      });
-      if (!allValid) {
+      // filter for uniqueness (e.g. 2 different domains entered could have the same root-level domain)
```
alex-yau-ttd, line 48: I find this comment/naming a bit confusing because we already have `const uniqueDomains` above on line 39. Maybe we could think of a better name to avoid confusion between uniqueness and deduped?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -46,27 +45,16 @@ function CstgAddDomainDialog({
...
+      const dedupedDomains = deduplicateStrings(uniqueDomains);
+      const invalidDomains = await onAddDomains(dedupedDomains, deleteExistingList);
+      if (invalidDomains.length > 0) {
         setError('root.serverError', {
```
alex-yau-ttd, line 52: For my understanding, how does this `root.serverError` work? It's a bit confusing to me as it is not an error from the server 🤔

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
```diff
@@ -46,27 +45,16 @@ function CstgAddDomainDialog({
...
-      let allValid = true;
-      uniqueDomains.forEach((newDomain) => {
-        if (!isValidDomain(newDomain)) {
-          invalidDomains.push(newDomain);
-          allValid = false;
-        }
-      });
-      if (!allValid) {
+      // filter for uniqueness (e.g. 2 different domains entered could have the same root-level domain)
+      const dedupedDomains = deduplicateStrings(uniqueDomains);
```
alex-yau-ttd, line 49: Could we not just do this in admin?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
alex-yau-ttd: I think this method is still pretty low, especially given that the validation can be done on the back-end now. Perhaps it might be worth refactoring this a bit to improve readability? Something like:
```
const onSubmit = async (formData: AddDomainNamesFormProps) => {
  const newDomains = separateStringsList(formData.newDomains);
  const uniqueDomains = getUniqueDomains(newDomains, deleteExistingList, existingDomains);

  if (uniqueDomains.length === 0) {
    handleError('The domains entered already exist.');
    return;
  }

  await handleValidDomains(uniqueDomains, deleteExistingList);
};

const getUniqueDomains = (
  newDomains: string[],
  deleteExistingList: boolean,
  existingDomains?: string[]
): string[] => {
  return deleteExistingList
    ? newDomains
    : newDomains.filter((domain) => !existingDomains?.includes(domain));
};

const handleError = (message: string) => {
  setError('root.serverError', {
    type: '400',
    message: message,
  });
};

const handleValidDomains = async (
  uniqueDomains: string[],
  deleteExistingList: boolean
) => {
  const dedupedDomains = deduplicateStrings(uniqueDomains);
  const invalidDomains = await onAddDomains(dedupedDomains, deleteExistingList);

  if (invalidDomains.length > 0) {
    handleError(
      `The domains entered are invalid root-level domains: ${formatStringsWithSeparator(
        invalidDomains
      )}`
    );
  }
};
```
(you don't need to use that verbatim)

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.stories.tsx
```diff
@@ -12,10 +13,10 @@ type Story = StoryObj<typeof CstgDomainsTable>;
 export const WithDomains: Story = {
   args: {
     domains: ['test.com', 'abc.com', '123.com'],
-    onUpdateDomains: (domains) => {
+    onUpdateDomains: (domains: string[]) => {
```
alex-yau-ttd, line 16: I wonder if we can just simplify this to the following?
```
onUpdateDomains: (domains: string[]) => {
  console.log('update domains to:', domains);
  return Promise.resolve({ domains, isValidDomains: true });
},
```

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -107,18 +111,28 @@ export function CstgDomainsTable({
...
         ...[updatedDomainName],
       ],
       'edited'
     );
-    if (editedDomains) {
-      setPagedDomains(getPagedDomains(editedDomains, pageNumber, rowsPerPage));
-    }
+    const editedDomains = editedDomainsResponse?.domains;
+    const isValidDomains = editedDomainsResponse?.isValidDomains;
+    if (editedDomains)
```
alex-yau-ttd, line 128: We could simplify this to:
```
if (editedDomains && isValidDomains) {
  setPagedDomains(getPagedDomains(editedDomains, pageNumber, rowsPerPage));
  return true;
}
return false;
```

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
```diff
@@ -132,18 +146,26 @@ export function CstgDomainsTable({
...
-    if (newDomains) {
-      setPageNumber(initialPageNumber);
-      setRowsPerPage(initialRowsPerPage);
-      setPagedDomains(getPagedDomains(newDomains, initialPageNumber, initialRowsPerPage));
-      setSearchedDomains(newDomains);
+  ): Promise<string[]> => {
+    const newDomainsResponse = await onAddDomains(newDomainsFormatted, deleteExistingList);
+    const newDomains = newDomainsResponse?.domains;
+    const isValidDomains = newDomainsResponse?.isValidDomains;
+    if (isValidDomains) {
```
alex-yau-ttd, line 153: Perhaps the following would be slightly more readable?
```
if (isValidDomains) {
  setShowAddDomainsDialog(false);
  setSearchedDomains(domains);
  setSelectedDomains([]);
  setSearchText('');
  if (newDomains) {
    setPageNumber(initialPageNumber);
    setRowsPerPage(initialRowsPerPage);
    setPagedDomains(getPagedDomains(newDomains, initialPageNumber, initialRowsPerPage));
    setSearchedDomains(newDomains);
  }
  return [];
}
if (newDomains) {
  return newDomains;
}
return [];
```

src/web/utils/textHelpers.ts
```diff
@@ -26,8 +26,9 @@ export function formatUnixDate(timeValue: number) {
 
 export const separateStringsList = (strings: string): string[] => {
   if (strings === '') return [];
-  const stringsTrimmed = strings.replace(/, |,| {2}|\n|;/gi, ' ').trim();
-  return stringsTrimmed.split(/[ ]+/);
+  const stringsTrimmed = strings.replace(/, |,| {2}|\n|;|]|\[/gi, ' ').trim();
+  const ts = stringsTrimmed.split(/[ ]+/);
```
alex-yau-ttd, line 30: This is because the result we are getting from admin is like a fake array? i.e. `"[abc.com, 123.com]"`? Could we just change admin so that it returns a proper string array? 

If not, perhaps we could rename `ts` to something that is a bit more clear? This method is now separating the string and trimming the brackets, maybe it deserves it's own method or a new name?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
```diff
@@ -86,17 +77,14 @@ describe('CstgDomainAddDomainDialog', () => {
       />
     );
 
-    await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'test');
-
     await submitDialog();
 
     expect(screen.getByRole('alert')).toBeInTheDocument();
```
alex-yau-ttd, line 82: I believe this test isn't actually testing the right case here, the alert it is producing seems to be the empty input case, rather than the invalid domain case. i.e. If I add this line: `expect(screen.getByText('Please specify domains.')).toBeInTheDocument();` it still passes. 

I think we need to do something similar to what we did in the happy case, i.e. use `user.type(...)` in order to set this up properly - perhaps the `onAddDomainsMock` is confusing.

If we do that, then we should be able to assert on the correct error text.

So I think this test should become:
```
it('should render error when user types single incorrect domain', async () => {
  const user = userEvent.setup();

  const onAddDomainsMock = jest.fn(() => {
    return Promise.resolve(['newDomains']);
  });

  render(
    <CstgAddDomainDialog
      onAddDomains={onAddDomainsMock}
      onOpenChange={() => {}}
      existingDomains={[]}
    />
  );

  await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'newDomains');

  await submitDialog();

  expect(
    screen.getByText('The domains entered are invalid root-level domains', { exact: false })
  ).toBeInTheDocument();
});
```

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
```diff
@@ -61,21 +61,12 @@ describe('CstgDomainAddDomainDialog', () => {
...
     await submitDialog();
 
-    await waitFor(() => {
-      expect(onAddDomainsMock).toHaveBeenCalledWith(
-        ['test.com', 'test2.com', 'test3.com', 'test4.com'],
-        expect.anything()
-      );
-    });
-
     expect(screen.queryByRole('alert')).not.toBeInTheDocument();
```
alex-yau-ttd, line 64: We should probably assert on something stronger here - e.g. if the dialog is not submitted, this test will still pass.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
```diff
@@ -61,21 +61,12 @@ describe('CstgDomainAddDomainDialog', () => {
 
     await submitDialog();
 
-    await waitFor(() => {
```
alex-yau-ttd, line 64: What's the reason for removing this?

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
alex-yau-ttd: I think this also applies to the below test: `should render error if user types in at least one incorrect domain in a list`

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
ashleysmithTTD: good catch, much simpler

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: In admin, duplicate domains will throw an error - but functionality-wise, on the front end it is not an error.  When the add domains dialog was being built, we discussed de-depuplicating automagically for the user if they enter duplicates and still have it be a successful add (unless every domain they enter already exists).  We could maybe do it in admin, it would just require admin code changes/approvals.  Lmk what you think

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: This is the documentation for setError with React Hook Forms https://react-hook-form.com/docs/useform/seterror where the type is 'root.serverError' - these types of errors are good for resetting on each form submission

alex-yau-ttd: thanks for linking that! what exactly do you mean by `root.serverError` being good for resetting on each form submission?

By reading the docs, I would have thought that we should be using the single error case, something like:
```
setError('newDomains', {
  type: 'manual',
  message: 'The domains entered already exist.',
});
```

The `root.serverError` examples seem to be genuine API calls that use the status code from the response from the backend. 

I do recognise that there is a slightly different UX in that the error appears underneath the input rather than at the top of the form, but I think that's fine: 
image

Am I missing something else?

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.tsx
ashleysmithTTD: Better fix - thanks!

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
ashleysmithTTD: Removed this because the change from URL to root-level domain is no longer happening on the front-end, but on the backend instead, so this test was originally to make sure all domains sent to the back-end were root-level (i.e. http://test4.com turning to test4.com ).  I'm thinking maybe we don't even need to test altogether anymore

src/web/utils/textHelpers.ts
ashleysmithTTD: Completely missed deleting this - I was just using it as a test at one point. Thank you for catching. I added the two brackets as replace values which I think still applies to this function in general.  I don't think we should change admin - the reviewer requested that it be in the format that it is, so it does error with the array on the backend.

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
ashleysmithTTD: I can take this test out (expect alert to not be in the document) - but is the test above checking what it was called with should be sufficient?  I was looking at the tests in KeyCreationDialog.spec file for testing the creation of a successful key and it looks like the "toHaveBeenCalledWith" test is all that is used.   please let me know if there is a better/different solution, thanks!

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: New ticket will be created to address this

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.tsx
ashleysmithTTD: New ticket will be created to fix this - not broken, just switching front end to back end

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
ashleysmithTTD: New ticket to be created for pattern changes around testing at a higher level

src/web/utils/textHelpers.ts
ashleysmithTTD: Put it back

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
ashleysmithTTD: Put it back

src/web/components/ClientSideTokenGeneration/CstgAddDomainDialog.spec.tsx
```diff
@@ -95,9 +108,13 @@ describe('CstgDomainAddDomainDialog', () => {
       />
     );
 
+    await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'newDomains');
```
alex-yau-ttd, line 111: nit: perhaps we could make the input and test case clearer here by doing something like:
```
await user.type(screen.getByRole('textbox', { name: 'newDomains' }), 'invalid domain');
```

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.ts
```diff
@@ -26,3 +26,15 @@ export const getPagedDomains = (
       index < (pageNumber - 1) * rowsPerPage + rowsPerPage
   );
 };
+
+export const getUniqueDomains = (
```
alex-yau-ttd, line 30: Is it worth adding tests for this? Answer might be no 😄

src/web/utils/textHelpers.ts
```diff
@@ -26,7 +26,7 @@ export function formatUnixDate(timeValue: number) {
 
 export const separateStringsList = (strings: string): string[] => {
   if (strings === '') return [];
-  const stringsTrimmed = strings.replace(/, |,| {2}|\n|;/gi, ' ').trim();
+  const stringsTrimmed = strings.replace(/, |,| {2}|\n|;|]|\[/gi, ' ').trim();
```
alex-yau-ttd, line 29: Do we need to add a test case in `textHelpers.spec.ts` to cover this?

alex-yau-ttd (approved):

PR #436, opened 2024-05-17
Update Home screen to use loader data and shared loading component.

lionell-pack-ttd:
Update Home screen to use loader data and shared loading component.
Minor changes to avoid content reflow on main screen.

Before: 
chrome_XtWLfzXjH1

After:
chrome_K3BB7fDe8u

### DIFF
--- a/src/web/components/Home/DocumentationCard.tsx
+++ b/src/web/components/Home/DocumentationCard.tsx
@@ -3,7 +3,7 @@ import './DocumentationCard.scss';
 function DocumentationCard() {
   return (
     <div className='document-card'>
-      <img src='/document.svg' alt='document' />
+      <img src='/document.svg' alt='document' width='50px' />
       <div>
         <h2>Documentation</h2>
         <span>Check out documentation for implementation resources</span>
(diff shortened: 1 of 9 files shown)


src/web/components/Home/DocumentationCard.tsx
```diff
@@ -3,7 +3,7 @@ import './DocumentationCard.scss';
 function DocumentationCard() {
   return (
     <div className='document-card'>
-      <img src='/document.svg' alt='document' />
+      <img src='/document.svg' alt='document' width='50px' />
```
alex-yau-ttd, line 6: nit: perhaps this width styling should live in `DocumentationCard.scss` for consistency?

alex-yau-ttd (approved):

src/web/components/Home/DocumentationCard.tsx
lionell-pack-ttd: Usually I'd agree, but I think because the size is tied to the image file (which is referenced here), it should be here. I'd like to find a plugin to automatically add these at some point.

PR #437, opened 2024-05-20
Introduce an async error pattern for customizing loader error display.

lionell-pack-ttd:
Some refactoring to support async front-end errors. Notably, renamed our useAsyncError function because the name colided with a framework function and I wanted to avoid confusion.

Introduce an async error pattern for customizing display of errors that occur during route loaders and use this for the home screen.

Full route failure, default behaviour:
image

Custom async handler (i.e. home.tsx):
image

### DIFF
--- a/src/api/helpers/errorHelpers.ts
+++ b/src/api/helpers/errorHelpers.ts
@@ -1,0 +1,20 @@
+import { Request, Response } from 'express';
+
+import { getTraceId } from './loggingHelpers';
+
+export function errorResponse(
+  req: Request,
+  res: Response,
+  errorCode: number,
+  response?: { message: string } & Record<string, unknown>
+) {
+  const traceId = getTraceId(req);
+
+  return res.status(errorCode).json({
+    errorHash: traceId,
+    ...response,
+  });
+}
+
+export const siteIdNotSetError = (req: Request, res: Response) =>
+  errorResponse(req, res, 400, { message: 'Site id is not set' });
--- a/src/api/routers/participants/participantsKeyPairs.spec.ts
+++ b/src/api/routers/participants/participantsKeyPairs.spec.ts
@@ -78,7 +78,6 @@ describe('#getParticipantKeyPairs', () => {
     await getParticipantKeyPairs(participantRequest, res);
 
     expect(res.status).toHaveBeenLastCalledWith(400);
-    expect(res.send).toHaveBeenLastCalledWith('Site id is not set');
   });
 
   test.each([
--- a/src/api/routers/participants/participantsKeyPairs.ts
+++ b/src/api/routers/participants/participantsKeyPairs.ts
@@ -1,12 +1,13 @@
 import { Response } from 'express';
 
+import { siteIdNotSetError } from '../../helpers/errorHelpers';
 import { getKeyPairsList } from '../../services/adminServiceClient';
 import { ParticipantRequest } from '../../services/participantsService';
 
 export async function getParticipantKeyPairs(req: ParticipantRequest, res: Response) {
   const { participant } = req;
   if (!participant?.siteId) {
-    return res.status(400).send('Site id is not set');
+    return siteIdNotSetError(req, res);
   }
   const siteId = participant?.siteId;
   const allKeyPairs = await getKeyPairsList(siteId!);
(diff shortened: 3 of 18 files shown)


src/api/routers/participants/participantsKeyPairs.spec.ts
```diff
@@ -78,7 +78,6 @@ describe('#getParticipantKeyPairs', () => {
     await getParticipantKeyPairs(participantRequest, res);
 
     expect(res.status).toHaveBeenLastCalledWith(400);
-    expect(res.send).toHaveBeenLastCalledWith('Site id is not set');
```
lionell-pack-ttd, line 81: This wasn't very useful.

src/api/routers/participants/participantsKeyPairs.ts
```diff
@@ -1,12 +1,13 @@
 import { Response } from 'express';
 
+import { siteIdNotSetError } from '../../helpers/errorHelpers';
 import { getKeyPairsList } from '../../services/adminServiceClient';
 import { ParticipantRequest } from '../../services/participantsService';
 
 export async function getParticipantKeyPairs(req: ParticipantRequest, res: Response) {
   const { participant } = req;
   if (!participant?.siteId) {
-    return res.status(400).send('Site id is not set');
```
lionell-pack-ttd, line 9: So many magic strings :( this also means our error code for this error is in one place.

src/web/components/Core/EnvironmentBanner.tsx
```diff
@@ -2,15 +2,15 @@ import { useEffect, useState } from 'react';
 
 import { EnvironmentVariable, GetEnvironmentVariables } from '../../services/environmentVariables';
 import { ApiError } from '../../utils/apiError';
-import { useAsyncError } from '../../utils/errorHandler';
```
lionell-pack-ttd, line 5: I renamed this because useAsyncError collided with useAsyncError from react router.

src/web/components/Core/ErrorView.tsx
```diff
@@ -1,23 +1,36 @@
...
+export function ErrorView({ message, errorId, errorHash }: ErrorViewProps) {
   return (
     <div className='error-content'>
       <img alt='Error icon' src='/uid2-logo.png' />
-      <div>Error</div>
-      <div>
-        {message ??
-          'There was an unexpected error. Please try again. If the problem persists, contact Support and provide the following information: '}
-      </div>
+      <div>There was an unexpected error. Please try again.</div>
```
lionell-pack-ttd, line 17: Some minor changes here - I think they make a little more sense. It's probably sensible to always provide the stock text even if we provide a custom message.

src/web/utils/errorHelpers.ts
```diff
@@ -10,15 +10,15 @@ export const analyticsIdentifier = (errorId?: string, errorHash?: string) => {
 export const extractMessageFromAxiosError = (err: Error) => {
   // message does not exist on the data field for Axios Error.Response, but we add it
   // in our API for custom messages
-  let message = null;
   if (axios.isAxiosError(err)) {
     if (err.response?.data?.message) {
```
lionell-pack-ttd, line 14: Just updating this so the return type is string|undefined (instead of string|null) so I can use it in component props.

src/api/helpers/errorHelpers.ts
alex-yau-ttd, line 1: We already have a file called `errorHelpers.ts` in `src/web/utils` - what do you think about renaming this to avoid confusion?

src/api/routers/participants/participantsRouter.ts
```diff
@@ -12,6 +12,7 @@ import {
   ParticipantStatus,
 } from '../../entities/Participant';
 import { UserDTO, UserRole } from '../../entities/User';
+import { errorResponse, siteIdNotSetError } from '../../helpers/errorHelpers';
```
alex-yau-ttd, line 15: probably worth fixing this :D

alex-yau-ttd (approved):

src/api/routers/participants/participantsRouter.ts
lionell-pack-ttd: Done :)

PR #443, opened 2024-05-24
Move several nav routes to top nav menu

alex-yau-ttd:
- Rename Account Information route, components and files
- Comment out dark mode switch
- Hide several links from appearing on dashboard. Instead, show them in the top nav.
- Update styling for portal header dropdown

https://github.com/IABTechLab/uid2-self-serve-portal/assets/137864392/7da58ccd-b2b1-4f04-b6a7-c530e3e31cc8

PR #427, opened 2024-05-06
Upgrade stories to CSF3

alex-yau-ttd:
- Run Storybook's migration to [Component Story Format 3](https://storybook.js.org/blog/storybook-csf3-is-here/), i.e. `npx storybook@latest migrate csf-2-to-3 --glob="**/*.stories.tsx" --parser=tsx`
- Make some adjustments to remove unused imports
- Follow existing pattern of strongly typed Storybook Meta, rather than exporting default with `as  Meta<typeof MyType>`
- Use `StoryObj` instead of `StoryFn` for simple cases
- Simplify some stories by using the spread operator for arguments.

PR #444, opened 2024-05-24
New dashboard notifications

ashleysmithTTD:
What Changed:

- Banner on home page when user has no email contacts
- Box warning notification on home page if at least one key needs rotating (older than one year)
- Warning sign on keys in API keys table for the keys that need rotating

Test Plan:
- Delete all email contacts and check for banner
- Add email contact back and make sure banner disappears
- Test if only one API key needs rotating (can change the time to be less than a year to test) 
- Test if multiple API keys need rotating
- Test when zero API keys need rotating 
- Test when at least one API key needs rotating and email contacts need adding that page still looks good
- When clicking 'Rotate API Key' button, it should direct to API Key Management


src/web/styles/tables.scss
```diff
@@ -51,7 +51,6 @@ table {
     display: flex;
     justify-content: flex-end;
     flex-wrap: wrap;
-    align-items: baseline;
```
ashleysmithTTD, line 54: Removed so the warning sign lines up with the other buttons (edit and trash)

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const needsRotating = (apiKey: ApiKeyDTO): boolean => {
+  const currentDate = new Date().getTime();
+  // apiKey date is in seconds, current date is in milliseconds
+  const currentDateFormat = Math.floor(currentDate / 1000);
+  if (currentDateFormat - apiKey.created > yearInSeconds) {
+    return true;
+  }
+  return false;
+};
```
ashleysmithTTD, line 13: Went to add a test file but unsure how to test a function like this that uses current date that is always changing

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiKeyDTO } from '../../../api/services/adminServiceHelpers';
+
+const yearInSeconds = 31556952;
```
alex-yau-ttd, line 3: Doesn't really matter, but isn't the number of seconds in an average year 60*60*24*365.25=31557600? Or 31536000 ignoring leap days. maybe we could just use this so it's less magic: `const yearInSeconds = 60 * 60 * 24 * 365.25;`

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const needsRotating = (apiKey: ApiKeyDTO): boolean => {
+  const currentDate = new Date().getTime();
```
alex-yau-ttd, line 6: I think we can just use `Date.now()` instead, it's a bit more concise.

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiKeyDTO } from '../../../api/services/adminServiceHelpers';
+
+const yearInSeconds = 31556952;
+
+export const needsRotating = (apiKey: ApiKeyDTO): boolean => {
```
alex-yau-ttd, line 5: I'd recommend a more informative name, such as `shouldRotateApiKey`

src/web/components/ApiKeyManagement/KeyHelper.ts
alex-yau-ttd: Should be possible - I think we can just build in enough tolerance to the test. e.g. for the case where the key is older than a year, just make the key 366 days old:
```
/* eslint-disable camelcase */
import { ApiRoleDTO } from '../../../api/entities/ApiRole';
import { ApiKeyDTO } from '../../../api/services/adminServiceHelpers';
import { shouldRotateApiKey } from './KeyHelper';

const Mapper: ApiRoleDTO = { id: 1, roleName: 'MAPPER', externalName: 'Mapper', order: 1 };
const Generator: ApiRoleDTO = {
  id: 2,
  roleName: 'GENERATOR',
  externalName: 'Generator',
  order: 2,
};
const Bidder: ApiRoleDTO = {
  id: 3,
  roleName: 'ID_READER',
  externalName: 'Bidder',
  order: 4,
};
const Sharer: ApiRoleDTO = { id: 4, roleName: 'SHARER', externalName: 'Sharer', order: 3 };

const buildApiKey = (apiKey: Partial<ApiKeyDTO>): ApiKeyDTO => {
  return {
    contact: apiKey.contact ?? 'ApiKey',
    name: apiKey.name ?? 'ApiKey',
    created: apiKey.created ?? 1702830516,
    key_id: apiKey.key_id ?? 'F4lfa.fdas',
    site_id: apiKey.site_id ?? 1,
    disabled: apiKey.disabled ?? false,
    roles: apiKey.roles ?? [Mapper, Generator, Bidder, Sharer],
    service_id: apiKey.service_id ?? 0,
  };
};

describe('does api key need rotating', () => {
  const yearInSeconds = 31556952;
  const dayInSeconds = 60 * 60 * 24;

  it('should return true if the API key is older than one year', () => {
    const apiKey = { created: Math.floor(Date.now() / 1000) - yearInSeconds - dayInSeconds };
    expect(shouldRotateApiKey(buildApiKey(apiKey))).toBe(true);
  });

  it('should handle the case where the API key was just created', () => {
    const apiKey = { created: Math.floor(Date.now() / 1000) };
    expect(shouldRotateApiKey(buildApiKey(apiKey))).toBe(false);
  });
});
```

Note there's some code here that should be extracted out to a new file and used elsewhere in the codebase, but that should be a good start anyway.

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -42,6 +50,13 @@ function KeyItem({ apiKey: apiKeyInitial, onEdit, onDisable, availableRoles }: K
       {availableRoles.length > 0 && (
         <td className='action'>
           <div className='action-cell'>
+            {keyNeedsRotating && (
+              <FontAwesomeIcon
+                icon='triangle-exclamation'
+                title='We recommend you rotate your API key every year.'
```
alex-yau-ttd, line 56: Perhaps we could use a `<Tooltip>` here for the message rather than relying on the title?

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -42,6 +50,13 @@ function KeyItem({ apiKey: apiKeyInitial, onEdit, onDisable, availableRoles }: K
...
+                className='warning-button'
```
alex-yau-ttd, line 57: nit: perhaps `warning-icon` rather than `warning-button` given it's not a button

src/web/components/Home/RotateAPIKeysCard.scss
alex-yau-ttd, line 1: nit: we typically use `Api` rather than `API` for file names, types, etc. - perhaps we should remain consistent

src/web/components/Home/RotateAPIKeysCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+function RotateApiKeysCard({ apiKeysToRotate }: RotateApiKeysCardProps) {
+  const getDatesCreated = () => {
+    const datesCreated = apiKeysToRotate.map((apiKey) => formatUnixDate(apiKey.created).toString());
+    const datesCreatedString = formatStringsWithSeparator(deduplicateStrings(datesCreated));
+    return datesCreatedString;
+  };
+
+  return (
+    <div className='rotate-api-keys-card'>
```
alex-yau-ttd, line 25: I feel like this should use the same shared component as `<DocumentationCard>`. Perhaps they can both use `<Card>`?What do you think?

src/web/components/Home/RotateAPIKeysCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <span>
+          {`Your API key${
+            apiKeysToRotate.length > 1 ? 's were' : ' was'
+          } created on ${getDatesCreated()}. We recommend you rotate your API key${
+            apiKeysToRotate.length > 1 ? 's' : ''
+          } every year.`}{' '}
+        </span>
+      </div>
+
+      <Link to='/dashboard/apiKeys'>
```
alex-yau-ttd, line 38: I'm wondering if we should actually import `ApiKeyManagementRoute` then use `<Link to={ApiKeyManagementRoute.path}>`. That's the approach I used here: https://github.com/IABTechLab/uid2-self-serve-portal/pull/443/files. Happy to discuss if that seems wrong though

src/web/screens/home.scss
alex-yau-ttd, line 1: It probably makes more sense to have `.dashboard-cards` inside `dashboard-cards-container`:|
```
.dashboard-cards-container {
  ...

  .dashboard-cards {
  ...
  }
}
```

src/web/screens/home.tsx
```diff
@@ -59,18 +82,51 @@ function Home() {
...
-          </AwaitTypesafe>
-        </Suspense>
-        <DocumentationCard />
+        <div>
+          <Suspense fallback={<Loading />}>
+            <AwaitTypesafe resolve={data.apiKeysToRotate} errorElement={<AsyncErrorView />}>
+              {(apiKeysToRotate) =>
+                apiKeysToRotate.length > 0 && (
+                  <div className='dashboard-cards'>
+                    <RotateApiKeysCard apiKeysToRotate={apiKeysToRotate} />{' '}
```
alex-yau-ttd, line 107: I don't think we need the `{' '}`

src/web/screens/home.tsx
```diff
@@ -59,18 +82,51 @@ function Home() {
...
-            )}
-          </AwaitTypesafe>
-        </Suspense>
-        <DocumentationCard />
+        <div>
+          <Suspense fallback={<Loading />}>
+            <AwaitTypesafe resolve={data.apiKeysToRotate} errorElement={<AsyncErrorView />}>
+              {(apiKeysToRotate) =>
+                apiKeysToRotate.length > 0 && (
+                  <div className='dashboard-cards'>
```
alex-yau-ttd, line 106: The className is `dashboard-cards` but it only have seems to be used for an individual card - should we just call it `dashboard-card`? It'd also be nice if we didn't need this className - if you could find a way to do that? e.g. We already know these are cards within `dashboard-cards-container`.

src/web/components/Home/RotateAPIKeysCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+function RotateApiKeysCard({ apiKeysToRotate }: RotateApiKeysCardProps) {
+  const getDatesCreated = () => {
+    const datesCreated = apiKeysToRotate.map((apiKey) => formatUnixDate(apiKey.created).toString());
+    const datesCreatedString = formatStringsWithSeparator(deduplicateStrings(datesCreated));
+    return datesCreatedString;
+  };
+
+  return (
+    <div className='rotate-api-keys-card'>
+      <FontAwesomeIcon icon='triangle-exclamation' className='warning-button' />
```
alex-yau-ttd, line 26: We already have `exclamation-triangle` 😂 Let's use that instead. Then we can remove the changes in `configureFontAwesomeLibrary.ts`

src/web/components/ApiKeyManagement/KeyItem.scss
```diff
@@ -0,0 +1,6 @@
+.warning-button {
+  color: var(--theme-warning);
+  border-radius: 5px;
```
alex-yau-ttd, line 3: Not sure if we need this `border-radius`?

src/web/components/ApiKeyManagement/KeyItem.tsx
```diff
@@ -18,6 +21,7 @@ function KeyItem({ apiKey: apiKeyInitial, onEdit, onDisable, availableRoles }: K
   const [apiKey, setApiKey] = useState<ApiKeyDTO>(apiKeyInitial);
   const [showKeyDisableDialog, setShowKeyDisableDialog] = useState<boolean>(false);
   const [showKeyEditDialog, setShowKeyEditDialog] = useState<boolean>(false);
+  const [keyNeedsRotating, setKeyNeedsRotating] = useState<boolean>(false);
```
alex-yau-ttd, line 24: maybe `showKeyNeedsRotatingWarning` would be more consistent with the other state variables? Or something similar.

src/web/components/Home/RotateAPIKeysCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <h2>Rotate Your API Key{apiKeysToRotate.length > 1 && 's'}</h2>
+      <div className='rotate-api-keys-text'>
+        <span>
+          {`Your API key${
+            apiKeysToRotate.length > 1 ? 's were' : ' was'
+          } created on ${getDatesCreated()}. We recommend you rotate your API key${
+            apiKeysToRotate.length > 1 ? 's' : ''
+          } every year.`}{' '}
```
alex-yau-ttd, line 34: Don't think we need the `{' '}`?

src/web/components/ApiKeyManagement/KeyHelper.ts
ashleysmithTTD: Agreed - writing out the math makes more sense for anyone looking at it later

src/web/components/ApiKeyManagement/KeyItem.tsx
ashleysmithTTD: Agreed

src/web/components/ApiKeyManagement/KeyHelper.ts
ashleysmithTTD: Yep looks like they output the same number but Date.now() doesn't require the new keyword

src/web/components/ApiKeyManagement/KeyItem.tsx
ashleysmithTTD: Chagned - at first I thought it would end up being a button if/when we rotate keys directly from the portal but now that I think about it, the button would probably not be the warning icon

src/web/components/ApiKeyManagement/KeyHelper.ts
ashleysmithTTD: Thanks - this was super helpful.  I've exported a few things and added two more test cases

src/web/components/Home/RotateAPIKeysCard.tsx
ashleysmithTTD: Yep - changed all 3 components to use Card as their outer tag instead of <div> and <RotateApiKeysCard> and <DocumentationCard> both didn't need their own class names in that outer tag, the css for Card takes care of it.

src/web/components/ApiKeyManagement/KeyItem.scss
ashleysmithTTD: Yep thanks!

src/web/components/Home/RotateAPIKeysCard.tsx
ashleysmithTTD: What is the difference between the two - what makes the route a better solution?  I'm open to either, we should just have consistency because I took this example from `SharingPermissionsCard.tsx` so if we change this pattern, I'll make sure to change in both places

src/web/screens/home.tsx
ashleysmithTTD: The reason I have both is to match the UX design where two cards are vertical ( `RotateApiKeysCard` and `SharingPermissionsCard`) and `DocumentationCard` is horizontal to those two.  I think we do need both the achieve this design, however I did remove some duplicated code between the two that isn't needed in the actual card ( `display: flex` and `gap: 22px`)

src/web/components/Home/RotateAPIKeysCard.tsx
ashleysmithTTD: Ooh yes I did see that - I was just confused because on the FontAwesome webiste, the icon shows `triangle-exclamation`.  I agree we should only have one, but should I change the ones that use `exclamation-triangle` to `triangle-exclamation`.  I'm not sure how they both work, but maybe the order doesn't matter?

src/web/components/ApiKeyManagement/KeyItem.tsx
ashleysmithTTD: True - changed

src/web/components/Home/RotateAPIKeysCard.tsx
alex-yau-ttd: Just to avoid having magic strings lying around. In the unlikely event that we change the path for the `ApiKeyManagementRoute`, we would have to change it in several places if they are hardcoded. Whereas if we used `ApiKeyManagementRoute.path` we'd only have to update it at the source.

I can't think of any downside...

src/web/components/Home/RotateAPIKeysCard.tsx
alex-yau-ttd: yeah good point. I now see that [exclamation-triangle](https://fontawesome.com/v5/icons/exclamation-triangle?f=classic&s=solid) is version 5 and [triangle-exclamation](https://fontawesome.com/icons/exclamation-triangle?f=classic&s=solid) is version 6. We may as well use the latest version, so I'd suggest changing all of them to `triangle-exclamation` and removing `exclamation-triangle` :)

src/web/components/ApiKeyManagement/KeyHelper.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+describe('test if api key needs rotating', () => {
+  const dayInSeconds = 60 * 60 * 24;
+
+  it('should return true if the API key is older than one year', () => {
+    const apiKey = { created: Math.floor(Date.now() / 1000) - yearInSeconds - dayInSeconds };
+    expect(shouldRotateApiKey(buildApiKey(apiKey))).toBe(true);
+  });
+
+  it('should return true if the API key is much older than one year - created 3 years ago', () => {
```
alex-yau-ttd, line 44: nit: I suggest not using `-` alongside code that uses that as a minus sign, to avoid confusion :)

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -1,12 +1,12 @@
 import { ApiKeyDTO } from '../../../api/services/adminServiceHelpers';
 
-const yearInSeconds = 31556952;
+export const yearInSeconds = 60 * 60 * 24 * 365.25;
 
-export const needsRotating = (apiKey: ApiKeyDTO): boolean => {
-  const currentDate = new Date().getTime();
+// api key is recommended to rotate if older than one year
```
alex-yau-ttd, line 5: I think we can get rid of this comment - the tests should describe the behaviour clearly enough. Otherwise if we change the expiry time, then we'd need to update this comment as well.

src/web/components/ApiKeyManagement/KeyHelper.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  id: 3,
+  roleName: 'ID_READER',
+  externalName: 'Bidder',
+  order: 4,
+};
+const Sharer: ApiRoleDTO = { id: 4, roleName: 'SHARER', externalName: 'Sharer', order: 3 };
+
+const buildApiKey = (apiKey: Partial<ApiKeyDTO>): ApiKeyDTO => {
+  return {
+    contact: apiKey.contact ?? faker.string.alpha(),
```
alex-yau-ttd, line 25: Actually maybe `faker.string.alphanumeric()` might be more accurate for contact, name and key_id.

src/web/components/ApiKeyManagement/KeyHelper.ts
```diff
@@ -1,12 +1,12 @@
...
+export const shouldRotateApiKey = (apiKey: ApiKeyDTO): boolean => {
```
alex-yau-ttd, line 6: sorry - I'm now thinking that this method could include the `disabled` check, rather than doing the two checks inside `getApiKeysToRate` in `home.tsx` - that way all the logic lives in the same place and we can more easily cover with tests & stories. Does that seem reasonable?

src/web/components/ApiKeyManagement/KeyHelper.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+const Bidder: ApiRoleDTO = {
+  id: 3,
+  roleName: 'ID_READER',
+  externalName: 'Bidder',
+  order: 4,
+};
+const Sharer: ApiRoleDTO = { id: 4, roleName: 'SHARER', externalName: 'Sharer', order: 3 };
+
+const buildApiKey = (apiKey: Partial<ApiKeyDTO>): ApiKeyDTO => {
```
alex-yau-ttd, line 23: Sorry - we should probably call this `createApiKeyFake()`. We can then export it and use it in the `RotateApiKeyCard` story (unless you wanted to do that in a separate ticket/PR).

src/web/components/Home/RotateApiKeyCard.stories.tsx
alex-yau-ttd, line 1: This is using CSF2 (i.e. using `ComponentMeta` and `ComponentStory`). I've got a [PR](https://github.com/IABTechLab/uid2-self-serve-portal/pull/427/files) to move everything to CSF3 which I plan to merge today, so I'll provide the recommended code for this file, as a starting point. Note that it involves renaming, exporting, and slightly changing the `buildApiKey` method from `KeyHelper.spec.ts`.

```
/* eslint-disable camelcase */
import { Meta, StoryObj } from '@storybook/react';

import { ApiRoleDTO } from '../../../api/entities/ApiRole';
import { createApiKeyFake } from '../ApiKeyManagement/KeyHelper.spec';
import RotateApiKeyCard from './RotateApiKeyCard';

const Mapper: ApiRoleDTO = { id: 1, roleName: 'MAPPER', externalName: 'Mapper', order: 1 };
const Generator: ApiRoleDTO = {
  id: 2,
  roleName: 'GENERATOR',
  externalName: 'Generator',
  order: 2,
};
const Bidder: ApiRoleDTO = {
  id: 3,
  roleName: 'ID_READER',
  externalName: 'Bidder',
  order: 4,
};
const Sharer: ApiRoleDTO = { id: 4, roleName: 'SHARER', externalName: 'Sharer', order: 3 };

const meta: Meta<typeof RotateApiKeyCard> = {
  title: 'Home/Rotate API Key Card',
  component: RotateApiKeyCard,
};
export default meta;

type Story = StoryObj<typeof RotateApiKeyCard>;

export const Default: Story = {
  args: {
    apiKeysToRotate: [createApiKeyFake()],
  },
};

export const MutipleKeysToRotate: Story = {
  args: {
    apiKeysToRotate: [createApiKeyFake(), createApiKeyFake()],
  },
};
```

alex-yau-ttd: Hmm actually I don't think storybook likes importing from spec files - perhaps we need to extract to a separate file or potentially have two different methods. 

At a minimum, let's at least change it to use CSF3:
```
/* eslint-disable camelcase */
import { Meta, StoryObj } from '@storybook/react';

import RotateApiKeyCard from './RotateApiKeyCard';

const meta: Meta<typeof RotateApiKeyCard> = {
  title: 'Home/Rotate API Key Card',
  component: RotateApiKeyCard,
};
export default meta;

type Story = StoryObj<typeof RotateApiKeyCard>;

const testApiKey1 = {
  contact: 'ApiKey1',
  name: 'ApiKey1',
  created: 1672597393,
  key_id: 'F4lfa.fdas',
  site_id: 1,
  disabled: false,
  roles: [
    { id: 1, roleName: 'MAPPER', externalName: 'Mapper', order: 1 },
    { id: 2, roleName: 'GENERATOR', externalName: 'Generator', order: 2 },
  ],
  service_id: 0,
};

const testApiKey2 = {
  contact: 'ApiKey2',
  name: 'ApiKey2',
  created: 1675362193,
  key_id: 'FDSL,089',
  disabled: false,
  roles: [{ id: 1, roleName: 'MAPPER', externalName: 'Mapper', order: 1 }],
  service_id: 0,
  site_id: 1,
};

export const Default: Story = {
  args: {
    apiKeysToRotate: [testApiKey1],
  },
};

export const MutipleKeysToRotate: Story = {
  args: {
    apiKeysToRotate: [testApiKey1, testApiKey2],
  },
};
```

Perhaps we can do the other suggestions separately.

src/web/screens/home.tsx
```diff
@@ -64,7 +64,9 @@ async function getEmailContacts() {
 
 async function getApiKeysToRotate() {
   const apiKeys = await GetParticipantApiKeys();
-  return apiKeys.filter((apiKey) => apiKey.disabled === false && needsRotating(apiKey) === true);
+  return apiKeys.filter(
+    (apiKey) => apiKey.disabled === false && shouldRotateApiKey(apiKey) === true
```
alex-yau-ttd, line 68: Per my other comments - perhaps best we move the `disabled` check inside `shouldRotateApiKey()`?

alex-yau-ttd (approved):

src/web/components/Home/RotateAPIKeysCard.tsx
ashleysmithTTD: That makes sense - changed!

src/web/components/ApiKeyManagement/KeyHelper.ts
ashleysmithTTD: Yep sounds good - I've added two more test cases to check the disabled key as well

src/web/components/Home/RotateApiKeyCard.stories.tsx
ashleysmithTTD: Changed!

PR #448, opened 2024-05-29
Edit and Delete button new components

ashleysmithTTD:


src/web/components/Core/DeleteButton.tsx
alex-yau-ttd, line 1: There's a lot of shared code between this and EditButton - there's potential to go further with this and create a generic <Button> component that extracts these things even more. But maybe that's a bit overkill, so maybe keep as-is - though I'm curious to know what you think.

alex-yau-ttd (approved):

alex-yau-ttd:
Let's fix the styling inconsistencies with the action buttons

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
```diff
@@ -25,10 +25,9 @@
       display: flex;
       justify-content: flex-end;
       flex-wrap: wrap;
-    }
 
-    .action-cell > button {
-      padding: 5px 0 5px 10px;
+      
+
```
alex-yau-ttd, line 30: there's a few blank lines we can delete here

src/web/components/ClientSideTokenGeneration/CstgDomainItem.tsx
```diff
@@ -41,7 +40,7 @@ export function CstgDomainItem({
       <td className='domain'>{domain}</td>
       <td className='action'>
         <div className='action-cell'>
-          <EditButton onClick={onEditDialogChange} />
+          <ActionButton onClick={onEditDialogChange} className='action-button' icon={ActionButtonIcon.edit}/>
```
alex-yau-ttd, line 43: Could we possibly move the 'action-button' styling to `ActionButton.scss`? Then we won't need to specify the `className` every time. e.g. I noticed the `ActionButton` in `ApprovedParticipantItem.tsx` is missing the className, so is styled a bit differently.

src/web/components/Core/ActionButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import '../../styles/buttons.scss';
+
+export enum ActionButtonIcon {
+  edit = 'pencil',
+  delete = 'trash-can'
+};
+
+type ActionButtonProps = Readonly<React.ButtonHTMLAttributes<HTMLButtonElement>> & {
+  icon: ActionButtonIcon;
```
alex-yau-ttd, line 12: I'm not sure if we need this `ActionButtonIcon` enum - I think it is probably better to keep it more generic but also preserve type safety just have the `icon` prop be the type `IconProp`, like we have in `Notification.tsx`. Because that's the type that ends up getting passed into the `<FontAwesomeIcon>` anyway.

So I think this is better:
```
type ActionButtonProps = Readonly<React.ButtonHTMLAttributes<HTMLButtonElement>> & {
  icon: IconProp;
  iconClassName?: string;
};
```

Then the usage would remain something like:
```
<ActionButton
  onClick={onOpenChangeKeyEditDialog}
  icon='pencil'
  className='action-button'
/>
```

src/web/components/KeyPairs/KeyPairsTable.scss
```diff
@@ -36,6 +36,8 @@
       svg {
         font-size: 1rem;
       }
+
```
alex-yau-ttd, line 39: A few more lines we can delete

src/web/styles/tables.scss
```diff
@@ -51,10 +51,16 @@ table {
...
     flex-wrap: wrap;
+    
+    .action-button {
+      padding: 5px 5px 5px 10px;
+    }
 
     svg {
       font-size: 1rem;
     }
+
```
alex-yau-ttd, line 62: A few more lines we can delete

src/web/components/ClientSideTokenGeneration/CstgDomainsTable.scss
ashleysmithTTD: Sorry about that I assumed prettier would fix it but I think it was broken on my machine yesterday

src/web/components/Core/ActionButton.tsx
ashleysmithTTD: Makes sense - thanks!

PR #453, opened 2024-05-31
Add in-line delete button to Sharing Permissions Table

ashleysmithTTD:
What Changed:

- Added  in-line delete button for sharing permissions (also added Actions header)
- Extracted DeletePermissionsDialog to its own file 

Test Plans:

- Add and delete participants in the Sharing Permissions table - both individually and in bulk and make sure everything deletes correctly

### DIFF
--- a/src/web/components/SharingPermission/ParticipantItem.tsx
+++ b/src/web/components/SharingPermission/ParticipantItem.tsx
@@ -44,16 +44,24 @@ export function ParticipantItemSimple({ site }: ParticipantItemSimpleProps) {
 type ParticipantItemProps = ParticipantItemSimpleProps & {
   onClick: () => void;
   checked: boolean;
-  onDelete: () => void;
-  sharingSites: SharingSiteWithSource[]
+  onDelete?: (siteIdsToDelete: number[]) => void;
+  sharingSites?: SharingSiteWithSource[];
 };
 
-export function ParticipantItem({ site, onClick, checked, onDelete, sharingSites }: ParticipantItemProps) {
+export function ParticipantItem({
+  site,
+  onClick,
+  checked,
+  onDelete,
+  sharingSites,
+}: ParticipantItemProps) {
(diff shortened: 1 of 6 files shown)


ashleysmithTTD:
https://github.com/IABTechLab/uid2-self-serve-portal/assets/157655209/0b000464-8976-4cbe-8eb5-48aff3c7ff92

src/web/components/SharingPermission/ParticipantItem.tsx
```diff
@@ -61,6 +78,18 @@ export function ParticipantItem({ site, onClick, checked }: ParticipantItemProps
       </td>
       <ParticipantItemSimple site={site} />
       {isSharingParticipant(site) && <td>{formatSourceColumn(site.addedBy)}</td>}
+      {onDelete && (
```
ssundahlTTD, line 81: why is this being conditionally shown?  shouldn't there always be an onDelete prop?

ssundahlTTD:
Something is wrong with adding bulk permissions.  It doesn't show up in the table when added.  
image

lionell-pack-ttd:
The delete button shouldn't show up for participants that are only included by type. e.g. in this screenshot, "DSP example" shouldn't have a delete button. It's not in the list of individual participants so it can't be removed by deleting it.
image

src/web/components/SharingPermission/ParticipantItem.tsx
ashleysmithTTD: If this condition is not there, the delete button will show up on the dropdown when adding permissions because the same `ParticipantTable` component is used for the dropdown when searching and when displaying which permissions have been added

ssundahlTTD (approved):

ashleysmithTTD:
> Something is wrong with adding bulk permissions. It doesn't show up in the table when added. image
> 
> is it just me? seems to work in your video. Looks like just Publisher doesn't show up on the table for me.

We looked into this and it was just data not added, but once editing a publisher with a SHARER role in local admin, it showed up

PR #458, opened 2024-06-06
Ability to add Mobile App Ids in Client Side Integration

ashleysmithTTD:
What Changed:
- Ability to add, edit and delete mobile app ids
- Ability to view app ids in a table
- Validation for the three types of mobile app ids
- Changed several folders/files/function names to be more generalized between app ids and domains
- Add app ids as field where needed
- Changed instructions where necessary
- Corresponding storybook files
- Updated version of braces npm package to fix vulnerability
- Changed `clientSideIntegration.tsx` to use new loader pattern

Test Plan:
- Add mobile app ids
- Add invalid app ids
- Add duplicate app ids
- Add app ids that already exist in the list
- Add app ids while deleting the existing ones
- Edit app id
- Edit app id to something invalid
- Delete a single app id
- Bulk delete app ids
- Search app ids
- Test all steps above for domains as well 
- Test add, edit, delete for key pair
- Check that rows are inserted into audit trails db when an action is performed
- Check new storybook files


src/web/components/ClientSideTokenGeneration/CstgHelper.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { extractTopLevelDomain, getPagedDomains } from './CstgDomainHelper';
+import { extractTopLevelDomain, getPagedValues, isAppStoreId } from './CstgHelper';
```
ashleysmithTTD, line 1: Changed function name to `getPagedValues` to make it more general

src/web/components/ClientSideTokenGeneration/CstgDomainHelper.spec.ts
```diff
@@ -1,71 +0,0 @@
...
-      const rowsPerPage = 50;
-      expect(getPagedDomains(longDomainsList, pageNumber, rowsPerPage)).toEqual(longDomainsList);
-    });
-    it('should return an empty list', () => {
-      const pageNumber = 0;
-      const rowsPerPage = 10;
-      expect(getPagedDomains(longDomainsList, pageNumber, rowsPerPage)).toEqual([]);
-    });
-  });
-});
```
ashleysmithTTD, line 71: Moved to CstgHelper.ts

src/web/utils/textHelpers.spec.ts
```diff
@@ -216,4 +217,15 @@ describe('Text helper tests', () => {
       ]);
     });
   });
+
+  it('should check if string is alphanumeric with hyphen and dot', () => {
```
ssundahlTTD, line 221: Looks like you also have a helper method for underscores.  should add a test for that here, too.

src/web/screens/clientSideIntegration.tsx
```diff
@@ -47,6 +58,10 @@ function ClientSideIntegration() {
     loadDomainNames();
   }, [loadDomainNames]);
 
+  useEffect(() => {
```
ssundahlTTD, line 61: Can all of these be combined into one useEffect()?  And better yet, can we use the loader pattern that Lionell established to load all of these when the page itself loads?

src/web/components/ClientSideTokenGeneration/AppIds/CstgEditDomainDialog.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { Meta } from '@storybook/react';
```
ssundahlTTD, line 1: This file is named incorrectly (Domain rather than AppId)

src/web/services/appIdsService.ts
```diff
@@ -0,0 +1,38 @@
...
+export type EditAppIdFormProps = {
+  appId: string;
+};
+
+export async function GetAppIds(participantId?: number) {
+  try {
+    const result = await axios.get<string[]>(
+      `/participants/${participantId ?? 'current'}/appNames`
+    );
+    if (result.status === 200) {
```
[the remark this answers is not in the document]
ashleysmithTTD: Yep fallback handler will take care of it

src/web/components/ClientSideTokenGeneration/CstgAddDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+      });
+      if (invalidAppIds.length > 0) {
+        handleError(
+          `The mobile app ids entered are invalid: ' ${formatStringsWithSeparator(invalidAppIds)}`
+        );
+      } else {
+        await onAddCstgValues(newCstgValues, deleteExistingList);
+      }
+    }
+  };
```
ashleysmithTTD, line 95: This function may look a little funny between domain and app id - the reason is because currently, domain validation is on the backend and app id validation is on the front end.  Domains also need to be deduped a second time to filter out the case that two entered domains lead to the same +1 domain (ex. photos.google.com and drive.google.com both -> google.com)

src/web/components/ClientSideTokenGeneration/CstgEditDialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+    } else if (cstgValueType === CstgValueType.MobileAppId) {
+      if (!validateAppId(updatedCstgValue)) {
+        showInvalidError();
+      }
+    } else {
+      const editSuccess = await onEdit(updatedCstgValue, originalCstgValue);
+      if (!editSuccess) {
+        showInvalidError();
+      }
+    }
```
ashleysmithTTD, line 71: Again here, may look a little weird because validating an edited app id = front end, validating an edited top level domain = back end

src/api/services/adminServiceClient.ts
```diff
@@ -278,3 +278,13 @@ export const setSiteDomainNames = async (
   });
   return response.data;
 };
+
+export const setSiteAppNames = async (
+  siteId: number,
+  appNames: string[]
+): Promise<AdminSiteDTO> => {
+  const response = await adminServiceClient.post(`/api/site/app_names?=id=${siteId}`, {
```
lionell-pack-ttd, line 286: /api/site/app_names?=id=${siteId}

The first = looks unnecessary, but I can see the domain names call has the same extra =. Any ideas on why?

src/web/components/ClientSideTokenGeneration/CstgAddDialog.tsx
lionell-pack-ttd: Another way to approach this would be to inject the validation behaviour - either by providing a validator function or else just leaving validation to the `onAddCstgValues` handler. But ... we're expecting some re-work on this anyway when we do the UX enhancement for the app IDs, so I think this is fine for now.

src/web/components/ClientSideTokenGeneration/CstgEditDialog.tsx
lionell-pack-ttd: Perhaps injecting an optional validator (and maybe an optional transformer function as well - e.g. for domains it would be `extractTopLevelDomain`) might have been cleaner, but I think this approach is OK in this case. I don't think we're likely to extend it to 3+ types of things this table handles, and we can always tweak it later.

package.json
```diff
@@ -43,6 +43,7 @@
     "axios": "^1.6.7",
     "babel-plugin-transform-typescript-metadata": "^0.3.2",
     "body-parser": "^1.20.2",
+    "braces": "^3.0.3",
```
lionell-pack-ttd, line 46: Why is this added? I can't see it being used anywhere.

lionell-pack-ttd (approved):

PR #461, opened 2024-06-12
Show banner if user does not have sharing permissions

ashleysmithTTD:
What Changed:

- banner to show if user does not have sharing permissions on home page (see UX in Jira ticket)
- updated version number for braces npm package - was getting a high vulnerability that was causing the build to fail

Test Plan:
- test that the card looks correct when you do have sharing permissions
- change your participant's site id to a site id that does not have any keysets associated with it in local admin (I changed mine to site id = 1002) and see banner on home page


ashleysmithTTD:
NoSharingPermissions

ssundahlTTD (approved):

PR #462, opened 2024-06-13
Add a joyride setup and provide the first tour.

lionell-pack-ttd:


src/web/components/SiteTour/UpdatesTour.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+function callback(data: CallBackProps) {
+  const doNotShowAgainActions: Actions[] = ['close', 'skip', 'stop', 'reset'];
+  if (doNotShowAgainActions.includes(data.action)) markTourAsSeen();
+}
+
+export function UpdatesTour() {
+  const showTour = ShouldShowTour();
+  const actionButtonStyle = {
+    backgroundColor: '#cdf200', // --theme-button
```
lionell-pack-ttd, line 13: Sadly, joyride doesn't support classes or css - it sets styles directly on elements 😭

src/web/components/SiteTour/tourStorage.ts
```diff
@@ -0,0 +1,47 @@
...
+function isValidTourData(data: unknown | TourData): data is TourData {
+  return (
+    (data as TourData).seenForVersions !== undefined &&
+    Array.isArray((data as TourData).seenForVersions)
+  );
+}
+
+function getTourData(): TourData {
+  const noData = { seenForVersions: [] };
+  const unparsed = localStorage.getItem(tourStorageKey);
```
lionell-pack-ttd, line 28: Local storage for now, but we might need to think about some options etc. storage for each user for things like this.

PR #468, opened 2024-06-17
Add sorting to tables portal

ashleysmithTTD:
What Changed:

Added sorting to the following tables:

- Key Table in API Keys page
- Business Contacts table
- Key Pairs Table in Client Side Integration
- Approved Participants Table
- Team Members table 

Sorting on Domains table and Mobile App Ids table not added yet - these need to wait until we fix the pattern in `SortableTableHeader` to be able to sort on primitive types, not just objects.  A new ticket has been created for this.


src/web/components/KeyPairs/KeyPairsTable.tsx
```diff
@@ -59,20 +64,28 @@ function KeyPairsTable({
...
+              sortKey='created'
+              header='Created'
+            />
             <th className='action'>Actions</th>
           </tr>
         </thead>
         <tbody>
-          {keyPairs &&
-            keyPairs.map((k) => (
+          {sortedKeyPairs &&
```
alex-yau-ttd, line 83: Since we are defaulting to an empty array on line 40, I think we can get rid of this check. So I think we can just have:
```
<tbody>
  {sortedKeyPairs.map((k) => (
    <KeyPair
      key={k.publicKey}
      keyPair={k}
      existingKeyPairs={sortedKeyPairs}
      onEdit={onKeyPairEdit}
      onDisable={onKeyPairDisable}
    />
  ))}
</tbody>
```

alex-yau-ttd: sonarlint suggests using an optional chain but I don't think we even need that.

src/web/components/KeyPairs/KeyPairsTable.tsx
```diff
@@ -34,6 +36,9 @@ function KeyPairsTable({
     setShowKeyPairDialog(false);
   };
 
+  const { sortData } = useSortable<KeyPairModel>();
+  const sortedKeyPairs = keyPairs ? sortData(keyPairs) : [];
```
alex-yau-ttd, line 40: This line got me thinking; since https://github.com/IABTechLab/uid2-self-serve-portal/pull/466/files can't we guarantee that the keyPairs endpoint will return a defined type? If so, we can drop the `| undefined` type from the `keyPairs` prop and remove the check here, then this can just become:
```
const sortedKeyPairs = sortData(keyPairs);
```
or call it `sortedData` like the rest of the changes.

Thoughts?

ashleysmithTTD: Yes good call!

src/web/components/KeyPairs/KeyPairsTable.tsx
```diff
@@ -34,6 +36,9 @@ function KeyPairsTable({
     setShowKeyPairDialog(false);
   };
 
+  const { sortData } = useSortable<KeyPairModel>();
+  const sortedKeyPairs = sortData(keyPairs);
```
alex-yau-ttd, line 40: nit: in some places we would just call this variable `sortedData`. Perhaps we should be consistent? Maybe always name it something useful like we have here? Examples are in `KeyTable.tsx`, `BusinessContactsTable.tsx` and `ParticipantsTable.tsx`

alex-yau-ttd (approved):

src/web/components/KeyPairs/KeyPairsTable.tsx
lionell-pack-ttd: I think `sortedKeyPairs` is definitely better than `sortedData` - I don't think it's all that important to update other places though, if the meaning is clear.

PR #475, opened 2024-06-26
Various code improvements

alex-yau-ttd:
- Move away from deprecated methods
- Improve an error message
- Improve some test names
- Remove a use of "as"
- Improve a test assertion
- Fix and improve participant key pairs test
- Fix logic around the CSTG completion banner & added a story
- Other ESLint suggestions

### DIFF
--- a/src/api/routers/participants/participantsKeyPairs.spec.ts
+++ b/src/api/routers/participants/participantsKeyPairs.spec.ts
@@ -10,10 +10,15 @@ import { KeyPairDTO } from '../../services/adminServiceHelpers';
 import { ParticipantRequest } from '../../services/participantsService';
 import { getParticipantKeyPairs } from './participantsKeyPairs';
 
-const oneKeyPair: KeyPairDTO[] = [
+const noKeyPairsSiteId = 10;
+const singleKeyPairSiteId = 11;
+const multipleKeyPairsSiteId = 12;
+const disabledKeyPairSiteId = 13;
+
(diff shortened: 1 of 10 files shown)


src/api/routers/participants/participantsKeyPairs.spec.ts
```diff
@@ -81,10 +112,11 @@ describe('#getParticipantKeyPairs', () => {
   });
 
   test.each([
-    [10, []],
-    [11, oneKeyPair],
-    [12, multipleKeyPairs],
```
alex-yau-ttd, line 86: This test was previously failing because it was expecting to return all keypairs, not just the enabled ones. I've added a test for the disabled keypair case and extracted a bunch of magic numbers to constants.

src/web/services/apiKeyService.ts
```diff
@@ -56,6 +56,6 @@ export async function DisableApiKey(apiKey: ApiKeyDTO, participantId?: number) {
       data: { keyId: apiKey.key_id },
     });
   } catch (e: unknown) {
-    throw backendError(e, 'Could not disable API Key');
```
ashleysmithTTD, line 59: Is this something a user would see? If not, I think it makes sense to leave as disable since that is what is happening from an engineer's point of view, but if the user would see this, then I'm good with the change to 'delete'

alex-yau-ttd: Yes this flows through to the UI

PR #476, opened 2024-06-28
Allow user to belong to multiple participants (backend)

alex-yau-ttd:
## What changed
- Migration to add new table to store the mapping between user and participants, moving away from the participantId column in the User table.
- Entity model changes. Drop participantId from User entity in favour of a participants array. Change relation mapping from one-to-many to many-to-many
- API changes to support the change. Typically fetching the first participant associated with the user
- API test changes. The entity changes make it quite difficult to mock, so instead move towards service tests using the in-memory test database.
- Some minor storybook changes
- Add some TODOs for future work

## Manual testing
- Migrate to the latest schema, which will populate the new table
- (Optional) Null your participantId column for your user in `dbo.users`, and see the portal still working as expected

## Future work
- Removing participantId from the DB schema 
- Support ability to select an individual participant, rather than just returning the first one


src/api/services/usersService.ts
```diff
@@ -76,25 +78,15 @@ export const enrichWithUserFromParams = async (
   next: NextFunction
 ) => {
   const { userId } = userIdParser.parse(req.params);
-  const traceId = getTraceId(req);
   const user = await User.query().findById(userId);
   if (!user) {
     return res.status(404).send([{ message: 'The user cannot be found.' }]);
   }
-  if (
```
alex-yau-ttd, line 84: The above changes mean that the user must already be related to the participant by now

src/api/entities/BusinessContact.ts
```diff
@@ -12,12 +12,16 @@ export class BusinessContact extends BaseModel {
...
     return 'businessContacts';
   }
-  static relationMappings = {
+  static readonly relationMappings = {
     participant: {
       relation: Model.BelongsToOneRelation,
       modelClass: 'Participant',
       join: {
-        from: 'users.participantId',
+        from: 'users.id',
```
lionell-pack-ttd, line 20: I don't think users.id was ever right here. A business contact belongs to a participant, not a user.

src/api/entities/User.ts
```diff
@@ -46,7 +50,6 @@ export class User extends BaseModel {
   declare lastName: string;
   declare phone?: string;
   declare role: UserRole;
-  declare participantId?: number | null;
```
lionell-pack-ttd, line 49: Optional: You could add a Participant[] here, similar to how Participant has a User[]. I'm still not 100% confident with Objection though so not sure if it will work without other changes though.

src/api/entities/BusinessContact.ts
alex-yau-ttd: Fixed this to just join the Participants table directly

src/api/services/usersService.ts
lionell-pack-ttd: Sent you a DM.

src/api/tests/usersRoute.spec.ts
```diff
@@ -30,15 +30,5 @@ describe('Users API tests', () => {
       expect(res.statusCode).toBe(200);
       expect(sendInviteEmail).toHaveBeenCalled();
     });
-
-    test('Should deny access to an authenticated user without permission', async () => {
```
lionell-pack-ttd, line 34: I think this is related to my other comment - why do you think this test scenario isn't valid now?

src/api/services/usersService.ts
```diff
@@ -20,10 +20,15 @@ export const enrichUserWithIsApprover = async (user: User) => {
   };
 };
 
-export const createUserInPortal = async (user: Omit<UserDTO, 'id' | 'acceptedTerms'>) => {
+export const createUserInPortal = async (
```
lionell-pack-ttd, line 23: I think it's worth throwing a TODO: in here for the follow-up ticket - this is used for the /invite handler, but right now if an existing user is invited it'll find it and return it without adding the new participant.

src/api/entities/User.ts
alex-yau-ttd: I ended up adding in `Participant[]`!

src/api/tests/queryMocks.ts
```diff
@@ -48,13 +48,26 @@ export const mockUser = (user: PartialUserOrNull | PartialUserOrNull[] = {}) =>
               id: 1,
               email: 'test_user@example.com',
               name: 'Test User',
-              participantId: 1,
+              participants: u.participants ?? [
```
alex-yau-ttd, line 51: These mocks are still used by approversMiddleware and businessContactService tests. This change will allow those tests to still work, since it relies on logic in `hasBusinessContactAccess` instead.

src/api/tests/participantsService.spec.ts
alex-yau-ttd, line 1: This virtually a new file which tests the service rather than the API endpoint.

src/api/tests/usersRoute.spec.ts
alex-yau-ttd, line 1: This is replaced by usersService.spec.ts which tests at the service level, rather than the API endpoint level.

src/api/tests/usersService.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  return {
+    auth: {
+      payload: {
+        email,
+      },
+    },
+    params: {
+      userId,
+    },
+  } as unknown as UserRequest;
```
alex-yau-ttd, line 22: I don't love this but I couldn't figure out a better way to do this. Open to ideas

src/api/tests/participantsService.spec.ts
```diff
@@ -1,90 +1,112 @@
...
+  return {
+    auth: {
+      payload: {
+        email,
+      },
+    },
+    params: {
+      participantId,
+    },
+  } as unknown as ParticipantRequest;
```
alex-yau-ttd, line 25: I don't love this but I couldn't figure out a better way to do this. Open to ideas

lionell-pack-ttd: I think it's OK in this case, especially because it's a test.

src/api/tests/usersService.spec.ts
lionell-pack-ttd: Again, this makes sense. This is actually good - what we're doing isn't "correct" (in that this *isn't* a UserRequest), but it's close enough for this test. So we tell TypeScript "shhhhh.... this is wrong but it's OK for the purpose of this test".

src/api/controllers/userController.ts
```diff
@@ -59,13 +59,10 @@ export class UserController {
...
   public async acceptTerms(@request() req: UserRequest, @response() res: Response): Promise<void> {
-    if (!req.user?.participantId) {
-      res.status(403).json({
-        message: 'Unauthorized. You do not have the necessary permissions.',
-        errorHash: req.headers.traceId,
-      });
-    }
-    const participant = await Participant.query().findById(req.user!.participantId!);
+    // TODO: This just gets the user's first participant, but it will need to get the currently selected participant as part of UID2-2822
+    const currentParticipantId = req.user?.participants?.[0].id ?? 0;
```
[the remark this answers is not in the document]
alex-yau-ttd: done!

lionell-pack-ttd (approved):

PR #479, opened 2024-07-01
Moved core components into more specific subfolders

ashleysmithTTD:


src/web/components/Core/Popups/Dialog.tsx
alex-yau-ttd, line 1: I think Dialog probably deserves it's own folder

src/web/components/Core/Banner/EnvironmentBanner.tsx
alex-yau-ttd, line 1: I think components in the Core folder should typically be used more than once. The EnvironmentBanner will only really be used once, so probably doesn't belong inside the Banner folder. Perhaps a separate folder outside of Core?

src/web/components/Core/Headers/SortableTableHeader.tsx
alex-yau-ttd, line 1: I see this has being more related to Tables rather than Headers. Perhaps this, along with TableNoDataPlaceholder could go into a Tables or TableHelpers folder?

src/web/components/Core/Popups/Tooltip.tsx
alex-yau-ttd, line 1: I think Tooltip could probably be it's own folder, outside of Popups

src/web/components/Core/Messages/Messages.scss
alex-yau-ttd, line 1: Not related to your changes, but I wonder why we need this file when we already have `InlineMessage.scss` 🤔

src/web/components/Core/Dropdown/Collapsible.tsx
alex-yau-ttd, line 1: I wouldn't technically call this a Dropdown - perhaps it deserves it's own folder?

src/web/components/Core/Dropdown/SelectDropdown.tsx
alex-yau-ttd, line 1: Perhaps SelectDropdown and MultiSelectDropdown should move to Input?

src/web/components/Core/TermsAndConditions/TermsAndConditions.tsx
alex-yau-ttd, line 1: This can probably be moved outside of Core

src/web/components/Core/Buttons/CopySecretButton.tsx
alex-yau-ttd, line 1: The ViewSecretButton is currently in the DisplaySecret folder. Perhaps we could also move this there?

src/web/components/Core/Buttons/FormSubmitButton.tsx
alex-yau-ttd, line 1: Maybe this could live in a new Form folder?

ashleysmithTTD: I think for this one, it is good in the buttons folder along with other common buttons used

src/web/components/Core/Banner/EnvironmentBanner.tsx
ashleysmithTTD: I agree with that for something like the Terms and Conditions files/folder, but I think for the EnvironmentBanner, it still makes sense to be grouped with the Banner folder

src/web/components/Core/Messages/Messages.scss
ashleysmithTTD: Took another look and `Messages.scss` is only used within the Banner, so I just moved all of that css code to `Banner.scss` and then removed `Messages.scss`

alex-yau-ttd (approved):

PR #482, opened 2024-07-02
Fix create key dialog for consistency

ashleysmithTTD:
- Removed `hideCloseButtons` prop from the create key dialog - this prop is not used anywhere in the code now 
- Added the css for a dialog title for the title of the Create API Key form

### DIFF
--- a/src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
+++ b/src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
@@ -148,7 +148,7 @@ function KeyCreationDialog({
 
   return (
     <div className='key-creation-dialog'>
-      <Dialog onOpenChange={onOpenChange} hideCloseButtons>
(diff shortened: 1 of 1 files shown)


src/web/components/ApiKeyManagement/KeyCreationDialog.tsx
```diff
@@ -39,7 +39,7 @@ function CreateApiKeyForm({ onFormSubmit, availableRoles, closeDialog }: CreateA
   const { handleSubmit } = formMethods;
   return (
     <>
-      <h1>Add API Key</h1>
+      <h1 className='dialog-title'>Add API Key</h1>
```
ashleysmithTTD, line 42: The reason that the Dialog cannot have the title passed to it like dialogs in other files is because the dialog alternates depending on if we are showing the secret or not, and we don't want the title to be the same for both instances

ssundahlTTD (approved):

PR #483, opened 2024-07-04
Dark mode support for portal UI

ashleysmithTTD:
What changed:

- uncommented dark mode code
- added several new themes and changes to css files to make the color alternate between normal and dark mode


.storybook/main.ts
alex-yau-ttd, line 1: Should we just revert this change?

alex-yau-ttd (approved):

src/web/utils/errorHandler.scss
```diff
@@ -1,5 +1,31 @@
-.error-content {
-  align-self: flex-start;
-  text-align: center;
-  padding: 4.5rem 6.5rem;
+.app {
+  .error-content {
```
alex-yau-ttd, line 2: We could probably extract this part out so we don't need to define it twice.
Something like:
```
.error-content {
  // common stuff
}

.app .error-content {
  // light mode specific stuff
}

.app.darkmode .error-content {
 // dark mode specifif stuff
}
```

.storybook/main.ts
ashleysmithTTD: I'm not sure how to override my editor - everytime i try to save the file, it auto makes this format changes

PR #485, opened 2024-07-05
check for new tour versions

ssundahlTTD:
- Go to the UI and see that the tour is being shown
- If you've previously finished the tour and do not see it, enter the following in the console: delete localStorage['tour-data']
- Finish the tour
- Reload the page and verify that the tour does not show

### DIFF
--- a/src/web/components/SiteTour/tourStorage.spec.ts
+++ b/src/web/components/SiteTour/tourStorage.spec.ts
@@ -1,0 +1,64 @@
+import { compareVersions, GetTourSteps, tourStorageKey, VersionedTourStep } from './tourStorage';
+
+const mockTourData: VersionedTourStep[] = [
+  {
+    target: `.test-class`,
+    content: `Test tour.`,
+    disableBeacon: true,
+    version: '0.36.0',
+  },
+];
+
+describe('Tour tests', () => {
+  test('When the last tour seen is lower than a tour, it is included.', () => {
+    localStorage.setItem(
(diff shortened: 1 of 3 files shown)


src/web/components/SiteTour/tourStorage.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  test('Compare Versions sorts in correct order', () => {
+    const data = ['1', '2', '7', '13', '0', '7', '102', '53', '5'];
+    const majors = [...data];
+    const majorAndMinor = majors.flatMap((maj) => data.map((min) => `${maj}.${min}`));
+    const versions = majorAndMinor.flatMap((mm) => data.map((patch) => `${mm}.${patch}`));
+    expect(versions).toContainEqual('7.2.13');
+    versions.sort(() => Math.random());
+    versions.sort(compareVersions);
+    expect(versions[0]).toBe('0.0.0');
+    expect(versions.slice(-1)[0]).toBe('102.102.102');
```
alex-yau-ttd, line 62: nit: could we just use `pop()` here?

lionell-pack-ttd: I suppose, but I just want the last element - I could also use _versions[versions.length-1]_. I don't think it really matters.

alex-yau-ttd: yeah fair enough!

PR #489, opened 2024-07-26
Improve audit trail pattern and increase coverage

alex-yau-ttd:
## Overview
- Establish new pattern which reduces boilerplate to write audit logs
- Use this new pattern for all existing audit logs and also for features that lacked any audit logs (i.e. team member management)
- Simplify audit trail entity and service to be more generic
- Re-add participantId as a column in the audit trail table. Migrate existing participantIds from the eventData column into the participantId column.
- Some refactoring of the userController into the userService
- Fix a bug which meant that team members could not be updated

## The new audit trail pattern
- See `AuditTrail.ts` for the simplified entity, with less type-safety for the `eventData` field.
- See `auditTrailService.ts`, where many functions have been generalised into a single function. It also handles errors and updates the `succeeded` flag.
- There's many usages of the new pattern across the router and service files. Most of these changes live in `participantsRouter.ts`, and there are pretty much the same. Look at the code for deleting an API key as a fairly typical example of the new pattern.
- Any new features will just require adding to the `AuditTrailEvents` enum, then calling the existing function in `auditTrailService.ts`. No longer requires a new type, audit function as well as the need to manually update the succeeded flag. Just stick the event-specific data straight into the `eventData` field and move on.

## Manual testing
1. Run the latest db migration
1. Perform any action in the portal (bar Email Contact management). 
  a. Test Team Member management, particularly editing a team member
1. Observe the audit trails table and see that the information is accurate. Particularly userId, userEmail, event, eventData,  succeeded and participantId.


src/api/controllers/userController.ts
alex-yau-ttd, line 1: Did some factoring out into userService.ts

src/api/entities/AuditTrail.ts
```diff
@@ -123,7 +45,8 @@ export class AuditTrail extends BaseModel {
   declare userEmail: string;
   declare succeeded: boolean;
   declare event: AuditTrailEvents;
-  declare eventData: AuditTrailEventData;
+  declare eventData: unknown;
```
alex-yau-ttd, line 48: Just make it not as type-safe, hence can delete all the types above.

src/web/components/TeamMember/TeamMemberDialog.tsx
```diff
@@ -68,11 +69,11 @@ function TeamMemberDialog(props: TeamMemberDialogProps) {
           <TextInput
             inputName='email'
             label='Email'
-            disabled={!!props.person}
+            disabled={editMode}
             rules={{
               required: 'Please specify email.',
               validate: (value: string) => {
-                if (!validateUniqueTeamMemberEmail(value, props.teamMembers))
+                if (!editMode && !validateUniqueTeamMemberEmail(value, props.teamMembers))
```
alex-yau-ttd, line 76: We don't need to run this validation in edit mode, otherwise we can never actually update a team member as it fails with `Team member email already exists.`

src/database/migrations/20240724003623_AddParticipantIdColumnToAuditTrail.ts
alex-yau-ttd, line 1: We previously remove the participantId column from the audit trail table to remove the FK constraint as part of !245. This would allow us to delete sample participants while running the seed data. We want to re-introduce this column and FK constraint so that we can easily query audit logs per participant, so I've re-used the migration from that, and also updated the seeds/Participants file so that audit trails for the sample participants are just given a `null` participantId, hence it won't break the FK constraint.

src/api/services/auditTrailService.ts
```diff
@@ -1,279 +1,21 @@
...
-
-    return await AuditTrail.query().insert(appNamesAuditTrail);
-  } catch (error) {
-    const { errorLogger } = getLoggers();
-    errorLogger.error(`Audit trails inserted failed: ${error}`, traceId);
-    throw error;
-  }
-};
-
-export const insertApproveAccountAuditTrail = async (
```
alex-yau-ttd, line 218: This function and the below function had a slightly different pattern, where they accepted an object rather than a list of parameters. This was more elegant than all the above functions, hence the new pattern does not look as impressive where `insertApproveAccountAuditTrail` and `insertAddParticipantAuditTrail` were previously used.

ssundahlTTD (approved):

src/api/services/auditTrailService.ts
```diff
@@ -1,279 +1,21 @@
...
-};
-
-export const insertDomainNamesAuditTrails = async (
-  participant: Participant,
-  userId: number,
-  userEmail: string,
-  action: AuditAction,
-  domainNames: string[],
-  traceId: string
+  operation: () => Promise<T>
```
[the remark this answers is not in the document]
alex-yau-ttd: hmm I don't love auditOperation or auditHandler because `operation` itself doesn't do any auditing. I could call it `auditedOperation` but I'm not sure that's an improvement. I think I'll merge as-is for now but happy to change later if we can come up with a better name.

src/api/controllers/userController.ts
```diff
@@ -136,33 +128,22 @@ export class UserController {
...
-    const { user } = req;
-    if (req.auth?.payload?.email === user?.email) {
-      res.status(403).send([{ message: 'You do not have permission to delete yourself.' }]);
+    try {
+      await this.userService.deleteUser(req);
+      res.sendStatus(200);
+    } catch (error: unknown) {
+      if (
+        error instanceof Error &&
+        error.message === 'You do not have permission to delete yourself.'
```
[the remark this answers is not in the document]
alex-yau-ttd: yeah good shout - I was trying to make the controller layer as thin as possible and just pass the request to the service as-is, but I think some basic request parsing is fine here - fixed!

src/api/services/participantsService.ts
```diff
@@ -237,14 +239,30 @@ const updateParticipantParser = z.object({
...
     participantName,
     crmAgreementNumber,
   } = updateParticipantParser.parse(req.body);
-  try {
+  const { user } = req;
+  const traceId = getTraceId(req);
+
+  const auditTrailInsertObject: InsertAuditTrailDTO = {
+    userId: user!.id,
+    userEmail: user!.email,
```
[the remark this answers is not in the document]

src/api/services/userService.ts
```diff
@@ -46,4 +52,76 @@ export class UserService {
...
+    }
+    const data: DeletedUser = {
+      email: `${user?.email}-removed-${uuid()}`,
+      // TODO: Remove participantId in UID2-3821
+      participantId: null,
+      deleted: true,
+    };
+
+    const auditTrailInsertObject: InsertAuditTrailDTO = {
+      userId: requestingUser!.id,
```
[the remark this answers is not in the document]
alex-yau-ttd: done!

PR #491, opened 2024-07-29
Rename "role" to "jobFunction" in User table

alex-yau-ttd:
- Rename "role" on the Users table to avoid confusion with RBAC. Update usages of entity across services, front end, seed, stories, etc.
- Use the same term we already use in the UI: **jobFunction**
- No functional changes for end user

image

image

### DIFF
--- a/src/api/entities/AuditTrail.ts
+++ b/src/api/entities/AuditTrail.ts
@@ -61,7 +61,7 @@ export type AddParticipantEventData = {
   lastName: string;
   participantName: string;
   participantTypes: number[];
-  role: string;
+  jobFunction: string;
   siteId?: number;
   crmAgreementNumber: string;
 };
--- a/src/api/routers/participants/participantsRouter.ts
+++ b/src/api/routers/participants/participantsRouter.ts
@@ -252,7 +252,7 @@ export function createParticipantsRouter() {
         const { participant, user } = req;
         const { firstName, lastName, email, jobFunction } = invitationParser.parse(req.body);
         const traceId = getTraceId(req);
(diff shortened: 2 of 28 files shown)


src/api/entities/AuditTrail.ts
alex-yau-ttd, line 1: These types are going to be removed in https://github.com/IABTechLab/uid2-self-serve-portal/pull/489 anyway

src/api/routers/participants/participantsRouter.ts
```diff
@@ -11,7 +11,7 @@ import {
   ParticipantDTO,
   ParticipantStatus,
 } from '../../entities/Participant';
-import { UserDTO, UserRole } from '../../entities/User';
+import { JobFunction, UserDTO } from '../../entities/User';
```
ashleysmithTTD, line 14: Would we want to make this variable name UserJobFunction since it was UserRole before? And the change is Role -> JobFunction?

alex-yau-ttd: yeah I was on the fence about it since I figured `JobFunction` would be descriptive enough, but have made the change since you mentioned it.

ashleysmithTTD (approved):

src/api/services/userService.ts
```diff
@@ -99,9 +101,10 @@ export class UserService {
       AuditTrailEvents.ManageTeamMembers,
       {
         action: AuditAction.Update,
-        firstName: user?.firstName,
```
alex-yau-ttd, line 103: this was updating the audit with the old `user`, not the new `data`.

PR #492, opened 2024-08-07
User/Participant/Role relationship backend

alex-yau-ttd:
- Migration to create `userRoles` table
- Migration to add column to `usersToParticipantRoles` table, using default value of 1 (Admin)
- New `UserRole` and `UserToParticipantRole` entities
- Relate the `UserToParticipantRole` entity to `UserRole`, `User` and `Participant`.
- Alter participant/user creation methods so that the three-way mapping is also updated. Objection doesn't seem to handle this three-way mapping very well, so we've opted to manually insert into the mapping table rather than using graph inserts.
- Simplify some parsers to only parse the actual request from the front-end, rather than the request + properties we add in the back-end.
- Some minor refactoring out of router file
- Some minor renaming for clarity
- Modify `createUser` test method to accept a participant/role mapping, rather than just a `participantId`.

## Manual testing
1. Run the migrations to get the latest schema. 
2. Observe that all rows in `usersToParticipantRoles` have a `userRoleId` of 1.
3. Add yourself to another participant manually into the db, optionally with a different `userRoleId`. For me, this was done via:
    ```
    insert into dbo.userstoParticipantRoles (userId, participantId, userRoleId) values (3, 9, 3)
    ```
4. Log into the UI and check the response from `http://localhost:3000/api/users/current`. It should include the `currentUserRoleIds` array (as well as a separate `Participants` array), e.g.:
    ```
    "id": 3,
    "email": "abby@example.com",
    ...
    "participantId": 8, // will be changed/removed in future
    "participants": [
        {
            "id": 8,
            "name": "Participant A",
            ...
            "currentUserRoleIds": [
                1
            ]
        },
        {
            "id": 9,
            "name": "Participant B",
            ...
            "currentUserRoleIds": [
                3
            ]
        }
    ],
    ...
    ```

### DIFF
--- a/src/api/entities/UserRole.ts
+++ b/src/api/entities/UserRole.ts
@@ -1,0 +1,26 @@
+import { BaseModel } from './BaseModel';
+
+export class UserRole extends BaseModel {
+  static get tableName() {
+    return 'userRoles';
+  }
+  // TODO Consider if I need this relationMapping
+  // static readonly relationMappings = {
+  //   user: {
+  //     relation: Model.ManyToManyRelation,
+  //     modelClass: 'User',
+  //     join: {
+  //       from: 'userRoles.id',
+  //       through: {
+  //         from: 'usersToParticipantRoles.userRoleId',
+  //         to: 'usersToParticipantRoles.userId',
+  //       },
+  //       to: 'users.id',
+  //     },
+  //   },
+  // };
+  declare id: number;
+  declare roleName: string;
+}
+
+export const ADMIN_USER_ROLE_ID = 1;
--- a/src/api/entities/UserRole.ts
+++ b/src/api/entities/UserRole.ts
@@ -4,21 +4,6 @@ export class UserRole extends BaseModel {
   static get tableName() {
     return 'userRoles';
   }
-  // TODO Consider if I need this relationMapping
-  // static readonly relationMappings = {
-  //   user: {
-  //     relation: Model.ManyToManyRelation,
-  //     modelClass: 'User',
-  //     join: {
-  //       from: 'userRoles.id',
-  //       through: {
-  //         from: 'usersToParticipantRoles.userRoleId',
-  //         to: 'usersToParticipantRoles.userId',
-  //       },
-  //       to: 'users.id',
-  //     },
-  //   },
-  // };
   declare id: number;
   declare roleName: string;
 }
--- a/src/api/entities/UserToParticipantRole.ts
+++ b/src/api/entities/UserToParticipantRole.ts
@@ -61,9 +61,3 @@ export type UserToParticipantRoleDTO = ModelObjectOpt<UserToParticipantRole>;
 export type TestUserToParticipantRoleDTO = Partial<
   Pick<ModelObjectOpt<UserToParticipantRole>, 'participantId' | 'userRoleId'>
 >;
-
-// TODO remove unused comment
-// export type UserToParticipantRoleCreationDTO = Omit<
-//   ModelObjectOpt<UserToParticipantRole>,
-//   'userId'
-// >;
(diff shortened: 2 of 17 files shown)


src/api/entities/UserRole.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseModel } from './BaseModel';
```
alex-yau-ttd, line 1: This model doesn't need any relationMappings because the mappings are in the UserToParticipantRole mapping entity

src/api/entities/UserToParticipantRole.ts
```diff
@@ -0,0 +1,39 @@
+import { Model } from 'objection';
+
+export class UserToParticipantRole extends Model {
+  static get tableName() {
+    return 'usersToParticipantRoles';
+  }
+  static get idColumn() {
```
alex-yau-ttd, line 7: the PK of this table is the combination of all three columns, which we have to specify here

src/api/routers/participants/participantClasses.ts
```diff
@@ -22,13 +21,9 @@ export const ParticipantCreationAndApprovalPartial = ParticipantSchema.pick({
   name: true,
   types: true,
   apiRoles: true,
-  status: true,
```
alex-yau-ttd, line 25: Removing these as they don't actually come from the front-end request

src/api/routers/participants/participantsCreation.ts
```diff
@@ -61,6 +71,30 @@ export async function validateParticipantCreationRequest(
   return errorMessage;
 }
 
+const createUserAndAssociatedParticipant = async (
+  parsedUser: z.infer<typeof UserCreationPartial>,
+  participantData: z.infer<typeof ParticipantCreationAndApprovalPartial>
+) => {
+  await User.transaction(async (trx) => {
```
alex-yau-ttd, line 78: > Note: Even if you start a transaction using Person.transaction it doesn't mean that the transaction is just for Persons. It's just a normal knex transaction, no matter what model you use to start it. You can even use the Model base class if you want.

Source: https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction

src/api/routers/participants/participantsCreation.ts
```diff
@@ -61,6 +71,30 @@ export async function validateParticipantCreationRequest(
...
+    // create user
+    const newPortalUser = await User.query(trx).insertAndFetch(parsedUser);
```
alex-yau-ttd, line 80: `insertGraph` does not seem to support our use-case of the three-way mapping here, hence we are manually inserting the User, Participant and the User/Participant/Role mapping individually.

src/api/routers/participants/participantsCreation.ts
```diff
@@ -142,11 +172,59 @@ export async function createParticipant(req: ParticipantRequest, res: Response)
...
+    assignClientRoleToUser(kcAdminClient, user.email, 'api-participant-member');
 
     // send email
     await sendInviteEmail(kcAdminClient, newKcUser);
   });
 
   return res.sendStatus(200);
 }
+
+export const createParticipantFromRequest = async (req: ParticipantRequest, res: Response) => {
```
alex-yau-ttd, line 184: Extracted this method from `participantsRouter.ts`. Similar changes to `createParticipant` but decided against factoring out similar code because of slight differences in the data they would pass in and return.

src/api/services/participantsService.ts
```diff
@@ -164,7 +164,7 @@ export const updateParticipantApiRolesWithTransaction = async (
   }
 };
 
-export const updateParticipantAndTypesAndRoles = async (
+export const updateParticipantAndTypesAndApiRoles = async (
```
alex-yau-ttd, line 167: Rename to differentiate API roles from user roles

PR #494, opened 2024-08-08
Populate support user role for approvers

alex-yau-ttd:
- Use the existing `dbo.approvers` table to populate the `dbo.userToParticipantRoles` table.
- Since `participantId` is not nullable, it just finds the lowest `participantId` against the existing approver user, then inserts a new row with the `userRoleId` that matches "UID2 Support". This `participantId` here doesn't really matter, since the presence of UID2 Support will bypass any participant restrictions.
- No-op if there are no rows in the approvers table

### DIFF
--- a/src/database/migrations/20240808041539_PopulateUID2SupportUsers.ts
+++ b/src/database/migrations/20240808041539_PopulateUID2SupportUsers.ts
@@ -45,6 +45,7 @@ export async function up(knex: Knex): Promise<void> {
 
 export async function down(knex: Knex): Promise<void> {
   const approverUserIds = await getAllApproverUserIds(knex);
+
   await knex('usersToParticipantRoles')
     .whereIn('userId', approverUserIds)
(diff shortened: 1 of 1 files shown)


src/database/migrations/20240808041539_PopulateUID2SupportUsers.ts
```diff
@@ -0,0 +1,44 @@
...
+        participantId: userToParticipant.participantId,
+        userRoleId: UID2_SUPPORT_USER_ROLE_ID,
+      };
+      return data;
+    })
+  );
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex('usersToParticipantRoles').where('userRoleId', '=', UID2_SUPPORT_USER_ROLE_ID).del();
```
alex-yau-ttd, line 43: I tried to use `whereIn` here but I'm getting this error message 
image

ashleysmithTTD (approved):

PR #495, opened 2024-08-09
Check for UID2 Support role in middleware

alex-yau-ttd:
- Add checks in the users and participants middleware to check for the existence of the UID2 Support role
- Add a test case to cover the new functionality
- Change from exporting a single user role ID to an enum of all three.

### DIFF
--- a/src/api/middleware/participantsMiddleware.ts
+++ b/src/api/middleware/participantsMiddleware.ts
@@ -5,7 +5,7 @@ import { Participant } from '../entities/Participant';
 import { getTraceId } from '../helpers/loggingHelpers';
 import { ParticipantRequest } from '../services/participantsService';
 import { findUserByEmail } from '../services/usersService';
-import { isUserBelongsToParticipant } from './usersMiddleware';
+import { isUserBelongsToParticipant, userHasUid2SupportRole } from './usersMiddleware';
 
 const idParser = z.object({
   participantId: z.coerce.number(),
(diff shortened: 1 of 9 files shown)


src/api/middleware/participantsMiddleware.ts
```diff
@@ -20,7 +20,13 @@ const hasParticipantAccess = async (req: ParticipantRequest, res: Response, next
...
 
   const currentUserEmail = req.auth?.payload?.email as string;
-  if (!(await isUserBelongsToParticipant(currentUserEmail, participantId, traceId))) {
+  const currentUserIsUid2Support = await userHasUid2SupportRole(currentUserEmail);
+
+  const userHasAccessToParticipant =
+    currentUserIsUid2Support ||
+    (await isUserBelongsToParticipant(currentUserEmail, participantId, traceId));
+
+  if (!userHasAccessToParticipant) {
```
ashleysmithTTD, line 29: nit: based on naming conventions and the pattern of putting the verb first (is/can/has), maybe makes more sense to be `hasUserUid2SupportRole` or `hasUserAccessToParticipant` but up to you

ashleysmithTTD (approved):

src/api/middleware/participantsMiddleware.ts
alex-yau-ttd: good suggestion! I've renamed this and a few other variables/methods.

PR #501, opened 2024-08-21
Change URL scheme to include participant ID

alex-yau-ttd:
## What changed
- Use the new scheme of `/participant/:participantId/<page_name>` rather than `/dashboard/<page_name>`. Including giving the home page a well-deserved path name. Added a re-direct from legacy routes to the new root.
- Added a wrapper function for `makeLoader` which makes a loader that is enriched with the `participantId` from the URL. This allows all the loader functions to be called with the `participantId` - so we can start loading data before the React components are aware of the URL.
- Add a new landing component that is hit when the user logs in. This component loads the first participant of the user then navigates to the appropriate home URL for that participant.
- Get rid of usage of `<SnailTrail>` in the dashboard - I don't think it was actually doing anything.
- A bunch of other changes to navigation. Trying to use relative paths were possible (e.g. `RotateApiKeyCard.tsx`), but otherwise need to use the  `useParams()` hook to populate the `participantId` in the path properly (e.g. `PortalHeader.tsx`)

## Manual testing
1. Try do everything you usually would in the portal, including navigation, creating keys, etc.
2. You'll notice a bunch of the API calls have the `participantId` specified in the request URL (e.g. `http://localhost:3000/api/participants/8/apiKeys`), rather than relying on "current": `http://localhost:3000/api/participants/current/apiKeys`.
3. Hit a legacy URL e.g. `http://localhost:3000/dashboard/apiKeys` and either be routed to the new home route (if authenticated) or the login page (if unauthenticated)

## Future work
- Remove support for "current" in the web services - i.e. make `participantId` mandatory where it is often currently optional
- Remove `<SnailTrail>` component

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@types/react": "^18.2.54",
     "@types/react-dom": "^18.2.18",
     "@types/validator": "^13.11.10",
-    "axios": "^1.6.7",
+    "axios": "^1.7.4",
     "babel-plugin-transform-typescript-metadata": "^0.3.2",
     "body-parser": "^1.20.2",
     "braces": "^3.0.3",
--- a/src/web/components/Navigation/SideNav.tsx
+++ b/src/web/components/Navigation/SideNav.tsx
@@ -3,10 +3,11 @@ import {
   NavigationMenuItem,
   NavigationMenuList,
 } from '@radix-ui/react-navigation-menu';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useParams } from 'react-router-dom';
 
 import config from '../../../../package.json';
 import { PortalRoute } from '../../screens/routeUtils';
+import { getPathWithParticipant } from '../../utils/urlHelpers';
 
 import './SideNav.scss';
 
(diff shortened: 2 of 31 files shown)


src/web/components/Navigation/SideNav.tsx
```diff
@@ -16,14 +17,21 @@ function MenuItem({
   linkClass,
   isHidden,
 }: Pick<PortalRoute, 'path' | 'description' | 'isHidden'> & { linkClass?: string }) {
+  const { participantId } = useParams();
+  const pathWithParticipant = getPathWithParticipant(path, participantId);
```
alex-yau-ttd, line 21: We have to do this because the paths have `:participantId` in them - react-router is not smart enough to swap it out when defining the paths like this.

package.json
```diff
@@ -41,7 +41,7 @@
     "@types/react": "^18.2.54",
     "@types/react-dom": "^18.2.18",
     "@types/validator": "^13.11.10",
-    "axios": "^1.6.7",
+    "axios": "^1.7.4",
```
alex-yau-ttd, line 44: Fix CVE-2024-39338

src/web/utils/urlHelpers.ts
```diff
@@ -0,0 +1,31 @@
+export const parseParticipantId = (participantId: string | undefined) => {
+  if (!participantId) {
+    throw new Error('Participant ID is not defined');
```
ssundahlTTD, line 3: is there a reason that we're expecting undefined participant ids here?  And if we're expecting them, is there a better way of dealing with that than just throwing?

alex-yau-ttd: it should only happen if the user manually deletes the participantId from the URL, hence why I'm just throwing here. does that seem reasonable?

ssundahlTTD: sounds good.

ssundahlTTD (approved):

PR #503, opened 2024-08-23
Always provide participantId in API calls

alex-yau-ttd:
## What changed
- General idea: use `useParams()` hook to get `participantId` then populate the ParticipantContext. Then use the ParticipantContext from then on (rather than relying on 'current' or calling `useParams()` from every screen)
- Participant ID is never provided on initial log in, hence why we need to keep the call to `GetUsersDefaultParticipant()`
- Make `participantId` a required parameter where it used to be optional
- Remove support for 'current' in the participant middleware. Refactor to make it more readable. Remove unnecessary tests
- Update some TODOs

## Manual testing
1. Change the participantId in the URL to another participant you have access to (i.e. not the first participant for your user in the `usersToParticipantRoles` table). Access comes for free if you have the UID2 Support role:
    ```
    insert into dbo.userstoParticipantRoles (userId, participantId, userRoleId) values (<your userID>, <your participantId>, 3)
    ```
3. Perform operations in the portal, e.g. creating/editing API keys, modifying client side integration config, adding team members, etc.
4. Note all the API calls are using the participantId you currently have selected (via the URL), rather than just the first participant that your user belongs to.

## Quick demo

https://github.com/user-attachments/assets/a8a39f08-7e5a-4dec-a6ec-657773e10b6d

### DIFF
--- a/src/api/middleware/participantsMiddleware.ts
+++ b/src/api/middleware/participantsMiddleware.ts
@@ -10,14 +10,19 @@ const idParser = z.object({
   participantId: z.coerce.number(),
 });
 
-export const hasParticipantAccess = async (req: ParticipantRequest, res: Response, next: NextFunction) => {
+const enrichParticipant = async (req: ParticipantRequest, res: Response, next: NextFunction) => {
   const { participantId } = idParser.parse(req.params);
-  const traceId = getTraceId(req);
   const participant = await Participant.query().findById(participantId).withGraphFetched('types');
   if (!participant) {
     return res.status(404).send([{ message: 'The participant cannot be found.' }]);
   }
+  req.participant = participant;
+  return next();
+};
 
+const verifyUserAccessToParticipant = async (req: ParticipantRequest, res: Response) => {
+  const { participantId } = idParser.parse(req.params);
+  const traceId = getTraceId(req);
   const userEmail = req.auth?.payload?.email as string;
   const isUserUid2Support = await isUid2Support(userEmail);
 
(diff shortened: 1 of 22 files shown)


src/api/middleware/tests/participantsMiddleware.spec.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '../../../testHelpers/apiTestHelpers';
```
alex-yau-ttd, line 9: Remove tests associated with 'current' - test `hasParticipantAccess` directly

src/api/routers/participants/participantsRouter.ts
```diff
@@ -210,7 +210,13 @@ export function createParticipantsRouter() {
 
   participantsRouter.put('/', createParticipant);
 
-  participantsRouter.use('/:participantId', checkParticipantId);
+  participantsRouter.use('/:participantId', hasParticipantAccess);
+
+  participantsRouter.get('/:participantId', async (req: ParticipantRequest, res: Response) => {
```
alex-yau-ttd, line 215: This method sort of looks like it doesn't do anything, but the participant has been enriched by the middleware, so it now contains the fully loaded participant

src/api/middleware/participantsMiddleware.ts
alex-yau-ttd, line 1: Remove any functionality related to 'current' participant ID. Refactor to separate concerns between authorisation check and loading the participant graph.

src/web/utils/urlHelpers.ts
```diff
@@ -1,10 +1,10 @@
 export const parseParticipantId = (participantId: string | undefined) => {
```
alex-yau-ttd, line 1: @ssundahlTTD I changed my mind on this method - I'm now expecting an invalid `participantId` to be a genuine case when the user first signs in, see `ParticipantProvider.tsx`.

ssundahlTTD (approved):

ssundahlTTD (approved):

PR #507, opened 2024-08-28
Copy changes for root-level domains and mobile app IDs

alex-yau-ttd:
- Changed the success toast for CSTG to always use the term "saved", because there's not currently an easy way to tell how many are being added or deleted.
- Removed some unnecessary/inaccurate strings from the tests
- Fixed some more casing issues
- Improved some error messages

## Non-exhaustive demo
https://github.com/user-attachments/assets/bf6bf569-963e-4f74-9c6b-780135298181

https://github.com/user-attachments/assets/6ea439cf-7bab-4b46-af5f-09073ffe6e1a

### DIFF
--- a/src/web/components/ClientSideTokenGeneration/CstgAddDialog.spec.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgAddDialog.spec.tsx
@@ -28,7 +28,7 @@ describe('CstgAddDialog', () => {
         onOpenChange={() => {}}
         existingCstgValues={[]}
         cstgValueType={CstgValueType.Domain}
-        addInstructions='Add one or more domains.'
+        addInstructions=''
       />
     );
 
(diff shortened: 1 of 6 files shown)


src/web/components/ClientSideTokenGeneration/CstgAddDialog.spec.tsx
```diff
@@ -163,7 +163,7 @@ describe('CstgAddDialog', () => {
         onOpenChange={() => {}}
         existingCstgValues={[]}
         cstgValueType={CstgValueType.MobileAppId}
-        addInstructions='Please register the Android App ID, iOS/tvOS Bundle ID and iOS App Store ID.'
```
alex-yau-ttd, line 166: This string is no longer accurate and isn't even needed. Figured I would remove all of them in this file

ashleysmithTTD (approved):

PR #509, opened 2024-08-28
Enhance user middleware to cover edge case

alex-yau-ttd:
- Cover the edge case where the user has access to the requestor participant, but the target user does not belong to that participant
- Make the distinction clearer between the requesting user and the target user (the one that is being changed). Do the same in the tests
- Add a test to cover this case.

### DIFF
--- a/src/api/middleware/usersMiddleware.ts
+++ b/src/api/middleware/usersMiddleware.ts
@@ -44,10 +44,8 @@ const canUserAccessParticipant = async (
   participantId: number,
   traceId: string
 ) => {
-  const isRequestingUserUid2Support = await isUid2Support(requestingUserEmail);
-
   return (
-    isRequestingUserUid2Support ||
+    (await isUid2Support(requestingUserEmail)) ||
     (await isUserBelongsToParticipant(requestingUserEmail, participantId, traceId))
   );
 };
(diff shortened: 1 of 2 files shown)


src/api/middleware/usersMiddleware.ts
```diff
@@ -39,6 +39,19 @@ export const isUserBelongsToParticipant = async (
...
+  participantId: number,
+  traceId: string
+) => {
+  const isRequestingUserUid2Support = await isUid2Support(requestingUserEmail);
+
+  return (
+    isRequestingUserUid2Support ||
+    (await isUserBelongsToParticipant(requestingUserEmail, participantId, traceId))
+  );
+};
```
ashleysmithTTD, line 53: super nit: just a thought, fine with what you decide - it seemed a little weird to put the result of `isUid2Support` into a variable, but not do the same for `isUserBelongsToParticipant` - I would say it could be more consistent, either both variables or both calling the function in the return statement

ashleysmithTTD (approved):

src/api/middleware/usersMiddleware.ts
alex-yau-ttd: good point - I created this variable before splitting out the `canUserAccessParticipant` function - it's probably not necessary now. I'm putting them both in the return statement so we can benefit from an early return if the user is UID2 support

PR #514, opened 2024-09-02
Refactor some user invite code

alex-yau-ttd:
## Why?
https://github.com/IABTechLab/uid2-self-serve-portal/pull/516 will be changing the functionality to allow user invitations to still work when the user already belongs to another participant. I'm doing this refactoring to make it easier for me to add that functionality as well as make it easier to be reviewed. 

## What?
- There are no functional changes.
- Rename a function to avoid confusion with an upcoming function to send an invite to an existing user
- Split out the handler for inviting a new participant/user to `participantsUsers.ts`
- Extract a method out of the handler for creating a new participant. This started as more of an exercise but figured it was still worth merging.
- Wrapper method to assign keycloak user role. It's the only role we use today anyway.
- Remove an unused file

### DIFF
--- a/src/api/routers/participants/participantsRouter.ts
+++ b/src/api/routers/participants/participantsRouter.ts
@@ -8,7 +8,7 @@ import {
   Participant,
   ParticipantApprovalPartial,
   ParticipantDTO,
-  ParticipantStatus
+  ParticipantStatus,
 } from '../../entities/Participant';
 import { UserDTO } from '../../entities/User';
 import { siteIdNotSetError } from '../../helpers/errorHelpers';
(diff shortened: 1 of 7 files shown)


src/api/routers/participants/participantsRouter.ts
```diff
@@ -210,7 +202,7 @@ export function createParticipantsRouter() {
     }
   );
 
-  participantsRouter.put('/', createParticipant);
+  participantsRouter.put('/', handleCreateParticipant);
```
alex-yau-ttd, line 205: Renamed because I split the handler into two functions, one of which is called `createParticipant`

src/api/routers/participants/participantsRouter.ts
```diff
@@ -220,62 +212,7 @@ export function createParticipantsRouter() {
...
-        return res.sendStatus(201);
-      } catch (err) {
-        if (err instanceof z.ZodError) {
-          return res.status(400).send(err.issues);
-        }
-        throw err;
-      }
-    }
-  );
+  participantsRouter.post('/:participantId/invite', handleInviteUserToParticipant);
```
alex-yau-ttd, line 215: extracted this out - functionality is the same (for now)

src/api/routers/participants/participantsRouter.ts
```diff
@@ -57,21 +55,17 @@ import {
   updateParticipant,
   updateParticipantAndTypesAndApiRoles,
   UpdateSharingTypes,
-  UserParticipantRequest,
+  UserParticipantRequest
```
ashleysmithTTD, line 58: I'm not sure if we have different format settings, but you've got a lot of changes with the extra comma for the last import from a class.  When I go into my portal code and remove the comma and press save, it adds the comma back.  I don't think the format really matters, but we should probably all be on the same page with it so we don't end up going back and forth with comma edits in PRs

alex-yau-ttd: good catch! turns out my ALT+SHIFT+O command  to remove unused imports also removed the trailing commas. Saving the file added them back

ashleysmithTTD (approved):

PR #515, opened 2024-09-02
Align property names in `AddParticipantForm`

alex-yau-ttd:
Match property names in `AddParticipantForm` to the input names in `AddParticipantDialog`. This mismatch prevented defaults to be set on the contact inputs, which was slowing down developer iteration.

## Manual testing
**Still working like normal**

https://github.com/user-attachments/assets/6d6e57e6-86df-40fe-bc63-91719811024f

**With a default value set**
```
const formMethods = useForm<AddParticipantForm>({
  defaultValues: { siteIdType: 0, firstName: 'hello reviewer' },
});
```
https://github.com/user-attachments/assets/7c86ef32-2a6f-40e6-98a2-cbc2114f2c43

### DIFF
--- a/src/web/services/participant.ts
+++ b/src/web/services/participant.ts
@@ -174,9 +174,6 @@ export type UpdateParticipantForm = {
 };
 
 export type AddParticipantForm = {
-  contactFirstName: string;
-  contactLastName: string;
-  contactEmail: string;
   participantName: string;
   apiRoles: number[];
   participantTypes: number[];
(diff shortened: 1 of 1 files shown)


src/web/services/participant.ts
```diff
@@ -185,6 +182,9 @@ export type AddParticipantForm = {
   siteName?: string;
   jobFunction: string;
   crmAgreementNumber: string;
+  firstName: string;
+  lastName: string;
+  email: string;
```
ashleysmithTTD, line 187: Nit: maybe keep them at the top to preserve the order as before since it doesn't look like its in alphabetical order

ashleysmithTTD (approved):

src/web/services/participant.ts
alex-yau-ttd: I moved them to the bottom to align with `UpdateParticipantForm` (and the UI itself). I'm not too concerned about preserving order, so I think I'll merge as-is. thanks!

PR #516, opened 2024-09-03
Support user invites for multiple participants

alex-yau-ttd:
- Allow for existing user to be added to a brand new participant
- Allow for existing user to be added to an existing participant
- New email template for existing user being invited

### Inviting to existing participant

https://github.com/user-attachments/assets/ebeaeb42-7b77-4a35-bb62-8a642d25d4b2

### Inviting to new participant

https://github.com/user-attachments/assets/aff9b542-1678-432d-a6ad-b79ec1db8654

### New email 
image

### Future work
Support user deletion in the multiple-participant access world.

### DIFF
--- a/src/api/routers/participants/participantsCreation.ts
+++ b/src/api/routers/participants/participantsCreation.ts
@@ -6,7 +6,7 @@ import { AuditAction, AuditTrailEvents } from '../../entities/AuditTrail';
 import {
   Participant,
   ParticipantCreationPartial,
-  ParticipantStatus,
+  ParticipantStatus
 } from '../../entities/Participant';
 import { User, UserCreationPartial } from '../../entities/User';
 import { UserRoleId } from '../../entities/UserRole';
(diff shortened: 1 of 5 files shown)


src/api/routers/participants/participantsCreation.ts
```diff
@@ -16,68 +16,61 @@ import { getKcAdminClient } from '../../keycloakAdminClient';
...
 export async function validateParticipantCreationRequest(
   participantRequest: z.infer<typeof ParticipantCreationRequest>
 ) {
-  let errorMessage = null;
   const existingParticipant = await Participant.query().findOne(
     'name',
     participantRequest.participantName
   );
   if (existingParticipant) {
-    errorMessage = 'Duplicate participant name';
```
alex-yau-ttd, line 50: These two cases are now valid, so should not error

src/api/routers/participants/participantsCreation.ts
```diff
@@ -16,68 +16,61 @@ import { getKcAdminClient } from '../../keycloakAdminClient';
...
   participantData: z.infer<typeof ParticipantCreationAndApprovalPartial>
-) => {
-  await User.transaction(async (trx) => {
-    // create user
-    const newPortalUser = await User.query(trx).insertAndFetch(parsedUser);
+): Promise<Participant | undefined> => {
+  const participant = await User.transaction(async (trx) => {
+    let user = await findUserByEmail(parsedUser.email);
+    if (!user) {
+      user = await User.query(trx).insertAndFetch(parsedUser);
```
alex-yau-ttd, line 72: Only insert/update the user if they don't exist already. Otherwise leave them as-is (don't update their name/jobFunction)

src/api/routers/participants/participantsUsers.ts
```diff
@@ -33,39 +27,23 @@ const invitationParser = z.object({
...
-        {
-          email,
-          jobFunction,
-          firstName,
-          lastName,
-        },
-        participant!.id
-      );
-      await sendInviteEmailToNewUser(kcAdminClient, newUser);
+      await inviteUserToParticipant(userPartial, participant!, traceId);
```
alex-yau-ttd, line 46: This new method handles both cases (existing user & new user)

src/api/routers/participants/participantsCreation.ts
```diff
@@ -6,7 +6,7 @@ import { AuditAction, AuditTrailEvents } from '../../entities/AuditTrail';
 import {
   Participant,
   ParticipantCreationPartial,
-  ParticipantStatus,
+  ParticipantStatus
```
ashleysmithTTD, line 9: Same comment as the other PR with the difference in comma formatting

src/api/routers/participants/participantsCreation.ts
ashleysmithTTD: If this isn't an error message anymore, are we doing anything with the strings that get returned (and does the function care or use the different strings being returned)?  Should it just be a boolean function instead?

src/api/routers/participants/participantsUsers.ts
```diff
@@ -4,26 +4,20 @@ import { z } from 'zod';
...
   const { participant } = req;
   const users = await getAllUserFromParticipant(participant!);
   return res.status(200).json(users);
 }
 
-const invitationParser = z.object({
+export const UserInvitationParser = z.object({
   firstName: z.string(),
   lastName: z.string(),
   email: z.string(),
```
ashleysmithTTD, line 23: nit: Is the convention normally that the first word is capitalized? I see it changed from the first letter being lowercase to upper case

alex-yau-ttd: good point, I'd say most of them are actually camelCase: 
image

Looking at the [zod docs](https://zod.dev/), it looks like they use PascalCase for [objects](https://zod.dev/?id=objects) and camelCase for [primitives](https://zod.dev/?id=basic-usage). 

ChatGPT argues that PascalCase should be used for models or types that are intended to be reused across the application, while camelCase should be used for schemas that are defined within a specific, narrow scope, like a single function. I tend to agree with that.

Thinking about this more, it probably doesn't make the most sense to include `Parser` in the name either, since it's just a `ZodObject` which has a `parse` method on it.

I've actioned this for this instance - I'm happy to apply that convention more broadly in a separate PR /ticket if we think it's the right move.

thanks!

src/api/routers/participants/participantsCreation.ts
alex-yau-ttd: the method still returns an error message if there's an error or `null` if there is no error, so the functionality is still the same as before. If we change it to return a `boolean` then we'd lose the error message. I'm inclined to leave as-is unless you feel strongly? Perhaps my inital change/comment was confusing.

src/api/routers/participants/participantsUsers.ts
ashleysmithTTD: Definitely not urgent, but we could make a separate continuous improvement, quick win ticket for changing this at some point

ashleysmithTTD (approved):

PR #517, opened 2024-09-04
Single command for running portal

alex-yau-ttd:
- New powershell script that runs admin and portal in one-go. See README for more details
- Should take about 30s to run, for existing developers (i.e. MSSQL database container is not being created for the first time)
- Succeeds even if the docker images are already running

## Demo

https://github.com/user-attachments/assets/68cc4641-6d4e-4b66-a5d7-6d05f14ec83e

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "api": "run-p check-env-file api-watch",
     "check-env-file": "npx ts-node src/checkEnvFile.ts",
     "api-watch": "nodemon -e \".ts\" --exec npm run babel-node-api",
-    "babel-node-api": "npx kill-port 6540 && babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
+    "babel-node-api": "babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
     "build-api": "npx tsc --build ./tsconfig-api.json",
     "test-api": "npx jest --testTimeout=5000 --forceExit --runInBand --config=./src/api/jest.config.ts",
     "storybook": "storybook dev -p 6006",
--- a/run_portal.ps1
+++ b/run_portal.ps1
@@ -1,27 +1,41 @@
-# Change to uid2-admin directory, run docker compose and then maven compile
-Set-Location -Path "../uid2-admin" -ErrorAction Stop
-docker compose up -d
-if ($LASTEXITCODE -ne 0) {
-    Write-Error "Docker compose failed in uid2-admin"
+# Change to uid2-admin directory, and push location to stack
+Push-Location -Path "../uid2-admin" -ErrorAction Stop
+
+# Run admin docker & application
+try {
+    Write-Host "Running docker compose for uid2-admin..."
+    docker compose up -d
+    if ($LASTEXITCODE -ne 0) {
+        Write-Error "Docker compose failed in uid2-admin with exit code $LASTEXITCODE"
+        exit 1
+    }
+
+    Write-Host "Running admin application..."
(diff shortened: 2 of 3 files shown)


run_portal.ps1
```diff
@@ -0,0 +1,27 @@
+# Change to uid2-admin directory, run docker compose and then maven compile
+Set-Location -Path "../uid2-admin" -ErrorAction Stop
+docker compose up -d
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "Docker compose failed in uid2-admin"
+    exit 1
+}
+
+Start-Process "mvn" -ArgumentList "clean compile exec:java", "-Dvertx-config-path=conf/local-config.json", "-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory", "-Dlogback.configurationFile=conf/logback.xml" -WindowStyle Hidden
```
alex-yau-ttd, line 9: `Start-Process` runs the command in a separate terminal, allow the original terminal to continue exiting. I've decided to hide the separate terminal using `-WindowStyle Hidden` because it doesn't really add much value and is yet another terminal.

package.json
```diff
@@ -101,7 +101,7 @@
     "api": "run-p check-env-file api-watch",
     "check-env-file": "npx ts-node src/checkEnvFile.ts",
     "api-watch": "nodemon -e \".ts\" --exec npm run babel-node-api",
-    "babel-node-api": "npx kill-port 6540 && babel-node -r dotenv/config -x \".ts\" ./src/api/api.ts",
```
alex-yau-ttd, line 104: In my testing, this does not seem to be necessary when running `npm run dev`. It comes in handy when running the api and web app separately, but that does not seem to be a common workflow on the team.

run_portal.ps1
```diff
@@ -0,0 +1,27 @@
+# Change to uid2-admin directory, run docker compose and then maven compile
+Set-Location -Path "../uid2-admin" -ErrorAction Stop
+docker compose up -d
```
alex-yau-ttd, line 3: `-d` is for detached mode, which starts the containers in the background and leaves them running. This frees up the terminal so it can keep going.

Source: https://docs.docker.com/reference/cli/docker/compose/up/

run_portal.ps1
```diff
@@ -0,0 +1,41 @@
...
+try {
+    Write-Host "Running docker compose for uid2-admin..."
+    docker compose up -d
+    if ($LASTEXITCODE -ne 0) {
+        Write-Error "Docker compose failed in uid2-admin with exit code $LASTEXITCODE"
+        exit 1
+    }
+
+    Write-Host "Running admin application..."
+    Start-Process "mvn" -ArgumentList "clean compile exec:java", "-Dvertx-config-path=conf/local-config.json", "-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory", "-Dlogback.configurationFile=conf/logback.xml" -WindowStyle Hidden
```
alex-yau-ttd, line 14: Start-Process runs the command in a separate terminal, allow the original terminal to continue exiting. I've decided to hide the separate terminal using -WindowStyle Hidden because it doesn't really add much value and is yet another terminal.

This behaviour ends up being similar to docker's detached (-d) mode, which is used throughout this script.

run_portal.ps1
```diff
@@ -0,0 +1,41 @@
+# Change to uid2-admin directory, and push location to stack
+Push-Location -Path "../uid2-admin" -ErrorAction Stop
+
+# Run admin docker & application
+try {
+    Write-Host "Running docker compose for uid2-admin..."
+    docker compose up -d
```
alex-yau-ttd, line 7: -d is for detached mode, which starts the containers in the background and leaves them running. This frees up the terminal so it can keep going.

Source: https://docs.docker.com/reference/cli/docker/compose/up/

run_portal.ps1
alex-yau-ttd: There's no point in checking for the $LASTEXITCODE after this line, as it is run asynchronously so it won't return an exit code here anyway.

ssundahlTTD (approved):

PR #518, opened 2024-09-04
Support user deletion/removal for multiple participants

alex-yau-ttd:
## What changed
- No longer delete the keycloak user
- In the portal db, no longer mark the portal user with the deleted flag or change their email
- Instead, remove their access to the given participant by deleting the appropriate rows in the `UserToParticipantRole` table
- If the user no longer has access to any participants, revoke the keycloak role
- Provide a friendly UX when a user logs in but does not have access to any participants (they can still log out)
- Rename "delete" to "remove"
- Ensure that a pending account (participant not approved) does not result in a navigation loop
- Some refactoring + deletion of unnecessary code

## Automated tests
New test file for testing the user removal functionality. 
- Leverages the test in-memory db
- Manually calls the middleware to enrich the request
- After removing the user, the test either fetches the updated user to check their participants, or checks if a call to the keycloak service has been made.

### Single participant
1. Add a user using the Team Members page
2. Sign in to the user and accept terms (which will grant the `api-participant-member` keycloak role)
3. Switch back to your other user and delete them using the Team Members page
4. Observe that the user will still exist in the db, but they will not have access to any participants or the API
5. If you try to login with that user that no longer belongs to any participants, they will see the following error:
image

### Multiple participants
1. Add a user to multiple participants
2. Delete them from one participant
3. Note that they should still have full access to their other participants

### Standard successful login

https://github.com/user-attachments/assets/68bc3773-8b5b-499b-8404-f5e1cf01b9b8

### User who has been removed from all their participants

https://github.com/user-attachments/assets/84ee9816-ab39-4117-82fd-e186b462f54a

### User who has not yet accepted terms and conditions

https://github.com/user-attachments/assets/aa45f1f3-b714-4f7d-9cf7-a3db9b1414ca

### User who belongs to a participant that has not yet been approved

https://github.com/user-attachments/assets/b8d055b9-07d1-4f28-b0a4-b1d4a5fee181


src/api/jest.setup.ts
```diff
@@ -1,3 +1,5 @@
+import 'reflect-metadata';
```
alex-yau-ttd, line 1: This import is required by inversify for the tests for to run properly

src/api/middleware/tests/participantsMiddleware.spec.ts
```diff
@@ -4,29 +4,13 @@ import { Knex } from 'knex';
...
   createParticipant,
+  createParticipantRequest,
   createResponseObject,
   createUser,
 } from '../../../testHelpers/apiTestHelpers';
 import { UserRoleId } from '../../entities/UserRole';
-import { ParticipantRequest } from '../../services/participantsService';
 import { verifyAndEnrichParticipant } from '../participantsMiddleware';
 
-const createParticipantRequest = (
```
alex-yau-ttd, line 14: Extracted to `apiTestHelpers.ts` no other change.

src/api/middleware/tests/usersMiddleware.spec.ts
```diff
@@ -6,30 +6,11 @@ import {
...
   createResponseObject,
   createUser,
+  createUserParticipantRequest,
 } from '../../../testHelpers/apiTestHelpers';
-import { Participant } from '../../entities/Participant';
 import { UserRoleId } from '../../entities/UserRole';
-import { UserParticipantRequest } from '../../services/participantsService';
 import { verifyAndEnrichUser } from '../usersMiddleware';
 
-const createUserParticipantRequest = (
```
alex-yau-ttd, line 15: Extracted to `apiTestHelpers.ts` no other change.

src/api/services/kcUsersService.ts
```diff
@@ -73,19 +73,6 @@ export const updateUserProfile = async (
   );
 };
 
-export const deleteUserByEmail = async (kcAdminClient: KeycloakAdminClient, userEmail: string) => {
```
alex-yau-ttd, line 76: We don't delete keycloak users anymore.

src/api/services/userService.ts
```diff
@@ -14,16 +14,10 @@ import {
   constructAuditTrailObject,
   performAsyncOperationWithAuditTrail,
 } from './auditTrailService';
-import { deleteUserByEmail, updateUserProfile } from './kcUsersService';
+import { removeApiParticipantMemberRole, updateUserProfile } from './kcUsersService';
 import { UserParticipantRequest } from './participantsService';
 import { enrichUserWithIsApprover, findUserByEmail, UserRequest } from './usersService';
 
-export type DeletedUser = {
```
alex-yau-ttd, line 21: No longer used

src/testHelpers/apiTestHelpers.ts
```diff
@@ -81,14 +117,20 @@ export async function createUser({
   };
 
   const user = await User.query().insert(data);
-  const userToParticipantRolesData = participantToRoles?.map((item) => ({
-    ...item,
-    userId: user.id,
-    userRoleId: item.userRoleId ?? UserRoleId.Admin,
-  }));
-  if (userToParticipantRolesData) {
-    await UserToParticipantRole.query().insert(userToParticipantRolesData);
```
alex-yau-ttd, line 90: This was causing problems in the SQLite DB when trying to insert multiple rows, e.g. multiple participants for the same user. This is because it relies on batch insert which is only supported by Postgres and MSSQL.

src/web/components/TeamMember/TeamMember.tsx
alex-yau-ttd, line 1: Renaming only

src/web/components/TeamMember/TeamMemberRemoveDialog.tsx
alex-yau-ttd, line 1: Renaming only

src/api/services/userService.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+describe('User Service Tests', () => {
+  let knex: Knex;
+  let next: NextFunction;
+  let res: Response;
+
+  beforeEach(async () => {
+    knex = await TestConfigure();
+    next = jest.fn();
+    ({ res } = createResponseObject());
+    jest.spyOn(kcUsersService, 'removeApiParticipantMemberRole').mockResolvedValue();
```
alex-yau-ttd, line 25: I think mocking keycloak here is fine - I don't think it's crucial enough to the production behaviour to warrant calling a "real" keycloak and checking for actual client roles.

src/api/services/userService.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+        const targetParticipant = await createParticipant(knex, {});
+        const anotherParticipant = await createParticipant(knex, {});
+        const user = await createUser({
+          participantToRoles: [
+            { participantId: targetParticipant.id },
+            { participantId: anotherParticipant.id },
+          ],
+        });
+        const request = createUserParticipantRequest(user.email, targetParticipant, user.id);
+        await verifyAndEnrichUser(request, res, next);
```
alex-yau-ttd, line 44: We are calling the middleware here to enrich the request with more information. i.e. without this, the `{ user }` in the request is undefined. Don't love this but not sure of another approach that is better.

src/api/services/userService.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+        const userService = new UserService();
+        await userService.removeUser(request);
+
+        const userParticipantIds = await getParticipantIdsOfUser(user.email);
+        expect(userParticipantIds).not.toContain(targetParticipant.id);
+        expect(userParticipantIds).toContain(anotherParticipant.id);
+      });
+    });
+    it('user keycloak roles are unchanged', async () => {
+      const targetParticipant = await createParticipant(knex, {});
```
ashleysmithTTD, line 55: Can you reuse come code between these two tests, for example, defining `targetParticipant` , `anotherParticipant` and `user` outside of the it block?

src/api/services/userService.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+        });
+        const request = createUserParticipantRequest(user.email, participant, user.id);
+        await verifyAndEnrichUser(request, res, next);
+
+        const userService = new UserService();
+        await userService.removeUser(request);
+
+        expect(kcUsersService.removeApiParticipantMemberRole).toHaveBeenCalledWith(
+          expect.anything(),
+          user.email
```
ashleysmithTTD, line 99: Same comment here with reusing code of `participant` and `user`

src/web/services/userAccount.ts
```diff
@@ -74,7 +74,6 @@ export async function GetAllUsersOfParticipant(participantId: number) {
   }
 }
 
-// TODO: make this only remove the user from the given participant in UID2-3852
```
ashleysmithTTD, line 77: nit: should you change the error message below to 'Could not remove user' instead of delete

ashleysmithTTD:
Would there ever be a use case for removing a user from every participant at once? For example, if a someone leaves TTD or leaves an agency, and we are instructed to remove them from every participant they belong to?  Or in that case, would we request that each participant deletes this user themselves?

ssundahlTTD:
How can we ever get this person back if we need to for some reason?  Aside from directly updating the database?  Since they're still actually in the system, they can't be added again.  Or are we not concerned about this?

src/web/App.tsx
```diff
@@ -31,7 +32,11 @@ export function App() {
     else rootRef.current!.classList.remove('darkmode');
   };
 
+  const participantError = useParticipantCheck();
+  if (participantError) return participantError;
```
ssundahlTTD, line 36: This doesn't seem to work.  When I log in with a user with no participant access, I still get the ugly error message.

src/api/services/userService.spec.ts
ssundahlTTD: Or can you even just combine it all into one test and assert multiple things?  Seems inefficient to do the exact same setup in multiple tests.

src/api/services/userService.spec.ts
ssundahlTTD: same comment from me too.  :)

src/web/services/userAccount.ts
alex-yau-ttd: done!

src/api/services/userService.spec.ts
alex-yau-ttd: I did try factor out some shared setup but functions directly in the `describe` block cannot be awaited because jest [prevents async describes](https://github.com/jest-community/eslint-plugin-jest/blob/v25.7.0/docs/rules/valid-describe-callback.md). 

I've gone with Scott's suggestion to just combine them into fewer tests. My original intention was to keep the assertions separate but I think the tradeoff for readability is fine.

alex-yau-ttd:
> Would there ever be a use case for removing a user from every participant at once? For example, if a someone leaves TTD or leaves an agency, and we are instructed to remove them from every participant they belong to? Or in that case, would we request that each participant deletes this user themselves?

Good question - I hadn't considered that scenario. My stance is that participants themselves should be responsible for their team members. I don't think it's something we need to necessarily support in the UI any time soon (ever?). It would be pretty straight forward to handle manually in the db, and should be pretty rare. 

@ashleysmithTTD

alex-yau-ttd:
> How can we ever get this person back if we need to for some reason? Aside from directly updating the database? Since they're still actually in the system, they can't be added again. Or are we not concerned about this?

@ssundahlTTD this should be fixed now - I forgot to merge in my changes from https://github.com/IABTechLab/uid2-self-serve-portal/pull/516

src/api/services/userService.ts
```diff
@@ -83,11 +69,16 @@ export class UserService {
     );
 
     await performAsyncOperationWithAuditTrail(auditTrailInsertObject, traceId, async () => {
-      const kcAdminClient = await getKcAdminClient();
-      await Promise.all([
-        deleteUserByEmail(kcAdminClient, user?.email!),
-        user!.$query().patch(data),
-      ]);
+      await UserToParticipantRole.query()
+        .where('userId', '=', user!.id)
```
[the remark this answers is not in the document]
alex-yau-ttd: good shout, I've made that change in several places.

ssundahlTTD (approved):

PR #527, opened 2024-09-12
Developer experience improvements

alex-yau-ttd:
- Set up a proper dependency for keycloak on the db, so that keycloak only starts when the db is definitely up. It will make it slightly slower to start up but will be more reliable.
- Filter out paths with whitespace for metrics, to stop the API from logging the following error: `unable to capture route for prom-metrics: Error: argument must not contain whitespace`

## Manual testing
1. Run `./run_portal.ps1
2. Notice that the keycloak container only starts when the database is healthy.

https://github.com/user-attachments/assets/b79a2aff-bfef-44bc-998d-1a8757573db1

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,11 @@ services:
     environment:
       ACCEPT_EULA: 'yes'
       MSSQL_SA_PASSWORD: 'D3velopmentP0'
+    healthcheck:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -32,7 +37,8 @@ services:
       - 18080:8080
       - 19990:9990
     depends_on:
-      - database
```
alex-yau-ttd, line 35: I don't think this ever really worked

ssundahlTTD (approved):

PR #533, opened 2024-09-19
Participant Switcher UX Improvements

alex-yau-ttd:
Better design for short/long participant names

## Before

https://github.com/user-attachments/assets/f7bcfc4a-61f7-44af-ac3b-ea5c9be6e1e6

image

## After

https://github.com/user-attachments/assets/2551c0e4-bf0d-4cd1-95ea-0978e6d7f752

image

### Typical length name

image


ashleysmithTTD:
Could you add a border for the longer participant names too? The highlight covers the whole name now with your change, but there is no border space around it, like in the one-line participant names

PR #535, opened 2024-09-19
Split up `participantsRouter` file

alex-yau-ttd:
- Split up the `participantsRouter.ts` file based on domain/feature
- Logically group the routes, ensuring the correct middleware still gets run
- Refactoring only - no functional changes

## Why?
- We are about to change how permissions work, so this makes it easier to make those changes. i.e. it is easier to identify which middleware needs run against certain endpoints.

### DIFF
--- a/src/api/routers/participants/participantsRouter.ts
+++ b/src/api/routers/participants/participantsRouter.ts
@@ -1,675 +1,139 @@
-import { AxiosError } from 'axios';
 import express, { Response } from 'express';
-import { z } from 'zod';
 
 import { ApiRoleDTO } from '../../entities/ApiRole';
-import { AuditAction, AuditTrailEvents } from '../../entities/AuditTrail';
-import {
-  Participant,
-  ParticipantApprovalPartial,
-  ParticipantDTO,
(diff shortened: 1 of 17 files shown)


src/api/routers/participants/participantsRouter.ts
```diff
@@ -1,675 +1,139 @@
...
+import {
+  handleAddSharingPermission,
+  handleGetSharingPermission,
+  handleRemoveSharingPermission,
+  handleUpdateSharingTypes,
+} from './participantsSharingPermissions';
+import { handleGetSignedParticipants } from './participantsSigned';
+import { handleGetParticipantUsers, handleInviteUserToParticipant } from './participantsUsers';
+
+const handleUpdateParticipant = async (req: UserParticipantRequest, res: Response) => {
```
alex-yau-ttd, line 52: A few of these handlers still made sense to me to belong in this file, mainly because I couldn't think of a good file name that would indicate that they are directly related to the participant. Open to suggestions though.

ashleysmithTTD (approved):

PR #544, opened 2024-10-11
Participant Activity Logs

ssundahlTTD:
Admin and Support users can now view a participant's audit trail.  The menu option is in the top right navigation.

image


src/web/components/AuditTrail/AuditTrailTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+          title='No audit logs for this participant'
+        >
+          <span>There are no audit logs.</span>
+        </TableNoDataPlaceholder>
+      )}
+    </div>
+  );
+}
+
+export default function BusinessContactsTable(props: AuditLogTableProps) {
```
alex-yau-ttd, line 118: should probably change the name here ;)

src/web/components/AuditTrail/AuditTrail.scss
```diff
@@ -0,0 +1,73 @@
...
+    font-size: 0.875rem;
+    width: 100%;
+
+    thead {
+      font-weight: 700;
+      border: none;
+    }
+
+    td {
+      padding-top: 15px;
```
alex-yau-ttd, line 12: We could probably add 20px of padding here to space the columns out a bit better, i.e.
`padding: 15px 20px 15px 0;`
image

src/web/components/AuditTrail/AuditTrail.scss
```diff
@@ -0,0 +1,73 @@
...
+        color: green;
+      }
+    }
+  }
+
+  .audit-trail-table-header {
+    display: flex;
+    justify-content: end;
+    align-items: baseline;
+    padding-bottom: 6px;
```
alex-yau-ttd, line 38: We could make this 10px to align with the search on the sharing permissions screen

src/web/components/AuditTrail/AuditTrail.scss
```diff
@@ -0,0 +1,73 @@
...
+    .audit-trail-search-bar-container {
+      display: flex;
+      align-items: center;
+      border-bottom: 2px solid var(--theme-action);
+      padding: 6px 2px;
+      ::placeholder {
+        color: var(--theme-search-text);
+        opacity: 1;
+      }
+      margin-right: 20px;
```
alex-yau-ttd, line 54: We can probably remove this padding so that it lines up with the end of the table, like so:
image

src/web/components/AuditTrail/AuditTrailTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return (
+    <div className='audit-trail-table-container'>
+      <div className='audit-trail-table-header'>
+        <div className='audit-trail-table-header-right'>
+          <div className='audit-trail-search-bar-container'>
+            <input
+              type='text'
+              className='audit-trail-search-bar'
+              onChange={handleSearch}
+              placeholder='Search Audit Trail'
```
alex-yau-ttd, line 76: suggestion: use 'Search audit trail' to align with other searches (e.g. cstg tables)

src/web/utils/apiRoles.ts
```diff
@@ -5,3 +5,21 @@ export function sortApiRoles(apiRoles: ApiRoleDTO[]) {
...
+
+export function getRoleNamesByIds(ids: number[]) {
+  const names: string[] = ids.map((item) => {
+    switch (item) {
+      case 1:
+        return 'Mapper';
+      case 2:
+        return 'Generator';
+      case 3:
+        return 'ID Reader';
```
alex-yau-ttd, line 17: this should be 'Bidder'

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const getPrettyAuditDetails = (eventType: AuditTrailEvents, eventData: unknown) => {
+  const data = eventData as Record<string, unknown>;
+  const outputArray = [];
+  // eslint-disable-next-line guard-for-in
+  for (const key in data) {
+    let val = data[key];
+    // make data more human-readable
+    if (key.toLowerCase().indexOf('apiroles') > -1) {
```
alex-yau-ttd, line 24: using [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) might improve the readability here

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+      const roleData = val as string[];
```
alex-yau-ttd, line 25: I think it would be good to differentiate this from user roles. Perhaps we use `apiRoleData` instead.

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+      // some audit records use ids rather than names for the roles
+      if (roleData.length > 0 && !isNaN(Number(roleData[0]))) {
+        val = getRoleNamesByIds(val as number[]);
+      }
+    } else if (key.toLowerCase().indexOf('participanttypes') > -1) {
```
alex-yau-ttd, line 30: Perhaps prettifying of each audit record type could be extracted to a separate method, to improve readability? it would be nice if it was something like:
```
for (const key in data) {
  let val = data[key];
  const lowerCaseKey = key.toLowerCase();
  if (lowerCaseKey.includes('apiRoles')) {
    val = prettifyApiRoleAuditDetails(val);
  } else if (lowerCaseKey.includes('participantTypes')) {
    val = prettifyParticipantTypeAuditDetails(val);
  }
```

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+export const GetAuditLogs = async (participantId: number) => {
+  try {
+    const result = await axios.get<AuditTrailDTO[]>(`/participants/${participantId}/auditTrail`);
+    return result.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Could not get audit trail.');
+  }
+};
+
+export const getPrettyAuditDetails = (eventType: AuditTrailEvents, eventData: unknown) => {
```
alex-yau-ttd, line 17: `eventType` is unused here - we can probably remove it

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+  const data = eventData as Record<string, unknown>;
+  const outputArray = [];
+  // eslint-disable-next-line guard-for-in
+  for (const key in data) {
```
alex-yau-ttd, line 21: I think we can appease the eslint rule by using [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys):
`for (const key of Object.keys(data)) {`

src/web/services/auditTrailService.ts
```diff
@@ -0,0 +1,44 @@
...
+      }
+    } else if (key.toLowerCase().indexOf('participanttypes') > -1) {
+      const typeData = val as string[];
+      // some audit records use ids rather than names for the types
+      if (typeData.length > 0 && !isNaN(Number(typeData[0]))) {
+        val = mapClientTypeIdsToAdminEnums(val as number[]);
+      }
+    }
+    // convert key from camelCase
+    outputArray.push(
```
alex-yau-ttd, line 38: maybe this conversion deserves it's own method in `textHelpers.ts`?

src/api/routers/participants/participantsAuditTrail.ts
```diff
@@ -0,0 +1,21 @@
+import { Response } from 'express';
+
+import { UserRoleId } from '../../entities/UserRole';
+import { GetParticipantAuditTrail } from '../../services/auditTrailService';
+import { UserParticipantRequest } from '../../services/participantsService';
+
+export async function getAuditTrail(req: UserParticipantRequest, res: Response) {
```
alex-yau-ttd, line 7: nit: can we call this `handleGetAuditTrail` to match the convention in `participantsRouter.tsx`? My thinking is that methods that have a request and response as parameters should have the `handle` prefix.

src/web/services/auditTrailService.ts
ssundahlTTD: good idea.  changed.

src/web/services/auditTrailService.ts
ssundahlTTD: good suggestion. done.

src/api/routers/participants/participantsAuditTrail.ts
ssundahlTTD: I changed it.

src/web/services/auditTrailService.ts
ssundahlTTD: good idea.  I moved it there.

src/web/services/auditTrailService.ts
ssundahlTTD: I mostly agree and that's what I started with, but it feels unsustainable as we continue to add different audit content and it runs counter to the way we create the audit logs (untyped data).  We can maybe revisit this but I think it does what we want it to do.

src/api/routers/participants/participantsAuditTrail.ts
alex-yau-ttd: I see you changed it to `handleAuditTrail`. Can this be `handleGetAuditTrail`?

src/web/components/AuditTrail/AuditLog.tsx
```diff
@@ -5,12 +5,12 @@ import { getPrettyAuditDetails } from '../../services/auditTrailService';
...
 import './AuditTrail.scss';
 
-type AuditLogProps = Readonly<{
+type AuditTrailProps = Readonly<{
   log: AuditTrailDTO;
 }>;
 
-function AuditLog({ log }: AuditLogProps) {
-  const formattedEventData = getPrettyAuditDetails(log.event, log.eventData);
+function AuditLog({ log }: AuditTrailProps) {
```
alex-yau-ttd, line 12: This function is still called AuditLog - is that intentional?

src/web/components/AuditTrail/AuditLog.tsx
alex-yau-ttd, line 1: This file is still called AuditLog.tsx - is that intentional?

alex-yau-ttd (approved):

src/web/components/AuditTrail/AuditLog.tsx
ssundahlTTD: yeah, kindof.  AuditTrail was already singular.  I renamed it AuditTrailRow to make it easier to find.

src/web/components/AuditTrail/AuditLog.tsx
ssundahlTTD: yeah, kindof.  AuditTrail was already singular.  I renamed it AuditTrailRow to make it easier to find.

PR #549, opened 2024-10-20
Display team member user roles

alex-yau-ttd:
image

### DIFF
--- a/src/api/services/usersService.ts
+++ b/src/api/services/usersService.ts
@@ -2,7 +2,7 @@ import { Request } from 'express';
 
 import { Participant, ParticipantDTO } from '../entities/Participant';
 import { User, UserDTO } from '../entities/User';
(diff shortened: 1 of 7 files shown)


src/api/services/usersService.ts
```diff
@@ -87,9 +90,39 @@ export const enrichUserWithIsApprover = async (user: User) => {
     isApprover: userIsApprover,
   };
 };
-export const getAllUserFromParticipant = async (participant: Participant) => {
+
+const usersWithCurrentParticipantUserRoleNames = async (users: User[], participantId: number) => {
```
ashleysmithTTD, line 94: Is there a simpler function name for this? Its a little confusing

alex-yau-ttd: renamed to `mapUsersWithParticipantRoles`

PR #554, opened 2024-10-23
Team Members User Role Management UI

alex-yau-ttd:
- Ability to assign Admin or Operations role to users on the **Team Members** screen
- Prevent user from self-unassigning Admin role
- Improve style file

## Adding a new user

https://github.com/user-attachments/assets/6a59ee06-16a7-4305-987a-a53dd2f78a4f

## Adding an existing user to a different participant

https://github.com/user-attachments/assets/dfc50e48-2ff8-4d8f-9cd0-3a1cfd5d29fb

## Audit trail result for first participant

image

Currently just shows the IDs, but I think we will display the actual user role names in a separate ticket

## Prevent user from self-unassigning Admin role

https://github.com/user-attachments/assets/770e148c-69d9-4739-92a2-6a302e83230e

### DIFF
--- a/src/api/controllers/userController.ts
+++ b/src/api/controllers/userController.ts
@@ -126,6 +126,7 @@ export class UserController {
 
     if (req.auth?.payload?.email === user?.email) {
       res.status(403).send([{ message: 'You do not have permission to remove yourself.' }]);
+      return;
     }
 
     await this.userService.removeUser(req);
--- a/src/api/controllers/userController.ts
+++ b/src/api/controllers/userController.ts
@@ -13,6 +13,7 @@ import {
 
 import { TYPES } from '../constant/types';
 import { ParticipantStatus } from '../entities/Participant';
+import { UserRoleId } from '../entities/UserRole';
 import { getTraceId } from '../helpers/loggingHelpers';
 import { getKcAdminClient } from '../keycloakAdminClient';
 import {
@@ -21,7 +22,11 @@ import {
   sendInviteEmailToNewUser,
 } from '../services/kcUsersService';
 import { LoggerService } from '../services/loggerService';
-import { SelfResendInvitationSchema, UserService } from '../services/userService';
+import {
+  SelfResendInvitationSchema,
+  UpdateUserRoleIdSchema,
+  UserService,
+} from '../services/userService';
 import { SelfResendInviteRequest, UserRequest } from '../services/usersService';
 
 @controller('/users')
@@ -135,6 +140,16 @@ export class UserController {
 
   @httpPatch('/:userId')
   public async updateUser(@request() req: UserRequest, @response() res: Response): Promise<void> {
+    const { user } = req;
+    const userRoleData = UpdateUserRoleIdSchema.parse(req.body);
+    if (req.auth?.payload?.email === user?.email && userRoleData.userRoleId !== UserRoleId.Admin) {
+      res
+        .status(403)
+        .send([
+          { message: 'You do not have permission to unassign the Admin role from yourself.' },
+        ]);
+      return;
+    }
     await this.userService.updateUser(req);
     res.sendStatus(200);
   }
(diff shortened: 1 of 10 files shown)


src/api/routers/participants/participantsUsers.ts
```diff
@@ -30,6 +31,7 @@ export async function handleInviteUserToParticipant(req: UserParticipantRequest,
   try {
     const { participant, user } = req;
     const userPartial = userInvitationSchema.parse(req.body);
+    const userRoleIdData = UpdateUserRoleIdSchema.parse(req.body);
```
alex-yau-ttd, line 34: I'm using a separate schema here because the userRoleId is not part of the User table, whereas everything inside `userData` is. That means the `userData` can be used as a "partial user" to do inserts/updates directly on the User entity, while including `userRoleId` in that schema would break that.

src/api/services/userService.ts
```diff
@@ -105,21 +111,32 @@ export class UserService {
...
+        lastName: userData.lastName,
+        jobFunction: userData.jobFunction,
+        userRoleIds: userRoleData.userRoleId,
       },
       participant!.id
     );
 
     await performAsyncOperationWithAuditTrail(auditTrailInsertObject, traceId, async () => {
       const kcAdminClient = await getKcAdminClient();
+      await UserToParticipantRole.query()
```
alex-yau-ttd, line 124: couldn't find a better way to do this, given we can't update the table because the PK is all three columns.

src/api/services/usersService.ts
```diff
@@ -29,7 +29,7 @@ export interface SelfResendInviteRequest extends Request {
 
 export type UserWithIsApprover = UserDTO & { isApprover: boolean };
 export type UserWithCurrentParticipantRoleNames = UserDTO & {
-  currentParticipantUserRoleNames?: string[];
+  currentParticipantUserRoles?: UserRoleDTO[];
```
alex-yau-ttd, line 32: Changing this to include the whole object, so we can use those IDs

src/web/components/Input/RadioInput.scss
```diff
@@ -1,44 +1,45 @@
 .radio-group-root {
   display: flex;
-  gap: 10px;
+  gap: 1rem;
```
alex-yau-ttd, line 3: Made this gap a bit bigger

src/web/components/Input/RadioInput.scss
alex-yau-ttd, line 1: Some refactoring here to make better use of scss.

ashleysmithTTD:
I think it would be helpful from a UX perspective to provide a short description of Admin vs Operations role in the UI so they don't have to reference the docs to look for it.  Maybe just a tool tip in the Roles column header that has this info

src/api/controllers/userController.ts
```diff
@@ -126,6 +126,7 @@ export class UserController {
 
     if (req.auth?.payload?.email === user?.email) {
       res.status(403).send([{ message: 'You do not have permission to remove yourself.' }]);
+      return;
```
alex-yau-ttd, line 129: this prevents the user from removing themself cc: @ashleysmithTTD

src/api/controllers/userController.ts
```diff
@@ -134,6 +140,16 @@ export class UserController {
 
   @httpPatch('/:userId')
   public async updateUser(@request() req: UserRequest, @response() res: Response): Promise<void> {
+    const { user } = req;
+    const userRoleData = UpdateUserRoleIdSchema.parse(req.body);
+    if (req.auth?.payload?.email === user?.email && userRoleData.userRoleId !== UserRoleId.Admin) {
```
alex-yau-ttd, line 145: erroring out here so that the participant is not left with 0 admins.

ashleysmithTTD (approved):

src/api/services/userService.ts
```diff
@@ -121,23 +121,27 @@ export class UserService {
...
-        }),
-        user!.$query().patch(userData),
-        await UserToParticipantRole.query().insert({
-          userId: user!.id,
-          participantId: participant?.id!,
-          userRoleId: userRoleData.userRoleId,
-        }),
-      ]);
+
+      await UserToParticipantRole.transaction(async (trx) => {
```
alex-yau-ttd, line 125: Using a transaction here to ensure data consistency for the user/participant/role mapping

PR #556, opened 2024-10-23
Remove isApprover from user and add isUid2Support to user

ashleysmithTTD:
What Changed:
1. Removed `isApprover` field from User and add `isUid2Support` (and add/remove any functions that support this)
2. Renaming to `uid2Support` in several places
3. User can only access ManageParticipants and AuditTrail if they have the `uid2support` role (will add functionality for admin role to see the AuditTrail in the next PR)
4. Added rerouting back to home in Manage Participants and Audit Trail if user tries to go to the link without the correct role

Test Plan:

- Test that the Manage Participants screen and Audit Trail can be seen in the side nav
- Remove yourself from Approvers table and confirm this is still the case
- Remove your uid2support role
- Test that you cannot see Manage Participants screen or Audit Trail table
- Test that if you navigate to /manageParticipants without the role, it will redirect to home + error toast
- Test the same outcome, but navigate to /auditTrail without the role

### DIFF
--- a/src/web/components/Navigation/SideNav.tsx
+++ b/src/web/components/Navigation/SideNav.tsx
@@ -35,9 +35,9 @@ function MenuItem({
 
 export type SideNavProps = Readonly<{
   standardMenu: PortalRoute[];
-  uid2SupportMenu: PortalRoute[];
+  adminMenu: PortalRoute[];
 }>;
-export function SideNav({ standardMenu, uid2SupportMenu }: SideNavProps) {
+export function SideNav({ standardMenu, adminMenu }: SideNavProps) {
   return (
     <NavigationMenu className='side-nav'>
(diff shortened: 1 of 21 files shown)


src/web/components/Navigation/SideNav.tsx
```diff
@@ -35,9 +35,9 @@ function MenuItem({
 
 export type SideNavProps = Readonly<{
   standardMenu: PortalRoute[];
-  adminMenu: PortalRoute[];
```
ashleysmithTTD, line 38: Decided to rename this to `uid2SupportMenu` everywhere so we didn't get it confused with the 'admin' role since it is only the 'uid2support` role that can view this menu

alex-yau-ttd (approved):

PR #557, opened 2024-10-25
Add role checks for viewing audit trail and managing team members

ashleysmithTTD:
What Changed:
- Add a new workflow to get the roles of a user for the current participant when we have both the participantId and userId
- Added check in the PortalHeader to determine if we show Audit Trail or not based on the roles
- Added boolean variable to conditionally show certain management features on the Team Members page 
- Added tests for differences in the UI based on roles

Test Plan:

With uid2support role:
- test that you can view Manage Participants, view Audit Trail, and perform all team member functionality

With admin role:
- test that you cannot view Manage Participants, but can view Audit Trail and perform all team member functionality

With operations role:
- test that you cannot view Manage Participants, cannot view Audit Trail, and cannot add, edit, delete, or resend invite for a team member.  should not be able to see "Add team member" button and should not have an Actions column in their table


src/web/components/PortalHeader/PortalHeader.tsx
```diff
@@ -36,10 +39,35 @@ export function PortalHeader({
...
+  const [routes, setRoutes] = useState<PortalRoute[]>([
+    ParticipantInformationRoute,
+    TeamMembersRoute,
+    EmailContactsRoute,
+  ]);
+
+  useEffect(() => {
+    const getUserRolesForCurrentParticipant = async () => {
+      const user = LoggedInUser?.user;
+      if (user?.isUid2Support) {
```
ashleysmithTTD, line 51: Is it okay to assume UID2Support is all or nothing for participants - meaning a user with UID2Support either has that role for all their participants or none? If not, I can add a check below to also check if the userRoles that are returned contain `UserRoleId.UID2Support`

src/web/services/userAccount.ts
```diff
@@ -12,7 +13,7 @@ import { backendError } from '../utils/apiError';
 
 export type UserAccount = {
   profile: KeycloakProfile;
-  user: UserWithIsUid2Support | null;
+  user: UserWithIsUid2Support | UserWithCurrentParticipantRoleNames | null;
```
ashleysmithTTD, line 16: Is there a reason why we kept `UserWithisUid2Support` and `UserWithCurrentParticipantRoleNames` separate? Could we make one object that has the UserDTO + `isUid2Support` + `currentParticipantRoleNames` 

Then this could be the only type in UserAccount for LoggedInUser and it would make the code plus testing simpler I believe

src/web/screens/manageParticipants.tsx
```diff
@@ -46,7 +47,8 @@ function ManageParticipants() {
   const navigate = useNavigate();
 
   useEffect(() => {
-    if (!LoggedInUser?.user?.isUid2Support && participant) {
+    const user = LoggedInUser?.user as UserWithIsUid2Support;
+    if (user?.isUid2Support && participant) {
```
alex-yau-ttd, line 51: I believe the first condition is reversed, it should be: `if (!user?.isUid2Support && participant) {`

src/web/services/userAccount.ts
alex-yau-ttd: nah there's no reason - please fix it up :D

Ideally the type we use for `LoggedInUser` in the `CurrentUserContext` should have everything we need role-wise directly, so we won't need to do stuff like: `const user = LoggedInUser?.user as UserWithIsUid2Support;` in `manageParticipants.tsx`. 

I also realised that `UserWithCurrentParticipantRoleNames` also has the wrong name now, given it has the whole user role object, not just the name.

src/web/components/TeamMember/TeamMembersTable.tsx
```diff
@@ -37,6 +42,16 @@ function TeamMembersTableContent({
   const { sortData } = useSortable<UserDTO>();
   const sortedTeamMembers = sortData(teamMembers);
 
+  useEffect(() => {
+    const user = teamMembers.find((teamMember) => teamMember.id === LoggedInUser?.user?.id);
```
alex-yau-ttd, line 46: there's a case where this logic won't work where the user is a UID2 support user but not actually a team member for the given participant. Hopefully this is an easy fix where we also just check the user directly for the UID2 support role, i.e. something like:
```
useEffect(() => {
  const user = teamMembers.find((teamMember) => teamMember.id === LoggedInUser?.user?.id);
  const isAdminUser = user?.currentParticipantUserRoles?.find(
    (role) => role.id === UserRoleId.UID2Support || role.id === UserRoleId.Admin
  );
  if (isAdminUser || LoggedInUser?.user.isUid2Support) {
    setAllowTeamMemberActions(true);
  }
}, [LoggedInUser, teamMembers]);
```
(once we fix up the type of the LoggedInUser to include `isUid2Support`.)

alex-yau-ttd: also, I'm guessing you did this for a good reason, but it seems a bit strange to iterate through all the team members here when we already know who the user is. I'm guessing it's because the `LoggedInUser` does not have `currentParticipantUserRoles` whereas the list of `teamMembers` does. Per your comment below, it would be nice if `LoggedInUser` had all the user role stuff we needed so we didn't need to iterate through `teamMembers` here.

src/web/components/TeamMember/TeamMembersTable.tsx
```diff
@@ -28,7 +30,10 @@ function TeamMembersTableContent({
   onRemoveTeamMember,
   onUpdateTeamMember,
 }: TeamMembersTableProps) {
+  const { LoggedInUser } = useContext(CurrentUserContext);
+
   const [showTeamMemberDialog, setShowTeamMemberDialog] = useState<boolean>(false);
+  const [allowTeamMemberActions, setAllowTeamMemberActions] = useState<boolean>(false);
```
alex-yau-ttd, line 36: nit: perhaps `[showTeamMemberActions, setShowTeamMemberActions]` makes more sense here? Consistent with the other use state variable, and refers more to rendering than an actual permission check.

src/web/components/TeamMember/TeamMembersTable.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    'should include all manage team member functionality',
+    (manageRole) => {
+      mockUser.currentParticipantUserRoles = [manageRole];
+      userContextValue = createUserContextValueWithParticipantRoleName(mockUser);
+      renderTeamMembersTableWithContext(userContextValue, participantContextValue, [mockUser]);
+      const addTeamMemberButton = screen.getByRole('button', { name: 'Add Team Member' });
+      expect(addTeamMemberButton).toBeInTheDocument();
+      const actionsHeader = screen.getByRole('columnheader', { name: 'Actions' });
+      expect(actionsHeader).toBeInTheDocument();
+      expect(screen.getByTestId('action-cell')).toBeInTheDocument();
```
alex-yau-ttd, line 70: I think we can use a better RTL query here, so that we can align more closely to the user behaviour (per: https://testing-library.com/docs/queries/about/#priority). If we add `aria-label` to each of the action buttons that don't have a label, we can improve accessibility but then also use `getByLabelText()` instead of `getByTestId()`. 

e.g. For the pencil button, we can change it to:
```
<ActionButton
  onClick={onOpenChangeTeamMemberDialog}
  icon='pencil'
  aria-label='Edit Team Member'
/>
```
then test it via: 
```
expect(screen.getByLabelText('Edit Team Member')).toBeInTheDocument();
```

How's that sound?

src/web/components/PortalHeader/PortalHeader.tsx
```diff
@@ -36,10 +40,39 @@ export function PortalHeader({
...
-  if (LoggedInUser?.user?.isUid2Support) {
-    routes.push(AuditTrailRoute);
-  }
+  const [routes, setRoutes] = useState<PortalRoute[]>([
+    ParticipantInformationRoute,
+    TeamMembersRoute,
+    EmailContactsRoute,
+  ]);
+
+  useEffect(() => {
```
alex-yau-ttd, line 49: I think this `useEffect` would become unneccessary if `LoggedInUser` had all the user role info

src/web/components/PortalHeader/PortalHeader.spec.tsx
```diff
@@ -17,4 +43,35 @@ describe('Portal Header tests', () => {
...
+      const dropdownMenuButton = screen.getByRole('button');
+      await userEvent.click(dropdownMenuButton);
+      const auditTrailMenuItem = screen.getByRole('menuitem', { name: 'Audit Trail' });
+      expect(auditTrailMenuItem).toBeInTheDocument();
+    });
+
+    test('when user does not have uid2support role', async () => {
+      mockUser.isUid2Support = false;
+      renderPortalHeaderWithContext(userContextValue, participantContextValue);
+      const button = screen.getByRole('button');
```
alex-yau-ttd, line 72: nit: I think this should have the same variable name as in the previous test, for consistency, since they are referring to the same button.

src/web/components/PortalHeader/PortalHeader.spec.tsx
```diff
@@ -17,4 +43,35 @@ describe('Portal Header tests', () => {
...
+      participantContextValue = createParticipantContextValue(mockParticipant);
+      mockUser = createMockUser([mockParticipant] as Participant[]);
+      userContextValue = createUserContextValue(mockUser);
+    });
+
+    test('when user has uid2support role', async () => {
+      mockUser.isUid2Support = true;
+      renderPortalHeaderWithContext(userContextValue, participantContextValue);
+      const dropdownMenuButton = screen.getByRole('button');
+      await userEvent.click(dropdownMenuButton);
```
alex-yau-ttd, line 64: while calling methods on `userEvent` itself is currently supported, the [recommendation ](https://testing-library.com/docs/user-event/intro/)(and what we usually do) is to use:
`const user = userEvent.setup();` then `user.click(dropdownMenuButton);`,

src/api/routers/participants/participantsRouter.ts
```diff
@@ -136,8 +140,10 @@ export function createParticipantsRouter() {
   participantsRouter.get('/:participantId/auditTrail', handleAuditTrail);
 
   participantsRouter.get('/:participantId/users', handleGetParticipantUsers);
+  participantsRouter.get('/:participantId/:userId', handleGetUserRolesForCurrentParticipant);
```
alex-yau-ttd, line 143: let's change this endpoint so that it's less confusing. e.g. an example call would be `http://localhost:3000/api/participants/8/3` - it is hard to know what the 8 and the 3 mean. Whereas the other endpoints in the `participantUsersRouter` are of the form `/:participantId/users/:userId` e.g. http://localhost:3000/api/participants/8/users/3`. This may require moving the endpoint to `participantUsersRouter.ts`

src/api/routers/participants/participantsRouter.ts
```diff
@@ -136,8 +140,10 @@ export function createParticipantsRouter() {
...
   const participantUsersRouter = createParticipantUsersRouter();
   participantsRouter.use('/:participantId/users', participantUsersRouter);
+  participantsRouter.use('/:participantId/:userId', participantUsersRouter);
```
alex-yau-ttd, line 146: I don't think this line is required - it implies that the request is handled by the `participantUsersRouter` but it is handled inside `participantUsers.ts`. 

If we follow the API design I [commented ](https://github.com/IABTechLab/uid2-self-serve-portal/pull/557/files#r1818358705)just above, then line 145 should be sufficient to make it route properly.

src/api/services/usersService.ts
```diff
@@ -111,6 +111,17 @@ export const getAllUsersFromParticipantWithRoles = async (participant: Participa
   return mapUsersWithParticipantRoles(usersWithParticipants, participant.id);
 };
 
+export const getUserRolesForCurrentParticipant = async (participant: Participant, user: User) => {
+  const userToParticipantRoles = await UserToParticipantRole.query()
+    .where('participantId', participant.id)
+    .where('userId', user.id);
+
+  const userRoles = await UserRole.query();
+  return userToParticipantRoles.map((role) => {
```
alex-yau-ttd, line 120: The return type of this method is `Promise<(UserRole | null)[]>`, partly because of the null coalescing operator on the line below. I think we can avoid both of these things by doing the following:
```
  return userToParticipantRoles
    .map((role) => userRoles.find((userRole) => userRole.id === role.userRoleId))
    .filter((role): role is UserRole => role !== undefined);
```

src/web/components/TeamMember/TeamMembersTable.tsx
ashleysmithTTD: Done!

src/web/components/TeamMember/TeamMembersTable.spec.tsx
ashleysmithTTD: Fixed!

src/web/components/PortalHeader/PortalHeader.tsx
ashleysmithTTD: Yep - not needed

src/web/components/PortalHeader/PortalHeader.spec.tsx
ashleysmithTTD: Done

src/web/components/PortalHeader/PortalHeader.tsx
```diff
@@ -34,14 +35,26 @@ export function PortalHeader({
...
+  const userRolesForCurrentParticipant = user?.participants?.find(
+    (p) => p.id === parsedParticipantId
+  )?.currentUserRoleIds;
 
   const routes = [ParticipantInformationRoute, TeamMembersRoute, EmailContactsRoute];
-  if (LoggedInUser?.user?.isUid2Support) {
+
+  if (
+    user?.isUid2Support ||
+    [UserRoleId.UID2Support, UserRoleId.Admin].some((role) =>
```
alex-yau-ttd, line 50: Is there any danger of `isUid2Support` not being populated properly? I guess we could technically just check `UserRoleId.Admin` here since we are checking `isUid2Support` in the line above

src/web/components/TeamMember/TeamMembersTable.tsx
```diff
@@ -28,15 +30,33 @@ function TeamMembersTableContent({
...
   const onOpenChangeTeamMemberDialog = () => {
     setShowTeamMemberDialog(!showTeamMemberDialog);
   };
 
-  const { sortData } = useSortable<UserDTO>();
+  const { sortData } = useSortable<UserResponse>();
   const sortedTeamMembers = sortData(teamMembers);
 
+  useEffect(() => {
+    const isUserAdminOrSupport =
```
alex-yau-ttd, line 50: We do this logic here and in `PortalHeader.tsx` - perhaps it could be extracted into a helper that receives a `UserWithParticipantRoles` and returns a boolean

src/web/components/TeamMember/TeamMembersTable.tsx
```diff
@@ -28,15 +30,33 @@ function TeamMembersTableContent({
...
 
   const onOpenChangeTeamMemberDialog = () => {
     setShowTeamMemberDialog(!showTeamMemberDialog);
   };
 
-  const { sortData } = useSortable<UserDTO>();
+  const { sortData } = useSortable<UserResponse>();
   const sortedTeamMembers = sortData(teamMembers);
 
+  useEffect(() => {
```
alex-yau-ttd, line 49: I don't think we actually need this `useEffect` or the corresponding `useState`. We can just use a local variable:
```
  const isUserAdminOrSupport =
    LoggedInUser?.user?.isUid2Support ||
    [UserRoleId.UID2Support, UserRoleId.Admin].some((role) =>
      userRolesForCurrentParticipant?.includes(role)
    );
```

Then change all the conditional rendering to use `isUserAdminOrSupport` directly.

src/web/components/TeamMember/TeamMembersTable.tsx
```diff
@@ -28,15 +30,33 @@ function TeamMembersTableContent({
   onRemoveTeamMember,
   onUpdateTeamMember,
 }: TeamMembersTableProps) {
+  const { LoggedInUser } = useContext(CurrentUserContext);
+  const { participant } = useContext(ParticipantContext);
+  const userRolesForCurrentParticipant = LoggedInUser?.user?.participants?.find(
```
alex-yau-ttd, line 35: nit: maybe this could also be extracted into a helper method

src/web/components/PortalHeader/PortalHeader.tsx
```diff
@@ -34,14 +35,26 @@ export function PortalHeader({
   logout,
 }: Readonly<PortalHeaderProps>) {
   const { participantId } = useParams();
+  const parsedParticipantId = parseParticipantId(participantId);
```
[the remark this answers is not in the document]
ashleysmithTTD: Fixed

src/web/components/TeamMember/TeamMembersTable.tsx
ashleysmithTTD: Changed!

src/web/components/TeamMember/TeamMembersTable.tsx
ashleysmithTTD: Extracted it into the same helper method as above - `isUserAdminorSupport`

src/web/components/PortalHeader/PortalHeader.spec.tsx
```diff
@@ -17,4 +44,67 @@ describe('Portal Header tests', () => {
...
+    let mockUser: UserWithParticipantRoles;
+
+    beforeAll(() => {
+      mockParticipant = createMockParticipant();
+      participantContextValue = createParticipantContextValue(mockParticipant);
+      mockUser = createMockUser([mockParticipant] as Participant[]);
+      userContextValue = createUserContextValue(mockUser);
+    });
+
+    it('should show Audit Trail when user has uid2support role', async () => {
```
alex-yau-ttd, line 62: nit: the existing tests use `test` while the new tests use `it`. They do the same thing but we should probably just use the same one everywhere (at least stick to one within the same file)

src/web/services/userAccount.ts
```diff
@@ -26,11 +23,11 @@ export type InviteTeamMemberForm = {
 export type UpdateTeamMemberForm = Omit<InviteTeamMemberForm, 'email'>;
 
 export type UserPayload = z.infer<typeof UserCreationPartial>;
-export type UserResponse = UserWithCurrentParticipantRoleNames;
+export type UserResponse = UserWithParticipantRoles;
```
alex-yau-ttd, line 26: Do we think this `UserResponse` still makes sense? I find it a bit confusing - it's used in a lot of places that aren't directly calling an API and getting a response. 

At the very least, perhaps we could replace a bunch of usages with `UserWithParticipantRoles`. But perhaps we should remove this type altogether - I think it's just adding a layer of abstraction that we don't need.

src/web/components/PortalHeader/PortalHeader.spec.tsx
ashleysmithTTD: Fixed - i like `it` better

src/web/components/TeamMember/TeamMembersTable.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+    userContextValue = createUserContextValue(mockUser);
+    renderTeamMembersTableWithContext(userContextValue, participantContextValue, [mockUser]);
+    const addTeamMemberButton = screen.getByRole('button', { name: 'Add Team Member' });
+    expect(addTeamMemberButton).toBeInTheDocument();
+    const actionsHeader = screen.getByRole('columnheader', { name: 'Actions' });
+    expect(actionsHeader).toBeInTheDocument();
+    expect(screen.getByLabelText('Edit Team Member')).toBeInTheDocument();
+  });
+
+  it('should not include any manage team member functionality when user does not have uid2support', () => {
```
alex-yau-ttd, line 99: nit: do we strictly need this test? it seems essentially the same as `should not include any manage team member functionality`

alex-yau-ttd (approved):

src/web/components/TeamMember/TeamMembersTable.spec.tsx
ashleysmithTTD: It is the same expected result, but since we're not really checking the support role within the roles array, I think it's still useful to test the direct `user.IsUid2Support` field.  If anything, we probably don't need to one above to also check UserRoleId.Uid2Support

ashleysmithTTD: Actually - I have to take out that check, or the test will fail with the change to `isUserAdminOrSupport` - so now I've updated the test above to just test with the Admin role

PR #559, opened 2024-10-29
API user role checks

alex-yau-ttd:
- Replace any `isApprover` API permission checks with `isUid2Support`.
- Protect certain endpoints from users without Admin or UID2 Support roles.
- Move shared auth checks to `userRoleMiddleware.ts`

## Manual testing

1. With the UID2 Support role, you should be able to do anything in the portal, despite not being in the Approvers table (though the Participants Awaiting Approval still relies on the Approvers table)
2. Remove yourself from the UID2 Support role, giving yourself only Admin
3. Modify the conditions in the front end code so that the team member management, manage participant, and audit trail screens are all reachable by applying [show-all-restricted-UI-actions.patch](https://github.com/user-attachments/files/17551746/show-all-restricted-UI-actions.patch). Alternatively call the endpoints via Postman
4. Admin should still be able to do everything expect Manage Participants
5. Remove yourself from Admin, giving yourself only Operations
6. You should no longer be able to manage (add/edit/remove/reinvite) team members or view audit trails

### Example: Removing a team member when the user only has `Operations`

https://github.com/user-attachments/assets/fd1e3f97-5788-433d-9058-39dc6bea84c8

Note: the front-end code has been modified so that the action buttons still appear.


src/api/routers/participants/participantsRouter.ts
```diff
@@ -88,24 +88,26 @@ export function createParticipantsRouter() {
...
   participantsRouter.put('/', handleCreateParticipant);
 
   participantsRouter.use('/:participantId', verifyAndEnrichParticipant);
 
   participantsRouter.get('/:participantId', handleGetParticipant);
   participantsRouter.get('/:participantId/apiRoles', handleGetParticipantApiRoles);
-  participantsRouter.post('/:participantId/invite', handleInviteUserToParticipant);
   participantsRouter.put('/:participantId/completeRecommendations', handleCompleteRecommendations);
 
+  participantsRouter.post('/:participantId/invite', isAdminCheck, handleInviteUserToParticipant);
```
alex-yau-ttd, line 108: I grouped these together because they require the same permissions

src/api/routers/participants/participantsRouter.ts
```diff
@@ -88,24 +88,26 @@ export function createParticipantsRouter() {
...
+  participantsRouter.get('/:participantId/auditTrail', isAdminCheck, handleGetAuditTrail);
```
alex-yau-ttd, line 109: I grouped these two together because they require the same permissions

src/testHelpers/apiTestHelpers.ts
```diff
@@ -57,6 +57,9 @@ export const createUserParticipantRequest = (
     params: {
       userId: targetUserId,
     },
+    headers: {
+      traceId: faker.string.uuid(),
```
alex-yau-ttd, line 61: A test was erroring without this

src/api/routers/participants/participantsAuditTrail.ts
```diff
@@ -1,19 +1,10 @@
 import { Response } from 'express';
 
-import { UserRoleId } from '../../entities/UserRole';
 import { GetParticipantAuditTrail } from '../../services/auditTrailService';
 import { UserParticipantRequest } from '../../services/participantsService';
 
-export async function handleAuditTrail(req: UserParticipantRequest, res: Response) {
-  const { participant, user } = req;
-  const userParticipant = user?.participants?.find((item) => item.id === participant?.id);
-  if (
```
alex-yau-ttd, line 10: This auth check has moved to the middleware, so we don't need it here anymore.

src/api/middleware/userRoleMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (!(await isUid2Support(req.auth?.payload?.email as string))) {
+    return res.status(403).json({
+      message: 'Unauthorized. You do not have the necessary permissions.',
+      errorHash: req.headers.traceId,
+    });
+  }
+  next();
+};
+
+export const isAdminOrUid2SupportCheck: Handler = async (req: ParticipantRequest, res, next) => {
```
ashleysmithTTD, line 27: It feels a little redundant to have `isUid2Support` and `isAdminOrUid2Support` - my opinion would be to have this function be `isAdminCheck` and then when you need both, you'd call `isUid2SupportCheck` || `isAdminCheck`.

src/api/routers/participantUsersRouter.ts
```diff
@@ -9,7 +10,7 @@ const createParticipantUsersRouter = () => {
   const participantUsersRouter = express.Router({ mergeParams: true });
   const userController = new UserController(new UserService(), new LoggerService());
 
-  participantUsersRouter.use('/:userId', verifyAndEnrichUser);
+  participantUsersRouter.use('/:userId', verifyAndEnrichUser, isAdminOrUid2SupportCheck);
```
ashleysmithTTD, line 13: Why does `isAdminOrUid2SupportCheck` need to go here? Wouldn't this line need to be hit if the user is Operations?

src/api/middleware/tests/userRoleMiddleware.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      const participant = await createParticipant(knex, {});
+      const user = await createUser({
+        participantToRoles: [{ participantId: participant.id, userRoleId: UserRoleId.UID2Support }],
+      });
+      const userParticipantRequest = createUserParticipantRequest(user.email, participant, user.id);
+
+      await isAdminOrUid2SupportCheck(userParticipantRequest, res, next);
+
+      expect(res.status).not.toHaveBeenCalled();
+      expect(next).toHaveBeenCalled();
```
ashleysmithTTD, line 73: I think it would still be readable, but with less lines, if you condense these two tests using it.each() and just switch out which `userRoleId` is set when creating the user.

src/api/middleware/userRoleMiddleware.ts
alex-yau-ttd: As I mentioned in stand-up, I can't find a good way to enforce this conditional logic in handlers. Using `||` doesn't really work because if the first check succeeds then we just call `next()` which seems to skip the next check. Alternatively, placing it as an additional handler argument makes it an "and" condition which isn't what we want. 

I'll leave as-is for now.

src/api/routers/participantUsersRouter.ts
alex-yau-ttd: All the routes in this file require `Admin` or `UID2 Support` - this is what this line is enforcing. Perhaps it is not clear enough - I guess we want to avoid the dev having to pay attention to the order of lines here. I have just updated it so that each endpoint explicitly has the middleware check.

src/api/middleware/tests/userRoleMiddleware.spec.ts
alex-yau-ttd: done! less lines isn't always the goal but happy to do it here

PR #561, opened 2024-10-29
Remove participantId from user table

ashleysmithTTD:
What Changed:

- Migration to remove `participantId` column and remove foreign key on participants
- New SQL query in ReadMe for approving participant

Test Plan:

- Create a test account and verify that the query works for approving your own participant during set up
- Run this query below in integ and prod to make sure all participantId/userId combos have been migrated over to `userToParticipantRoles`
 ```
SELECT *
FROM [dbo].[users]
WHERE deleted = 0 AND participantId is not null AND NOT EXISTS (
    SELECT 1
    FROM dbo.usersToParticipantRoles
    WHERE usersToParticipantRoles.userId = users.id AND usersToParticipantRoles.participantId = users.participantId
);
```
- Verify that running the latest migration removes the participantId column successfully and removes the foreign key from participants db


src/database/migrations/20241029012630_RemoveParticipantIdUserTable.ts
alex-yau-ttd, line 1: nit: perhaps `RemoveParticipantIdFromUserTable.ts` for the file name for clarity?

alex-yau-ttd (approved):

PR #563, opened 2024-10-31
names rather than ids in audit trail.  index on participantId in audit table

ssundahlTTD:
names rather than ids in audit trail.  index on participantId in audit table.

This helps the audit trail to be more human-readble.

### DIFF
--- a/src/api/services/participantsService.ts
+++ b/src/api/services/participantsService.ts
@@ -284,7 +284,6 @@ export const updateParticipant = async (participant: Participant, req: UserParti
   } = updateParticipantSchema.parse(req.body);
   const { user } = req;
   const traceId = getTraceId(req);
-  const updatedParticipantId = Number(req.params.participantId);
 
   const auditTrailInsertObject = constructAuditTrailObject(
     user!,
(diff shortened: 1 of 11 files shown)


src/web/components/AuditTrail/AuditTrailTable.tsx
alex-yau-ttd, line 1: sorry, new idea: it probably makes sense to have the `Date` column now be first in the table. That seems to be what most audit trails do.

src/api/services/participantsService.ts
```diff
@@ -279,17 +284,19 @@ export const updateParticipant = async (participant: Participant, req: UserParti
...
       action: AuditAction.Update,
-      apiRoles,
+      apiRoles: getRoleNamesByIds(apiRoles),
       participantName,
-      participantTypes: participantTypeIds,
+      participantTypes: mapClientTypeIdsToAdminEnums(participantTypeIds),
       crmAgreementNumber,
-    }
+    },
+    updatedParticipantId
```
alex-yau-ttd, line 299: I intentionally left off the `participantId` from this action, because it probably shouldn't be shown to participants. e.g. it would expose things like crmAgreementNumber. 

Happy to discuss if you disagree!

alex-yau-ttd (approved):

PR #568, opened 2024-11-04
Remove approvers table

ashleysmithTTD:
What Changed:

- Migration to delete Approvers table, plus remove code relating to approvers table or approvers mock data
- update `getParticipantsAwaitingApproval` to no longer use email or participant types since only uid2 support will see this
- Update the new participant request email to go to those with uid2support role, rather than being an approver
- Update ReadME when setting up portal 

Test Plan:
- Run this query in prod and integ to verify that all approvers have been migrated over as users. 
```
SELECT 
    a.email
FROM 
    approvers a
LEFT JOIN 
    users u ON a.email = u.email
LEFT JOIN 
    usersToParticipantRoles upr ON u.id = upr.userId AND upr.userRoleId = 3
WHERE 
    upr.userId IS NULL;
```

- Verify that in the Approved Participants table, the correct approver information still is displayed
- Verify that all participants with awaiting approval status are showing in the Manage Participants table
- Test sending a new participant request and make sure it goes to the emails of UID2 support role 
- Test that the new query in the ReadME for the user support role works


README.md
```diff
@@ -297,26 +297,20 @@ The following steps describe the minimal steps required to successfully log in t
...
-insert into dbo.approvers (displayName, email, participantTypeId)
-values
-(@displayName, @email, 1),
-(@displayName, @email, 2),
-(@displayName, @email, 3),
-(@displayName, @email, 4)
-
-select * from dbo.approvers where email = @email
+insert into dbo.usersToParticipantRoles (userId, participantId, userRoleId)
+select u.id, upr.participantId, 3
```
alex-yau-ttd, line 310: nit: perhaps we can avoid the magic number by doing:
```
declare @email as nvarchar(256) = 'example@example.com';
declare @uid2SupportRoleId as int = 3;
...
select u.id, upr.participantId, @uid2SupportRoleId

```

README.md
```diff
@@ -297,26 +297,20 @@ The following steps describe the minimal steps required to successfully log in t
 1. If using Visual Studio Code, you may need to set `FAST_REFRESH=false` in your .env file
 1. Azure Data Studio is an adequate program for manipulating the SQL Server database
 
-### Admin screens/routes
+### UID2 Support Screens/Routes
```
alex-yau-ttd, line 300: This renaming breaks the link on [line 290](https://github.com/IABTechLab/uid2-self-serve-portal/blob/main/README.md?plain=1#L290).

alex-yau-ttd (approved):

README.md
ashleysmithTTD: good call - updated this

PR #569, opened 2024-11-04
Fix participant no access when single participant

ashleysmithTTD:
Quick bug fix to show something different than the Participant Switcher if the user navigates to a participant they don't have access to, but only belong to one participant.

Test Plan:

- Make sure navigation with the switcher still works when a user has multiple participans
- Update your user to only have access to one participant and check the UX when navigating to a participant in the URL that doesn't exist or they don't have access to (example: http://localhost:3000/participant/123333/home)

### DIFF
--- a/src/web/components/Navigation/NoParticipantAccessView.tsx
+++ b/src/web/components/Navigation/NoParticipantAccessView.tsx
@@ -1,7 +1,7 @@
 import { useLocation, useNavigate } from 'react-router-dom';
 
 import { UserWithParticipantRoles } from '../../../api/services/usersService';
-import { getPathWithParticipant } from '../../utils/urlHelpers';
+import { getPathWithParticipant, parseParticipantId } from '../../utils/urlHelpers';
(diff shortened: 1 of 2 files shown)


src/web/components/Navigation/NoParticipantAccessView.tsx
```diff
@@ -1,17 +1,52 @@
...
+export function NoParticipantAccessView({ user }: NoParticipantAccessViewProps) {
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const onBackToParticipant = () => {
+    let participantId;
+    if (user?.participants) {
+      participantId = user.participants[0].id;
+    } else {
+      participantId = localStorage.getItem('lastSelectedParticipantId');
```
ssundahlTTD, line 22: shouldn't it be the first option to get the participant from localstorage?

ashleysmithTTD: Switched the option order!

ssundahlTTD (approved):

PR #570, opened 2024-11-04
Do not show UI tour with updates for new users

ashleysmithTTD:
What Changed:
- Added a check for local storage to see if user has nothing in their local storage, then do not show the UI tour since they will be a new user and everything is new to them 

*Note: this check relies on local storage, meaning if a user logs in on a different device or happens to clear their local storage, they would not see the tour steps, but I think it's safe to assume the case is rare where they would log in on a different device or clear their local storage and be needing to see new features for the first time*

Test Plan:
- Create an account and log in as that user for the first time (make sure there is nothing in local storage for tour data from your normal user)
- Make sure you don't see the UI tour

### DIFF
--- a/src/web/components/SiteTour/tourHelpers.spec.ts
+++ b/src/web/components/SiteTour/tourHelpers.spec.ts
@@ -4,7 +4,7 @@ import { createMockParticipant, createMockUser } from '../../../testHelpers/data
 import { createUserContextValue } from '../../../testHelpers/testContextProvider';
 import { compareVersions } from './tourHelpers';
 import { tourSteps } from './tourSteps';
-import { GetTourSteps,tourStorageKey } from './tourStorage';
+import { GetTourSteps, tourStorageKey } from './tourStorage';
 
 describe('testing tour storage helper functions', () => {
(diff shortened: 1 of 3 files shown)


src/web/components/SiteTour/tourHelpers.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('testing tour storage helper functions', () => {
     mockUser.currentParticipantUserRoles = [{ id: UserRoleId.Admin, roleName: 'Admin' }];
     mockUser.participants = [mockParticipant, mockParticipant2] as Participant[];
     mockParticipant.currentUserRoleIds = [UserRoleId.Admin];
-    localStorage.removeItem(tourStorageKey);
+    localStorage.setItem(tourStorageKey, JSON.stringify({ seenForVersions: ['0.1.0'] }));
```
ashleysmithTTD, line 33: Setting it to an early version so it shows all the steps are present - can't leave it empty or it won't show any tour steps

ssundahlTTD (approved):

PR #582, opened 2024-11-21
Remove request account feature

ashleysmithTTD:
Test Plan:

- check that the login page has no request account link


src/web/screens/dashboard.tsx
```diff
@@ -16,7 +16,6 @@ import { ParticipantInformationRoute } from './participantInformation';
 import { PortalRoute } from './routeUtils';
 import { SharingPermissionsRoute } from './sharingPermissions';
 import { TeamMembersRoute } from './teamMembers';
-import { TermsOfServiceRoute } from './termsOfService';
```
ashleysmithTTD, line 19: This route/screen is no longer needed because terms and conditions is now a dialog that pops up when the user logs in for the first time

ashleysmithTTD:
Besides this code PR, we will also turn off User Registration in keycloak so the user could never access a registration page through a URL

alex-yau-ttd:
> Besides this code PR, we will also turn off User Registration in keycloak so the user could never access a registration page through a URL

I'm hoping this is also do-able via the `realm-export.json` file. At least they should remain consistent

alex-yau-ttd (approved):

ashleysmithTTD:
> > Besides this code PR, we will also turn off User Registration in keycloak so the user could never access a registration page through a URL
> 
> I'm hoping this is also do-able via the `realm-export.json` file. At least they should remain consistent

Oh good call - I need to change `registrationAllowed` to `false` in `realm-export.json`

PR #589, opened 2024-12-20
 Password policy improvements

ashleysmithTTD:
What Changed:
- Remove list of password requirements
- Add password blacklist
- Remove text for "active your account"

expiredpassword


keycloak/themes/uid2-theme/login/template.ftl
```diff
@@ -45,7 +45,8 @@
     <div class="centralize-content">
     <#-- App-initiated actions should not see warning messages about the need to complete the action -->
     <#-- during login.                                                                               -->
-    <#if displayMessage && message?has_content && (message.type != 'warning' || !isAppInitiatedAction??)>
+    <br />
+    <#if displayMessage && message?has_content && message.type != 'warning' && !isAppInitiatedAction??>
```
ashleysmithTTD, line 49: just taking out the parentheses fixed the logic here and no longer shows the "activate your account" message but still shows the error messages if someone inputs the wrong password

src/web/App.tsx
```diff
@@ -24,12 +24,11 @@ function AppContent() {
   const { LoggedInUser } = useContext(CurrentUserContext);
   const { participant } = useContext(ParticipantContext);
   const isLocalDev = process.env.NODE_ENV === 'development';
-  const location = useLocation();
 
   if (LoggedInUser?.user?.participants!.length === 0) {
     return <ErrorView message='You do not have access to any participants.' />;
   }
-  if (location.pathname !== '/account/create' && LoggedInUser && !participant) {
+  if (LoggedInUser && !participant) {
```
ashleysmithTTD, line 31: this account/create page does not exist anymore and should be removed. randomly found this when testing stuff

keycloak/themes/uid2-theme/login/login-update-password.ftl
```diff
@@ -70,6 +66,43 @@
...
                 </div>
             </div>
+
+
+        <script type="text/javascript">
+            let blacklistedPasswords = [];
+
+            function loadBlacklist() {
+                // txt file of common passwords recommended to blacklist by NIST
+                fetch('https://raw.githubusercontent.com/danielmiessler/SecLists/master/Passwords/Common-Credentials/10-million-password-list-top-1000000.txt')
```
ashleysmithTTD, line 76: pulling from this file, but we can also store the txt file in our code if that makes sense too.  i went with this option because i didnt want to take up space in the code for a huge file, but it would be an issue if this URL changes

ssundahlTTD: I don't love relying on a 3rd party to be up in order for our blacklist to work.  You can just save the file locally and load it from there, too.

ssundahlTTD (approved):

PR #593, opened 2025-02-11
Clean up code around test api keys and roles

ashleysmithTTD:


ashleysmithTTD:
> Looks great - though I do wonder if we could go a bit further - there's still a few instances where we are manually defining these in the code - mainly in stories but also in another spec file. Is this within scope here?
> 
> image

Yeah definitely - i'll go back and narrow it down some more

alex-yau-ttd (approved):

PR #596, opened 2025-02-18
clean up problems tab in vs code

ssundahlTTD:
- changed spelling errors to hints
- fixed the one actual problem
-  exhaustively opened all files to check for problems (the file has to be open in the editor)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -44,5 +44,6 @@
     "uncopied",
     "undici",
     "unrelate"
-  ]
+  ],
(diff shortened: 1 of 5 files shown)


ashleysmithTTD:
Do you think its beneficial to add the common "words" that cspell isnt going to recognize like "ssportal" and "knex" to a file? that way it makes it easier, no matter if the spelling issues are showing up as a hint or error, to distinguish between a spelling error that we want to change and one we want to ignore because its correct, just not a normal word?

ashleysmithTTD (approved):

PR #597, opened 2025-02-28
Add super user role

ashleysmithTTD:
What Changed:
- Added super user role with id=4
- Added functions and checks around super user
- Renamed some values to be more general with support + super user
- Set up the Super User route, and its section in navigation menu as an empty array for now


src/api/services/uid2SupportService.ts
ashleysmithTTD, line 1: We are not using the one function in this file anywhere so i removed it

src/api/services/userService.ts
```diff
@@ -37,15 +37,15 @@ export class UserService {
...
       const allParticipants = await getAllParticipants();
-      userWithUid2Support.participants = allParticipants;
+      userWithSupportRoles.participants = allParticipants;
     }
-    userWithUid2Support.participants = userWithUid2Support?.participants?.sort((a, b) =>
+    userWithSupportRoles.participants = userWithSupportRoles?.participants?.sort((a, b) =>
       a.name.localeCompare(b.name)
     );
-    return userWithUid2Support;
+    return userWithSupportRoles;
```
ashleysmithTTD, line 48: renamed to encompass UID2 support and super user

src/api/services/usersService.ts
```diff
@@ -28,6 +28,7 @@ export interface SelfResendInviteRequest extends Request {
 
 export type UserWithParticipantRoles = UserDTO & {
   isUid2Support: boolean;
+  isSuperUser?: boolean;
   currentParticipantUserRoles?: UserRoleDTO[];
 };
```
ashleysmithTTD, line 33: i dont think isSuperUser needs to be a required field

PR #598, opened 2025-02-28
admin user management page

ssundahlTTD:
- new page for administrators to manage users


src/web/contexts/SortableTableProvider.tsx
```diff
@@ -58,8 +58,13 @@ export function SortableProvider<TData extends {}>({ children }: { children: Rea
     (a: TData, b: TData): number => {
       if (!sortKey) return 0;
 
-      if (a[sortKey] > b[sortKey]) return sortOrder === 'asc' ? 1 : -1;
-      if (a[sortKey] < b[sortKey]) return sortOrder === 'asc' ? -1 : 1;
+      const aSort =
```
ssundahlTTD, line 61: current sorting doesn't combine lowercase and capital letters

src/api/routers/businessContactsRouter.ts
```diff
@@ -18,14 +18,16 @@ export function createBusinessContactsRouter() {
 
   businessContactsRouter.get('/', async (req: ParticipantRequest, res: Response) => {
     const { participant } = req;
-    const businessContacts = await participant!.$relatedQuery('businessContacts');
```
ssundahlTTD, line 21: the compiler was pretty unhappy about this $relatedQuery.

src/api/routers/adminRouter.ts
```diff
@@ -0,0 +1,18 @@
...
+  return res.status(200).json(userList);
+};
+
+export function createAdminRouter() {
+  const adminRouter = express.Router();
+
+  adminRouter.get('/users', isSuperUserCheck, handleGetAllUsersAdmin);
+
+  return adminRouter;
+}
```
ashleysmithTTD, line 18: I dont love the use of the word `admin` for this router, its a good name, but I think it may be confusing since we have a role called `Admin` that doesn't match what this router is for. could we change it to superUser or a different name? Also same for `handleGetAllUsersAdmin` maybe could be renamed to a reference to being for user management instead

src/api/services/usersService.ts
```diff
@@ -198,3 +198,8 @@ export const inviteUserToParticipant = async (
     await createUserInPortal(userPartial, participant!.id, userRoleId);
   }
 };
+
+export const getAdminUserList = async () => {
+  const userList = await User.query().where('deleted', 0).orderBy('email');
+  return userList;
+};
```
ashleysmithTTD, line 205: sounds like it is only getting the users with the admin role, probably needs renaming

src/web/services/userAccount.ts
```diff
@@ -104,3 +104,12 @@ export async function SetTermsAccepted() {
...
 }
+
+export async function GetAllUsersAdmin() {
+  try {
+    const userResponse = await axios.get<UserDTO[]>('/admin/users');
+    return userResponse.data;
+  } catch (e: unknown) {
+    throw backendError(e, 'Unable to get admin user list.');
+  }
+}
```
ashleysmithTTD, line 115: needs renaming to not imply looking for users with admin role

src/api/routers/adminRouter.ts
ssundahlTTD: Makes sense.  I changed it to "manage."

PR #599, opened 2025-03-04
allow locking users

ssundahlTTD:
- migration added for locked row on user table.  Run `npm run knex:migrate:latest`
- disallow access to locked users
- allow superusers to lock user from the user management page
- do not return locked users in queries outside of the user management page.

Capture

https://github.com/user-attachments/assets/a8e054ec-f36f-44fc-a5ed-dd4e93e8b2d0


ashleysmithTTD:
Testing out the functionality, could you add to the instructions 'search user by email, first name or last name' since search of job type or accepted terms doesnt work

ashleysmithTTD:
Would there be any reason to add delete user functionality from this UI also? Or just lock?

ashleysmithTTD (approved):

PR #601, opened 2025-03-08
dyanmic imports of keycloak variables

ashleysmithTTD:


src/api/services/kcUsersService.ts
```diff
@@ -1,9 +1,21 @@
...
+async function loadKeycloakRequiredActionAlias() {
+  const module = await import(
+    '@keycloak/keycloak-admin-client/lib/defs/requiredActionProviderRepresentation'
+  );
+  RequiredActionAlias = module.RequiredActionAlias;
+}
+
+loadKeycloakRequiredActionAlias()
+  .then(() => {})
+  .catch(() => {});
```
ashleysmithTTD, line 17: I can't throw an error from here or use error logger because it requires a trace id so just left it blank

ssundahlTTD (approved):

PR #602, opened 2025-03-08
Show user permissions and user audit trail

ashleysmithTTD:
Test Plan:

-Test that viewing a UID2 support person does not show all participants
-Test that viewing a user with only Admin and Operations shows their participants.  Test a user with a single participant and with multiple participants (make sure there is a dateApproved column not null when testing)
-Test that viewing a user with no audit logs works
-Test viewing your own user that probably has a lot of audit logs that scrolling and paging works


src/api/middleware/usersMiddleware.ts
```diff
@@ -63,14 +63,6 @@ export const enrichUserWithSupportRoles = async (user: User) => {
   };
 };
 
-export const enrichUserWithSuperUser = async (user: User) => {
-  const userIsSuperUser = await isSuperUser(user.email);
-  return {
-    ...user,
-    isSuperUser: userIsSuperUser,
-  };
-};
```
ashleysmithTTD, line 72: We're not using this function at all, this was a mistake from the last ticket

src/api/services/managementService.ts
```diff
@@ -9,7 +9,10 @@ import { UserParticipantRequest } from './participantsService';
 import { findUserByEmail } from './usersService';
 
 export const getAllUsersList = async () => {
-  const userList = await User.query().where('deleted', 0).orderBy('email');
+  const userList = await User.query()
+    .where('deleted', 0)
+    .orderBy('email')
+    .withGraphFetched('userToParticipantRoles');
```
ashleysmithTTD, line 15: pulling all the roles the user has as well

src/api/routers/participants/participantsAuditTrail.ts
```diff
@@ -3,7 +3,7 @@ import { Response } from 'express';
 import { GetParticipantAuditTrail } from '../../services/auditTrailService';
 import { UserParticipantRequest } from '../../services/participantsService';
 
-export async function handleGetAuditTrail(req: UserParticipantRequest, res: Response) {
+export async function handleGetParticipantAuditTrail(req: UserParticipantRequest, res: Response) {
```
ashleysmithTTD, line 6: renamed to be more clear since we have a regular get audit trail now too

src/web/components/UserManagement/UserPartcipantsTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+        </tbody>
+      </table>
+      {userParticipants.length === 0 && (
+        <TableNoDataPlaceholder
+          icon={<img src='/document.svg' alt='email-icon' />}
+          title='No Participants'
+        >
+          <span>This user does not belong to any participant.</span>
+        </TableNoDataPlaceholder>
+      )}
```
ashleysmithTTD, line 57: this case above should not happen but good to have just in case

src/web/components/UserManagement/UserPartcipantsTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+}>;
+
+function UserParticipantRow({ participantName, roleName }: UserParticipantRowProps) {
+  return (
+    <tr>
+      <td className='participant-name'>{participantName ?? 'Cannot find participant name'}</td>
+      <td className='role-name'>{roleName ?? 'Cannot find participant role'}</td>
+    </tr>
+  );
+}
```
ashleysmithTTD, line 21: not having `participantName` or `roleName` should not happen but good to include just in case

src/api/services/managementService.ts
ssundahlTTD: do we still need to do this when we're loading all of this on loading the dialog?

PR #611, opened 2025-03-19
Fix some UX issues with adding logout button on error 

ashleysmithTTD:
What Changed:

- Add a logout button to PortalErrorBoundary 
- Add functionality to store `lastSelectedParticipantId` as an object with `userId` as the key and `participantId` as the value in the case that mutliple users are logging in to the same device
- Remove "Back to Your Participant" button if user is locked (and/or has no participants)

Test Plan:
- Remove `api-participant-member` role from your keycloak participant, log back into the portal and make sure you can log out and log back in even though it will error
- Log in as your user, log back in as another user who does not have access to your last selected participant and make sure it does not break the UI saying the user does not have access.
- Lock yourself, then login or refresh the page and back sure it just says "You do not have access" and there is no "Back to my participant" button

### DIFF
--- a/src/web/components/Navigation/NoParticipantAccessView.tsx
+++ b/src/web/components/Navigation/NoParticipantAccessView.tsx
@@ -19,10 +19,7 @@ export function NoParticipantAccessView({ user }: NoParticipantAccessViewProps) 
     const lastSelectedParticipantIds = (JSON.parse(
       localStorage.getItem('lastSelectedParticipantIds') ?? '{}'
     ) ?? {}) as UserIdParticipantId;
-    let participantId;
-    if (user) {
-      participantId = lastSelectedParticipantIds[user?.id];
-    }
+    let participantId = user ? lastSelectedParticipantIds[user.id] : undefined;
     if (!participantId && user?.participants && user?.participants.length > 0) {
       participantId = user.participants[0].id;
     }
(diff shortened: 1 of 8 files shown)


src/web/utils/PortalErrorBoundary.tsx
```diff
@@ -1,9 +1,27 @@
...
+
+  return (
+    <div className='portal-error-boundary'>
+      <button className='logout-button' type='button' onClick={logout}>
+        Log Out
+      </button>
+      <ErrorView errorId={errorId} errorHash={errorHash} />
+    </div>
+  );
 }
```
ashleysmithTTD, line 26: I replicated all instances of where we see `<ErrorView />` and `PortalErrorComponent` is the only one where the top-right menu that contains the logout button is not present.  So it just makes sense to add the logout button to this one component.

src/web/components/Navigation/NoParticipantAccessView.tsx
```diff
@@ -40,7 +42,8 @@ export function NoParticipantAccessView({ user }: NoParticipantAccessViewProps)
             <ParticipantSwitcher noInitialValue />
           </div>
         </>
-      ) : (
+      )}
+      {user?.participants?.length === 1 && (
         <div>
           <button className='small-button' type='button' onClick={onBackToParticipant}>
             Back to Your Participant
```
ashleysmithTTD, line 49: These are the changes to not display any ability to navigate to a participant if the user is locked (aka has 0 participants associated with them)

ssundahlTTD (approved):

PR #622, opened 2025-06-13
add service id to admin api calls

ssundahlTTD:
image


ashleysmithTTD:
I think you need to add this stuff to `createAuditLogData` to actually send it through

ashleysmithTTD (approved):

PR #637, opened 2025-06-20
Sas UI d2 4130 keycloak error spam

ssundahlTTD:
removed the custom welcome theme for keycloak.  it never worked and we don't actually care.


src/api/services/adminServiceClient.ts
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable camelcase */
 import axios, { AxiosError, AxiosRequestConfig } from 'axios';
-import crypto from 'crypto';
```
ssundahlTTD, line 3: no longer being used

ashleysmithTTD (approved):

PR #640, opened 2025-06-24
extract schemas into their own file as part of ESM upgrade

ashleysmithTTD:


ssundahlTTD (commented):
is there a good explanation of why these need to be moved?

ssundahlTTD (approved):

PR #641, opened 2025-06-24
direct imports of jest for esm upgrade

ashleysmithTTD:


src/testHelpers/apiTestHelpers.ts
```diff
@@ -16,13 +17,20 @@ type ParticipantToUserRoleDTO = Partial<
...
+
+  const json = jest.fn<JsonFn>((_body: unknown) => res);
+  const send = jest.fn<SendFn>((_body: unknown) => res);
+  const status = jest.fn<StatusFn>((_code: number) => res);
+
+	res.json = json as typeof res.json;
+	res.send = send as typeof res.send;
+	res.status = status as typeof res.status;
+
   return { res, json, send, status };
```
ssundahlTTD, line 34: This is all a little awkward.  we add all these properties onto the response and also return them all separately from the function.  looks like the code only ever uses res and json.  we could probably just return res and get json from it in the tests that need it.

ashleysmithTTD: `res.status` and `res.json` are both used in the `usersMiddleware.spec.ts`  so we do need all of them.  But i think i was able to clean the function up more than it was before - we dont need the casts, but we do need the arguments and their types inside `jest.fn()`

ssundahlTTD (approved):

PR #643, opened 2025-06-25
Extract shared SCSS for search bar

eiman-eltigani-ttd:
Ended up creating a seperate stylesheet for the search bar as the SearchBar component that already exists is a separate search bar style used in pop ups such as Add Participant Dialogue + in testing I believe. This one required a bit of refactoring which is why I wanted to break it up into a draft first because the other shared styles (i.e., .action-cell) will be smaller changes since they are already established in the styles folder. 

Main thing I wanted feedback on:
- Adding the extra classnames reduces redundancy in the scss but makes the classNames long within the component. I could remove the original className since most are no longer in use but didn't know if I should keep them in for organizational/readability purposes.

Based on your feedback/comfortability with this approach, I'll move forward with the remaining changes!

### DIFF
--- a/src/web/components/AuditTrail/AuditTrailTable.tsx
+++ b/src/web/components/AuditTrail/AuditTrailTable.tsx
@@ -67,8 +67,8 @@ function AuditTrailTableComponent({ auditTrail }: AuditTrailTableProps) {
 
   return (
     <div className='audit-trail-table-container'>
-      <div className='search-bar-table-header'>
-        <div className='search-bar-table-header-right'>
+      <div className='table-header'>
+        <div className='table-header-right'>
           <div className='search-bar-container'>
             <input
               type='text'
--- a/src/web/styles/search.scss
+++ b/src/web/styles/search.scss
@@ -1,15 +1,3 @@
-.search-bar-table-header {
-  display: flex;
-  justify-content: end;
-  align-items: baseline;
-  padding-bottom: 10px;
-}
-
-.search-bar-table-header-right {
-  display: flex;
-  justify-content: right;
-}
-
 .search-bar-container {
   display: flex;
   align-items: center;
--- a/src/web/components/ApiKeyManagement/KeyItem.scss
+++ b/src/web/components/ApiKeyManagement/KeyItem.scss
@@ -1,7 +1,7 @@
-.action-cell {
+
   .warning-icon {
     color: var(--theme-warning);
     padding: 0 10px;
     height: 23px;
   }
-}
+
--- a/src/web/components/ClientSideTokenGeneration/CstgTable.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgTable.tsx
@@ -223,7 +223,7 @@ export function CstgTable({
             </div>
           )}
         </div>
-        <div className='search-bar-table-header-right'>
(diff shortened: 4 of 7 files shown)


src/web/components/AuditTrail/AuditTrailTable.tsx
```diff
@@ -11,6 +11,7 @@ import { TableNoDataPlaceholder } from '../Core/Tables/TableNoDataPlaceholder';
 import AuditTrailRow from './AuditTrailRow';
 
 import './AuditTrail.scss';
+import '../../styles/search.scss';
```
ashleysmithTTD, line 14: you shouldn't need to explicitly import anything from the styles folder.  it will catch them regardless.

src/web/App.scss
```diff
@@ -6,6 +6,8 @@
   @import 'styles/tables';
   @import 'styles/text';
   @import 'styles/buttons';
+  @import 'styles/search.scss';
+  @import 'styles/forms.scss';
```
ashleysmithTTD, line 10: you shouldn't need the `.scss` here like above how they do not have them

src/web/styles/search.scss
```diff
@@ -0,0 +1,38 @@
...
+  display: flex;
+  justify-content: end;
+  align-items: baseline;
+  padding-bottom: 10px;
+}
+
+.search-bar-table-header-right {
+  display: flex;
+  justify-content: right;
+}
```
ashleysmithTTD, line 11: These two pieces are not explicitly part of search bar - for example if you see in `CtsgTable.tsx`, the div for `search-bar-table-header-right` encompasses both the search bar and the add cstg value button.  Its still good to extract this code, but I think it would make more sense to put `.search-bar-table-header` and `search-bar-table-header-right` in `tables.scss`

src/web/components/Search/SearchBar.tsx
```diff
@@ -107,7 +107,7 @@ export function SearchBarFormInput<
         <FontAwesomeIcon icon='search' className='search-icon' />
         <input
           type='text'
-          className={clsx('search-bar-input', inputClassName)}
+          className={clsx('search-bar-form-input', inputClassName)}
           aria-label={inputName}
           aria-invalid={error ? 'true' : 'false'}
           {...rest}
```
ashleysmithTTD, line 113: if it does not affect anything and you delete `search-bar-form-input` in the file above, you can also revert this to `search-bar-input` here

src/web/components/Search/SearchBar.scss
```diff
@@ -37,7 +37,7 @@
     border: 1px solid var(--theme-input-border);
   }
 
-  .search-bar-input {
+  .search-bar-form-input {
     width: 100%;
   }
```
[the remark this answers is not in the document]
eiman-eltigani-ttd: I thought that was the case but there was actually two components (SearchBarFormInput and SearchBarInput)  in SearchBar.tsx that both had the the 'search-bar-input' className. The SearchBarInput component is used on the 'Sharing Permissions' page that we checked and that remained the same. The SearchBarFormInput is in the Add Participant Dialogue in the 'Manage Participants' page, and that one does require the 100% width or else the input (like the actual letters, not the container) stops half way.

PR #644, opened 2025-06-26
upgrade storybook to version 9 (and ESM-compatible)

ashleysmithTTD:
What Changed:
- no importing from @storybook/react directly, must choose a framework (like Webpack) to choose from
- no more `composeStories` function (there is one in testing-react but not very compatible with ESM)
- no importing something defined in a `.spec` file into a `.stories` file


src/api/tests/apiKeyService.spec.ts
```diff
@@ -2,7 +2,7 @@ import {
   allApiRoles,
   Generator,
   Mapper,
-} from '../../web/components/ApiKeyManagement/KeyHelper.spec';
+} from '../../web/components/ApiKeyManagement/KeyHelper';
 import { ApiRoleDTO } from '../entities/ApiRole';
 import { validateApiRoles } from '../services/apiKeyService';
```
ashleysmithTTD, line 8: Storybook did not like that these reusable variables were defined in a spec file rather than just a regular .ts file.  Moved them and updating imports

src/setupTests.ts
```diff
@@ -4,13 +4,8 @@
...
-// eslint-disable-next-line import/no-extraneous-dependencies
-import { setProjectAnnotations } from '@storybook/react';
-
-import * as globalStorybookConfig from '../.storybook/preview';
 import { configureFontAwesomeLibrary } from './web/configureFontAwesomeLibrary';
 
-setProjectAnnotations(globalStorybookConfig);
 configureFontAwesomeLibrary();
 
 window.HTMLElement.prototype.scrollIntoView = jest.fn();
```
ashleysmithTTD, line 11: `setProjectAnnotations` no longer needs to be explicitly called in storybook 9

src/web/components/Core/Banner/Banner.spec.tsx
```diff
@@ -1,12 +1,10 @@
...
-const { Info } = composeStories(stories);
+import { Banner } from './Banner';
+import { Info } from './Banner.stories';
 
 test('has icon and message rendered', () => {
-  render(<Info />);
+  render(<Banner {...Info.args} />);
   const title = screen.getByTestId('banner-message');
   expect(title.textContent).toContain(Info.args?.message);
   expect(screen.getByTestId('banner-icon')).toBeInTheDocument();
```
ashleysmithTTD, line 10: `composeStories` no longer exists so we need a new solution in several spec files

.storybook/preview.tsx
```diff
@@ -1,14 +1,14 @@
...
 const preview: Preview = {
   decorators: [
     (Story) => {
       configureFontAwesomeLibrary();
       return (
-        <div className='app'>
+        <div id="root" className='app'>
           <BrowserRouter>
             <Story />
           </BrowserRouter>
```
ashleysmithTTD, line 14: id needs to be explicit for storybook v9 to find the themes that are stored in root

src/testHelpers/storybookHelpers.ts
```diff
@@ -0,0 +1,3 @@
+export const CreateStory = (story: { render: Function; args: {} }) => {
+  return () => story.render(story.args);
+};
```
ashleysmithTTD, line 3: replacement function to `composeStories` from storybook which no longer exists

craco.config.cjs
```diff
@@ -13,7 +13,7 @@ module.exports = {
           ...webpackConfig.resolve?.fallback,
           stream: require.resolve('stream-browserify'),
           path: false,
-					url: require.resolve('url/')
+					url: require.resolve('url')
         },
       };
```
ashleysmithTTD, line 18: this was needed for `npm run build`

PR #645, opened 2025-06-27
Fixed table deletion by passing correct disable value on keypair update

eiman-eltigani-ttd:
This solution ended up being pretty straight forward, we were updating the name after deletion but not the disabled column

### DIFF
--- a/src/web/components/ClientSideCompletion/ClientSideCompletion.tsx
+++ b/src/web/components/ClientSideCompletion/ClientSideCompletion.tsx
@@ -16,7 +16,8 @@ export function ClientSideCompletion({
   appIds,
 }: ClientSideCompletionProps) {
   const hasKeyPairData = !!keyPairData?.filter(
-    (kp) => !kp.disabled && typeof kp.name === 'string' && !kp.name.includes('-disabled')
+    (kp) =>
+      !kp.disabled && !kp.name?.includes('-disabled')
   ).length;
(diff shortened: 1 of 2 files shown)


ashleysmithTTD:
Nice catch! This fixes the problem if someone adds a keypair and then deletes it now.  But one edge case - the key pairs that have already been disabled since this bug started will still appear in the UI because they do not have their disabled field set to true.  I'll create a ticket to go back through keypairs later to manually set disabled=true for those.  

But for now, we need to also add the filtering, not just on the disabled field, but also on the keypair name containing the `disabled` word

src/web/components/ClientSideCompletion/ClientSideCompletion.tsx
```diff
@@ -15,7 +15,9 @@ export function ClientSideCompletion({
   domainNames,
   appIds,
 }: ClientSideCompletionProps) {
-  const hasKeyPairData = !!keyPairData?.filter((kp) => !kp.disabled).length;
+  const hasKeyPairData = !!keyPairData?.filter(
+    (kp) => !kp.disabled && typeof kp.name === 'string' && !kp.name.includes('-disabled')
+  ).length;
```
ashleysmithTTD, line 20: a quicker fix here is to use the `?` operator to denote only check anything to the right of the `?` if whatever is to the left of the `?` exists instead of doing `typeof kp.name` .  Instead:
```

const hasKeyPairData = !!keyPairData?.filter(
    (kp) => !kp.disabled && !kp.name?.includes('-disabled')
  ).length;
```

ashleysmithTTD (approved):

PR #668, opened 2025-07-03
create primaryContact table migration and model

eiman-eltigani-ttd:
See below for all the test scenarios I ran in SQL, please let me know if I'm missing anything important
 
Context: I only have one user in my db with userId = 1 & six participants with participantId 1-> 6:
Test cases:
o	INSERT INTO [dbo].[primaryContacts] VALUES (1,1)  
o	INSERT INTO [dbo].[primaryContacts] VALUES (1,2) - _CAN add same user to two different participants_
o	INSERT INTO [dbo].[primaryContacts] VALUES (2,2) - _CANNOT add the same participant twice_
    	Cannot insert duplicate key row in object 'dbo.primaryContacts' with unique index 'primarycontacts_participantid_unique'. The duplicate key value is (2)
o	INSERT INTO [dbo].[primaryContacts] VALUES (2,4) - _CANNOT add if userId does not exist_
    	The INSERT statement conflicted with the FOREIGN KEY constraint "primarycontacts_userid_foreign". The conflict occurred in database "uid2_selfserve", table "dbo.users", column 'id'
o	INSERT INTO [dbo].[primaryContacts] VALUES (1,7) - _CANNOT add if participantId does not exist_
    	The INSERT statement conflicted with the FOREIGN KEY constraint "primarycontacts_participantid_foreign". The conflict occurred in database "uid2_selfserve", table "dbo.participants", column 'id'.

### DIFF
--- a/src/database/migrations/20250708133655_PopulatePrimaryContactsTable.ts
+++ b/src/database/migrations/20250708133655_PopulatePrimaryContactsTable.ts
@@ -1,8 +1,8 @@
 import { Knex } from 'knex';
 
 interface PrimaryContactData {
-  userId: number;
   participantId: number;
+  userId: number;
 }
 
 export async function up(knex: Knex): Promise<void> {


eiman-eltigani-ttd:
Just noting here that primaryContact will be based on lowest userId in usersToParticipantRoles because that's how its sorrted, but code is updated now that no users that are exclusively Operation based will be chosen as default primaryContact.

Would also appreciate feedback on naming of variables if you think they're intuitive enough or can be improved!

src/database/migrations/20250708133655_PopulatePrimaryContactsTable.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export async function up(knex: Knex): Promise<void> {
+  const eligibleUsers = (await knex('usersToParticipantRoles')
+    .join('users', 'users.id', '=', 'usersToParticipantRoles.userId')
+    .select('usersToParticipantRoles.participantId', 'users.id as userId')
+    .where('users.deleted', 0)
+    .where('users.locked', 0)
+    .whereNot('usersToParticipantRoles.userRoleId', 2)) as PrimaryUserData[]; // exclude Operations role;
+
```
ashleysmithTTD, line 15: this query ends up being pretty long - let's separate out the last part where you filter on non-operations roles into a new line

eiman-eltigani-ttd: I tried to save this as one line but I think my formatter prevents that

ashleysmithTTD (approved):

PR #690, opened 2025-07-08
Create new primary contacts after adding new participant

eiman-eltigani-ttd:
Insert new entry into primaryContacts db table after creating initial Participant and User

### DIFF
--- a/src/api/entities/PrimaryContact.ts
+++ b/src/api/entities/PrimaryContact.ts
@@ -9,7 +9,7 @@ export class PrimaryContact extends Model {
   }
 
   static get idColumn() {
-    return ['participantId'];
+    return ['participantId', 'userId'];
   }
 
   static get relationMappings(): RelationMappings {
--- a/src/api/entities/Participant.ts
+++ b/src/api/entities/Participant.ts
@@ -13,69 +13,69 @@ export class Participant extends BaseModel {
     return 'participants';
   }
   static get relationMappings(): RelationMappings {
-		return {
-			types: {
-				relation: Model.ManyToManyRelation,
-				modelClass: () => ParticipantType,
-				join: {
-					from: 'participants.id',
-					through: {
-						from: 'participantsToTypes.participantId',
-						to: 'participantsToTypes.participantTypeId',
-					},
-					to: 'participantTypes.id',
-				},
-			},
-			apiRoles: {
-				relation: Model.ManyToManyRelation,
-				modelClass: () => ApiRole,
-				join: {
-					from: 'participants.id',
-					through: {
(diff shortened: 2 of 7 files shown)


eiman-eltigani-ttd:
Summary of recent changes: 
1. Created primaryContact during Participant Creation:
- Include logic to insert into the primaryContact table after Participant and User is created **(ParticipantCreation.ts)** + added additional error handling to prevent potential null values 

3. Introduced primaryContact relation within Participant Obejct (**entity/Participant.ts)** and updated corresponding queries:
- getAllParticipants() now fetches primaryContact **(services/participantsService.ts)**
- getPrimaryContactInformation() now reads Participant.primaryContact to populate info **(EditParticipantDIalogHelper.ts)**

Some things I learned about Objection.js in the process: when declaring a new field within an entity outside of the table columns, we use "Has" to describe the relation which denotes the data is contained in a seperate table. For example, the approverId is a column field that stores a static foreign key and requires direct changes to the Participants table for updates. In contrast, the primaryContact field is sources from the primaryContact table and changes are reflected dynamically. This reflects our approach to keep primaryContact retrieval efficient and flexible for future 1:N relationships (rather than creating a new primaryContact column within participant table).

src/api/entities/PrimaryContact.ts
```diff
@@ -9,7 +9,7 @@ export class PrimaryContact extends Model {
   }
 
   static get idColumn() {
-    return ['participantId'];
+    return ['participantId', 'userId'];
   }
```
ashleysmithTTD, line 13: Why did you also add userId here?

ashleysmithTTD: I don't think this is the case - I think this combination will mean that there can only be unique combinations of participantId & userId.  But we only want participantId to be unique

eiman-eltigani-ttd: participantId is still unique! It's specified directly on the column in the schema builder:
`table.integer('participantId').references('participants.id').unique();`
I made the primary key a composite key because I assumed we would add new entries for the same participant in the future. In that case, we would need both fields to ensure each id value is unique since we're not using an incremental integer count. That enables us to add more rows in the future, but currently *.unique()* locks us from adding the same participant twice and would need to later be removed.
 
It works the same practically (we can still query by the fields individually) but if you want to query by .findById(), you need to pass both fields along. I could remove userId in the entity file and it would work fine because we aren't using that function currently, but would break if we did. It also just hit me that we could potentially add a new column for each additional contact and in that case, we could just use participantId as the main id (this would require a schema change). Let me know if you think that approach is more likely and worth making the change

PR #694, opened 2025-07-10
Highlight Primary Contact within TeamMembers table

eiman-eltigani-ttd:
Created first UI change on TeamMembers table to add "Primary Contact" button on TeamMember row. Updated participant middleware to return primaryContact field for easy access.

### DIFF
--- a/src/api/middleware/participantsMiddleware.ts
+++ b/src/api/middleware/participantsMiddleware.ts
@@ -6,7 +6,9 @@ import { getTraceId } from '../helpers/loggingHelpers';
 import { ParticipantRequest } from '../services/participantsService';
 import { canUserAccessParticipant } from './usersMiddleware';
 
-const participantIdSchema = z.object({ participantId: z.coerce.number() });
+const participantIdSchema = z.object({
+  participantId: z.coerce.number(),
+});
 export const verifyAndEnrichParticipant = async (
   req: ParticipantRequest,
   res: Response,
--- a/src/web/components/TeamMember/TeamMember.tsx
+++ b/src/web/components/TeamMember/TeamMember.tsx
@@ -90,6 +90,15 @@ function TeamMember({
     }
   };
 
+  const actionButtonDelete = (
+    <ActionButton
+      onClick={onOpenChangeTeamMemberRemoveDialog}
+      icon='trash-can'
+      aria-label='Remove Team Member'
+      disabled={isPrimaryContact}
+    />
+  );
+
   return (
     <tr>
       <td>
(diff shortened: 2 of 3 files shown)


eiman-eltigani-ttd:
Summary of changes:
- Updated participant middleware to pull primaryContact field accessed via useContext
- Refactor some styling to fit tooltip icon neatly **<-- feedback on instructions appreciated**

src/api/middleware/participantsMiddleware.ts
```diff
@@ -6,9 +6,7 @@ import { getTraceId } from '../helpers/loggingHelpers';
 import { ParticipantRequest } from '../services/participantsService';
 import { canUserAccessParticipant } from './usersMiddleware';
 
-const participantIdSchema = z.object({
-  participantId: z.coerce.number(),
-});
+const participantIdSchema = z.object({ participantId: z.coerce.number() });
```
ashleysmithTTD, line 9: Looks like the formatting is off here again without a change

ashleysmithTTD:
I dont think we want this info logo here - that adjusts the spacing in the table - instead, to hover over the disabled deleted button and include the description.  There are examples of this in the Sharing Permissions table if you bulk add any participants.  
infotooltip
sharingpermissionsexample

eiman-eltigani-ttd:
Updated to include hover over message on disabled trashcans. Note - I had to remove an additional <div> in TeamMemeber.tsx because it was messing up the intended layout of the .action-cell style. From my testing, it doesn't look like it affects the look of the error message but wasn't sure if that was put there intentionally since there was no styling associated with it

src/web/components/TeamMember/TeamMember.tsx
```diff
@@ -111,49 +127,49 @@ function TeamMember({
...
+                onOpenChange={onOpenChangeTeamMemberRemoveDialog}
               />
-              {showTeamMemberRemoveDialog && (
-                <TeamMemberRemoveConfirmationDialog
-                  onRemoveTeamMember={handleRemoveUser}
-                  person={person}
-                  onOpenChange={onOpenChangeTeamMemberRemoveDialog}
-                />
-              )}
-            </div>
```
ashleysmithTTD, line 156: why did this <div> get removed?

eiman-eltigani-ttd: It was affecting the action-cell layout by making this extra <div> and the errorMessage the child components rather than the actions buttons/icons inside the div. It was causing alignment issues when I added the ToolTip which is its own container, I needed to make sure all elements are direct child elements for the flex style to work correctly

ashleysmithTTD (approved):

PR #711, opened 2025-07-22
upload and parse csv file to cstg table

eiman-eltigani-ttd:


src/web/components/ClientSideTokenGeneration/CstgAddDialog.tsx
```diff
@@ -95,6 +96,34 @@ function CstgAddDialog({
...
+    if (!file) return;
+
+    Papa.parse(file, {
+      complete: (results: Papa.ParseResult<unknown>) => {
+        const parsedValues = results.data
+          .flat()
+          .map((val) => String(val).trim())
+          .filter((val) => val.length > 0);
+
+        if (parsedValues.length === 0) return;
```
ashleysmithTTD, line 110: Instead of just returning here, we should do `handleError` here as well to let the user know that no parsed values came from the CSV file

src/web/components/ClientSideTokenGeneration/CstgAddDialog.scss
```diff
@@ -21,4 +20,36 @@
...
+
+      &:hover {
+        border-color: #888;
+      }
+
+      &:focus {
+        outline: none;
+        border-color: #007aff;
+        background-color: #fff;
+      }
```
ashleysmithTTD, line 52: we don't have many places where we use the hex code directly for the color - for #fff, you can just say white, for the border-color its a close enough blue that we should just use one of our blues from themes or colors.  there's also probably a gray that in themes or colors for hover

src/web/components/ClientSideTokenGeneration/CstgAddDialog.tsx
```diff
@@ -95,6 +96,34 @@ function CstgAddDialog({
...
+
+        const existing = formMethods.getValues('cstgValues') || '';
+        const newValues = existing
+          ? `${existing}\n${parsedValues.join('\n')}`
+          : parsedValues.join('\n');
+        formMethods.setValue('cstgValues', newValues);
+      },
+      error: (_err) => {
+        handleError('Failed to parse CSV file. Please try again.');
+      },
```
ashleysmithTTD, line 123: you don't need `_err` here since you don't use it in `handleError`

src/web/components/ClientSideTokenGeneration/CstgAddDialog.tsx
```diff
@@ -95,6 +96,37 @@ function CstgAddDialog({
...
+  function handleCsvUpload(e: React.ChangeEvent<HTMLInputElement>) {
+    const file = e.target.files?.[0];
+    if (!file) return;
+
+    Papa.parse(file, {
+      complete: (results: Papa.ParseResult<unknown>) => {
+        const parsedValues = results.data
+          .flat()
+          .map((val) => String(val).trim())
+          .filter((val) => val.length > 0);
```
ashleysmithTTD, line 108: getting a couple of typescript errors here due to `val` being "implicitly any type".  the first val should be `val: unknown` since we can't be sure what will come from the csv, then the second is `val: string`

src/web/components/ClientSideTokenGeneration/CstgAddDialog.scss
ashleysmithTTD: Still need to fix `border-color` in the hover

ashleysmithTTD (approved):

PR #714, opened 2025-07-24
Disable and add hover message on Operations role key

eiman-eltigani-ttd:


eiman-eltigani-ttd:
Changes:
- Track checkbox and disable 'Operations' key if  isPrimaryContact is checked
- Update radio button to include tooltip message
- Fixed minor styling discrepencies between enabled and disabled checkbox text

PR #724, opened 2025-08-04
Fix domain extraction to allow multi-tenant domain names to be passed

eiman-eltigani-ttd:
Previously, multi-tenant domains (i.e., .vercel.app, .github.io) failed to pass validation because our exctraction logic didn't account for private suffixes which has to be explicitly required.

### DIFF
--- a/src/web/components/ClientSideTokenGeneration/CstgEditDialog.tsx
+++ b/src/web/components/ClientSideTokenGeneration/CstgEditDialog.tsx
@@ -53,15 +53,14 @@ function CstgEditDialog({
 
     if (cstgValueType === CstgValueType.Domain) {
       const topLevelDomain = extractTopLevelDomain(updatedCstgValue);
-      if (topLevelDomain) {
-        updatedCstgValue =  topLevelDomain;
-      } else {
+      if (topLevelDomain === 'publicSuffix') {
         setError('root.serverError', {
           type: '400',
           message: `The ${formattedCstgValueType} is a public suffix and not allowed.`,
         });
         return;
       }
+      updatedCstgValue = topLevelDomain;
     }
     if (updatedCstgValue === originalCstgValue) {
       onOpenChange();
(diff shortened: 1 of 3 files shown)


src/web/components/ClientSideTokenGeneration/CstgHelper.ts
```diff
@@ -37,10 +37,14 @@ export const formatCstgValueType = (valueType: CstgValueType) => {
...
-  if (topLevelDomain && topLevelDomain !== domainName) {
-    return topLevelDomain;
+export const extractTopLevelDomain = (domainName: string): string | null => {
+  const topLevelDomainInfo = parse(domainName, { allowPrivateDomains: true });
+  if (topLevelDomainInfo.domain) {
+    return topLevelDomainInfo.domain;
+  }
+  // Return null if it's a public suffix
+  if (topLevelDomainInfo.isPrivate) {
+    return null;
```
ashleysmithTTD, line 47: we should return something more specific than just null to be more clear - since it returns a string, maybe return a string like "publicSuffix" then check for that string in the check above.  because maybe there could be other reasons we'd want to return null in the future.

src/web/components/ClientSideTokenGeneration/CstgEditDialog.tsx
```diff
@@ -52,16 +52,25 @@ function CstgEditDialog({
     const originalCstgValue = cstgValue;
 
     if (cstgValueType === CstgValueType.Domain) {
-      updatedCstgValue = extractTopLevelDomain(updatedCstgValue);
+      const topLevelDomain = extractTopLevelDomain(updatedCstgValue);
+      if (topLevelDomain) {
+        updatedCstgValue =  topLevelDomain;
```
ashleysmithTTD, line 57: too much spacing with the equal sign

src/web/components/ClientSideTokenGeneration/CstgHelper.ts
```diff
@@ -37,11 +37,15 @@ export const formatCstgValueType = (valueType: CstgValueType) => {
   }
 };
 
-export const extractTopLevelDomain = (domainName: string) => {
+export const extractTopLevelDomain = (domainName: string): string => {
   const topLevelDomainInfo = parse(domainName, { allowPrivateDomains: true });
   if (topLevelDomainInfo.domain) {
     return topLevelDomainInfo.domain;
   }
+  // Return null if it's a public suffix
```
ashleysmithTTD, line 45: don't need this comment anymore

PR #726, opened 2025-08-11
Standardize routing logic for user controller

eiman-eltigani-ttd:


src/api/services/userService.ts
```diff
@@ -32,115 +31,112 @@ export const UpdateUserRoleIdSchema = z.object({
...
+    const kcAdminClient = await getKcAdminClient();
 
-    await performAsyncOperationWithAuditTrail(auditTrailInsertObject, traceId, async () => {
-      await UserToParticipantRole.query()
+    await UserToParticipantRole.transaction(async (trx) => {
+      await UserToParticipantRole.query(trx)
+        .where('participantId', participant!.id)
         .where('userId', user!.id)
-        .andWhere('participantId', participant!.id)
+        .whereNot('userRoleId', UserRoleId.UID2Support)
```
ashleysmithTTD, line 93: Why was this check added?

src/api/routers/businessContactsRouter.ts
```diff
@@ -14,40 +14,45 @@ export const BusinessContactsDTO = BusinessContactSchema.pick({
   contactType: true,
 });
 
+// Handlers
```
ashleysmithTTD, line 17: I don't think these comments are needed for handler/router in a few of these files, its clear enough by the names of the functions

src/api/routers/managementRouter.ts
```diff
@@ -35,6 +36,7 @@ const handleChangeUserLock = async (req: ParticipantRequest, res: Response) => {
   return res.status(200).end();
 };
 
+// Router
```
ashleysmithTTD, line 39: same comment here as above - not needed

src/api/services/userService.ts
ashleysmithTTD: Oh i think the PR is mixing up `updateUser` and `removeUser` - could you switch the order so it matches the order of main, just so it makes the review of this file a little easier

ashleysmithTTD (approved):

PR #727, opened 2025-08-12
Clean up and organize various search bar components

eiman-eltigani-ttd:
I consilidated all of the search bar table headers into single reusable <TableSearchBar> in addition to the <SearchBar> component that already exists since they have different styling and funcionality. Hopefully this clarifies the distinction between SearchBar, ParticipantSearchBar (built off SearchBar.tsx), and TableSearchBar.


src/web/components/Search/TableSearchBar.tsx
```diff
@@ -0,0 +1,28 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React from 'react';
+
+import './TableSearchBar.scss';
+
+interface TableSearchBarProps {
+  // updated
```
ssundahlTTD, line 7: leftover comment

src/web/components/ClientSideTokenGeneration/CstgTable.scss
```diff
@@ -34,7 +34,7 @@
     justify-content: space-between;
     align-items: baseline;
 
-    .search-bar-container {
+    .table-search-bar-container {
```
ssundahlTTD, line 37: This feels a little weird.  You're referencing a self-contained component's internal CSS here.  It's a little dangerous relying on that component not changing or even permanently using that specific component.  Maybe wrap the search component in another container and style the outer container the way you want it.

ssundahlTTD (changes requested):

ssundahlTTD (approved):

PR #838, opened 2026-07-08
UID2-7444: Fix SSP login⇄logout trap for newly-invited users

mcollins-ttd:
A newly-invited SSP user who accepts their invite gets stuck bouncing
between login and logout, never reaching the portal.

The invite email redirects to /logout on purpose (UID2-1579: tear down
any stale inviting-account session and force a fresh login). /logout
renders inside a PrivateRoute, and its Logout screen calls
keycloak.logout() with no redirectUri. keycloak-js then defaults the
post-logout redirect to the current URL, so once the Keycloak server
tears down the session it sends the browser back to /logout — now
unauthenticated.

This was harmless until ee45cbd3 ("Pass kc_idp_hint to Keycloak login
when present in URL", 2026-03-09), which changed PrivateRoute's post-login
redirect from the site root to the current path (origin + pathname +
search) so deep links and kc_idp_hint survive login. Correct for real
pages, but on /logout it creates a trap: after logout the browser lands
back on /logout unauthenticated, PrivateRoute redirects the user to log
in, and logging in returns them right back to /logout.

Fix: pass redirectUri: window.location.origin to keycloak.logout() so the
browser lands on the site root after logout, breaking the cycle. Scoped to
/logout, so ee45cbd3's deep-link behavior is preserved everywhere else, and
UID2-1579's session teardown still runs — only where logout returns changes.

Add a regression test asserting Logout calls keycloak.logout with
redirectUri set to the origin.

### DIFF
--- a/src/web/screens/logout.test.tsx
+++ b/src/web/screens/logout.test.tsx
@@ -1,0 +1,46 @@
+import { render } from '@testing-library/react';
+
+import * as keycloakMocks from '../../testHelpers/keycloakMocks';
+import * as KeycloakProvider from '../contexts/KeycloakProvider';
+import { LogoutRoute } from './logout';
+
+// Needed to enable spying on useKeycloak
+jest.mock('../contexts/KeycloakProvider', () => ({
+  __esModule: true,
+  ...jest.requireActual('../contexts/KeycloakProvider'),
+}));
+
+describe('Logout', () => {
+  let kcMock: jest.SpyInstance;
+  let logoutSpy: jest.Mock;
+
+  beforeEach(() => {
+    logoutSpy = jest.fn(() => Promise.resolve());
+    const mockKeycloak = keycloakMocks.mockAuthenticatedKeycloak();
+    // Override logout on the Keycloak-typed object rather than rebuilding it via spread, which
+    // would drop the class's method members and no longer satisfy the Keycloak type.
+    mockKeycloak.keycloak.logout = logoutSpy;
+    kcMock = jest.spyOn(KeycloakProvider, 'useKeycloak').mockImplementation(() => ({
+      ...mockKeycloak,
+      authenticated: true,
+    }));
+  });
+
+  afterEach(() => {
+    kcMock.mockRestore();
+  });
+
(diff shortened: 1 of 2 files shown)


UID2SourceAdmin:
## 🤖 AI Code Review

_Generated by Azure OpenAI (gpt-5). Automated — human review is still required._

## Summary
This PR changes the logout route to pass an explicit `redirectUri` to Keycloak so users are sent to the site root after logout instead of back to `/logout`, avoiding the login/logout loop described in UID2-7444. It also adds a unit test verifying the logout call uses that redirect target.

## Strengths
- Good root-cause documentation in `logout.tsx`; the comment explains the regression and intent clearly.
- The added test checks the key behavioral change directly instead of only snapshotting the rendered output.
- Nice touch preserving the typed Keycloak mock by overriding `logout` in place rather than rebuilding the object.

## Findings

### 🟡 Important
**Redirect target is now hard-coded to the origin, which can break non-root deployments or Keycloak redirect-uri config**  
`window.location.origin` drops any app base path. If this portal is deployed under a subpath (for example `/portal`) or Keycloak is configured with exact post-logout redirect URIs for that path, logout can fail or send users to the wrong location. Consider deriving this from the app’s configured base URL / router basename rather than assuming `/`.  
`src/web/screens/logout.tsx:13`

### 🟢 Minor
**The regression test doesn’t cover the actual loop condition**  
The new test only asserts that `logout()` receives `{ redirectUri: window.location.origin }`. It does not exercise the unauthenticated `/logout` + `PrivateRoute` interaction described in the comment, so a future routing change could reintroduce the loop without failing this test. Add an integration-style test that navigates to `/logout` after logout and asserts the user is not redirected back into the same route.  
`src/web/screens/logout.test.tsx:36`

## Verdict
**Ready to merge?** With fixes

**Reasoning:** The functional intent is good and the change is small, but the new redirect target is a deployment-sensitive contract change. I’d want the redirect URI derived from configured app base URL (or the deployment assumptions documented and validated) before merging.

swibi-ttd (approved):

